PR #23, opened 2017-03-24
Erase flow tags

diegomura:


jbovenschen:
If you're interested we could also use flow within the codebase?

diegomura:
yeah! That would be cool. Meanwhile, we should clean the codebase from types tags

jbovenschen:
Sure thing, be free to merge!

PR #26, opened 2017-04-03
Add yoga as an layout engine, and write some internal pdf instance logic

jbovenschen:
This is far from release ready, but could make it easier for others to work on features beside me.


packages/react-pdf-dom/src/Document.js
```diff
@@ -0,0 +1,58 @@
...
+
+    this.state = {
+      document: undefined,
+    };
+  }
+
+  componentDidMount() {
+    this.mountNode = PDFRenderer.createContainer(this.container);
+    PDFRenderer.updateContainer(this.props.children, this.mountNode, this);
+    this.renderer.toBlob(this.container);
```
diegomura: This is duplicated, right?

packages/react-pdf/src/renderer.js
```diff
@@ -89,24 +92,28 @@ const PDFRenderer = ReactFiberReconciler({
...
     textInstance.text = newText;
   },
 
   appendChild(parentInstance, child) {
-    const index = parentInstance.children.indexOf(child);
-    if (index !== -1) {
-      parentInstance.children.splice(index, 1);
-    }
-    parentInstance.children.push(child);
+    child.inject(parentInstance);
```
diegomura: Shouldn't be adding the child to the parentInstance, and not the opposite as we are doing? I know it may seem the same. but it's a bit confusing. Would you mind explaining to me?

packages/react-pdf/src/renderer.js
```diff
@@ -138,8 +146,8 @@ const PDFRenderer = ReactFiberReconciler({
   },
 });
 
-function createPDFInstance(inst) {
-  const { doc, firstPageSkipped } = inst.rootContainerInstance;
+function createPDFInstance(inst, doc) {
+  const { firstPageSkipped } = inst.rootContainerInstance;
   const { children, ...props } = inst.props;
 
   switch (inst.type) {
```
diegomura: Program never enters this switch right?

packages/react-pdf/src/elements/View.js
```diff
@@ -0,0 +1,11 @@
+import Base from './Base';
+
+class View extends Base {
+  constructor() {
+    super();
+
+    this.type = 'VIEW';
```
jbovenschen: Not yet, will remove it for now good catch!

packages/react-pdf/src/renderer.js
jbovenschen: Not any more, will spend some time to fix the node renderer, and remove this part of the code.

packages/react-pdf/src/renderer.js
jbovenschen: Dit this the other way around, because of how yoga-layout works... Internally it would switch from child to parent, will check if I can easily make it work the other way around.

packages/react-pdf-dom/src/Document.js
```diff
@@ -0,0 +1,58 @@
+/* global URL */
+import React, { Component, PropTypes } from 'react';
+import { PDFRenderer, createElement, pdf } from 'react-pdf';
+
+class Document extends Component {
+  container = createElement('DOCUMENT');
+
+  static propTypes = {
+    children: PropTypes.any,
```
jbovenschen: 👍

packages/react-pdf-dom/src/Document.js
jbovenschen: Not really, because you don't have to talk to the public `PDFRenderer.render()` method for making it to work within react-dom and react-native. Took this from the react-art renderer, and how they did it.

packages/react-pdf/src/renderer.js
jbovenschen: Thank you for finding this one, there was really no need anymore to doing it this way!

jbovenschen:
I merge this one already, if you have any comments just add them and I will try to answer it.

PR #27, opened 2017-04-03
Add a document component for meta data.

jbovenschen:
This also include a stupid mistake within tree traversal :)

This was needed for including fonts and other meta data within the PDF file. Needs to be always the root of the tree.

### DIFF
--- a/examples/dom-bindings/src/App.js
+++ b/examples/dom-bindings/src/App.js
@@ -1,8 +1,8 @@
 import React, { Component } from 'react';
 import './App.css';
 
-import { Document } from 'react-pdf-dom';
-import { View, Text, Page, StyleSheet } from 'react-pdf';
+import { Document as Pdf } from 'react-pdf-dom';
+import { View, Text, Page, Document, StyleSheet } from 'react-pdf';
 
 const styles = StyleSheet.create({
   container: {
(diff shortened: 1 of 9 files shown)


examples/dom-bindings/src/App.js
```diff
@@ -23,32 +23,34 @@ class App extends Component {
...
         <p className="App-intro">
-          <Document height="100%" width="100%">
-            <Page size="A4">
-              <View style={styles.container}>
-                <Text>
-                  Text
-                </Text>
-                <View style={styles.block}>
+          <Pdf height="100%" width="100%">
+            <Document>
```
diegomura: This `Pdf` and `Document` seems a bit redundant right? Is it possible to use the Document of the specific platform as the root of the document, and do not have a Document from `react-pdf` ?

packages/react-pdf/src/pdf/index.js
```diff
@@ -219,6 +219,8 @@ ${offsets[offsets.length - 1]}
       pages,
     } = traverseTree(input);
 
+    console.log(pages);
```
diegomura: oops

examples/dom-bindings/src/App.js
jbovenschen: Yeah was also thinking about doing that, then the specific platform component would just render the Document internally, will change it ASAP.

jbovenschen: Still choose to keep the Document component in react-pdf, because it is still usefull in every renderer. But the react-pdf-dom component renders now the Document component from the react-pdf package, and only omits the react-dom specific properties.

diegomura:
Looks good to me!! Feel free to merge :shipit:

jbovenschen:
Thank you!

PR #31, opened 2017-05-03
Refactor pdf.js

diegomura:


diegomura:
@jbovenschen here I suggest a better way of handling the pdf module. I'm not very happy with the solution though. It works for now, but I'm not sure how difficult will be to implement more complex features.
I think we need to focus on this part before thinking on anything else. We need a layer to handle pdf generation elegantly in order to not having issues when we add the react part

jbovenschen:
Will update this PR this weekend, with some changes I have in my local branch for image support, which will make it possible to remove the tree traversal.

jbovenschen:
Was experimenting with this branch this weekend, one thing that came to me was;
Loss of support for nested pages.
```javascript
<Page>
   <Page>
      <Page><Text>Page 1.1.1.</Text></Page>
      <Text>Page 1.1</Text>
   </Page>
</Page>
```
This worked before but doesn't now.

Were some little things I experimented with afterwards like;
1. Removing the need of looping through the instances to generate the final PDF.
2. Returning multiple objects within the render method,
3. Caching render instances is not updated trees.

diegomura:
why should we allow rendering a page inside a page. that doesn't make sense to me

diegomura:
I been rewriting this implementation. This time without traversing the tree as I was doing before. I still have some work to do, but I'm pretty happy with the solution. Specially with the part of having each element know how he should render in the final pdf document. 

Doing this research I came up with new ideas to add, such as making use of the PDF incremental updates feature, or using style inheritance (which the PDF spec also supports for some objects). I'll be uploading this asap

diegomura:
@jbovenschen so, I finish most of the implementation, and I'm very happy about how it's working. Of course, there is a lot to do yet, but I think this solution give us the basic ground to start developing everything else. I'm going to merge this if you dont mind 😄

PR #43, opened 2017-05-19
Improve build tooling

jbovenschen:
I improved the build tooling to make it easier to work on react-pdf.

The flow how you can start to work on react-pdf is as following now; 

- Open 2 seperate terminal windows
- Run `$ yarn run build:watch` in the first
- Run `$ yarn run example fractals` in the second

Also did some nitpicking within the packages to make them really private so we don't accidentally publish them with lerna, and made the main point to the transpiled version of the libs.


diegomura:
About lerna, the `lerna publish` didnt work for me today. I didnt had much time so I went package by package doing `npm publish`, but we'll have to see whats going on (or what I did wrong). On the other stuff, sounds good to me

jbovenschen:
Will test it if it works for me now. If not I will update this PR.

PR #45, opened 2017-05-19
Add snapshot testing

diegomura:


diegomura:
Not completely sure if Travis works fine yet, but it should. I'll add more tests in the future

PR #46, opened 2017-05-20
Make examples a unique package

diegomura:
This roll back the examples package as it was. 
I truly don't see the purpose of having one package per example when we can have one main `examples` package and run all of them from here in the same way.
I see myself adding lots of different examples in the future, and it's easier if each folder has only the `js` files needed to run and nothing more. Also, it will be a lot easier to create a new one. No new `package.json` or `yarn.lock` needed.


jbovenschen:
I understand what you want, but the main difficulty is that it is not really suitable for the `dom` and future `native` examples. Isn't it a good idea to split those examples into:

- examples
  * dom
  * native
  * node

So we don't have react-native, react-dom and other dependencies inside one folder.

diegomura:
I like the idea of splitting examples in these three categories, even though I don't see that much harm on having all dependencies together. They aren't that much.

What I think is that we should definitely avoid having a package-per-example. I want examples to be easy to run for people to test it, easily understandable by people reading the source, and also easy to develop and add new ones.

Another thing we can do is instead of having separate examples for all platforms, reusing the same examples and letting the user decide on which platform he wants to run them. Something like:

```
yarn example -- fractals dom
// or
yarn example -- fractals node
```

It can also be using parameters like `-p` for platform or whatever. But that's basically the idea. 
@jbovenschen What do you think?

jbovenschen:
Great, then people could really see the benefit of a cross platform pdf-renderer :)

diegomura:
Exactly. I'll merge this and take this approach in another PR, since it wont be easy to do. Ill create a ticket

PR #49, opened 2017-05-20
Fix license recognition

HoBi:
By this change GitHub may be able to recognize the repository license and show it in the header.


jbovenschen:
Thanks for pointing this out!

PR #50, opened 2017-05-20
Rename some internal instances to match component names

jbovenschen:
Rename some internal instances to match component names:

Document => Root
Pages => Document

This makes it more clear where what happens


diegomura:
This is tricky 🤔 
I definitely see your point.
The current naming is based on the PDF file specification (where `Pages` renders a `/Type /Pages` pdf object), and not on the Component naming and structure.

So the question here is: shouldn't we keep the pdf rendering engine something that is independent of the dom/node/native bindings on these type of things? We talked a bit on the past about this.

I honestly don't have an strong opinion on this, but from the PDF document point of view, the current solution is clearer.

jbovenschen:
Yeah understand you're point totally, personally I am thinking about splitting the react primitives from the PDF related stuff.

So what you would get after that is done is that the Components you render keep track of the current props and styles, based on the props within the current state of your tree it will create new or remove old PDF instances from the nodes array or the PDF Root.

Also about the PDF issue, this will be solved when we treat our Document different.
So the document will not track pages this will be done within the Root of our application which is the glue between the React and PDF side, what the document is after that is nothing more than just something where we can assign our global props like fonts etc...

diegomura:
What doesn't makes sense to me is having this class `Document` in @react-pdf/core which by the name one should think is the main class of the pdf document, but actually just returns a pdf object with `/Type /Pages`.

So for what I see, we will always have some conflict of concepts between Root and Document. I think that's because we don't have the concept of `ROOT` in the Component layer, but we do on the renderer.

On your solution, any prop you pass to `<Docuement />` will be received by the element responsible of rendering the `/Pages`object (named yes, `Document`),  while it should be `Root`who receive them.

jbovenschen:
Currently I try to think in 2 concepts `Element` and `Instance` inside the `react-pdf/core`.
This will make it more clear I guess where what will happen, and makes it easier to do some memorizations in the near future.

**Element**
An element is something that can create an instance but not an element
and only used to track UI and needed instances.

So this is only created or removed at the moment that something in the tree is changed, and calculates layout based on that.

Those Elements exists only of the primitives used in the `react-primitives` and `react-svg-primitives` packages.

**Instance**
An instance is something that can create and instance but not an element
and is used to define what should be rendered within the pdf file.

So those are only created when an Element or Instance needs it to render within multiple or a single 'PDF' object. And only get an update if there where any related properties updated in the Element or Linked instance, if so It can be changed inside the PDF output.

Those instances exists of everything that can be added in a PDF like:
`Image`, `Shapes`, `Pages`, `Outlines`, `Catalog` and so forth... 

@diegomura what do you think?

diegomura:
I think thats a bit confusing haha. Sorry. At least the way you defined both concepts.

I want to keep this simple for the moment. There is no need to think on updates right now since there are a bunch of basic things left to do.

> On your solution, any prop you pass to <Docuement /> will be received by the element responsible of rendering the /Pages object (named yes, Document), while it should be Root who receive them.

Dont know if you got this, but this is a real issue. The props passed to `<Document />` are being received by the object responsible of rendering the pages, whether is called `Pages` or `Document`

jbovenschen:
#52 was a better fix for now.

Issue #47, opened 2017-05-20
Add a consistent way of applying default props and styles


diegomura:
We dont have yet a neat way of defining and appying what are the default props and styles to be applied to elements. 

I imagine something like:
```
class Text extends PDFEntry {
  defaultProps = {
    // props here
  }

 // ...
}
```

jbovenschen:
I'm ok with implementing something like this, but is there any need to name it defaultProps? It could be just named props right?

Also how do you think we can preserve the changed props within the PDFEntry instance?

jbovenschen:
Did implement it for only the props at this moment. I think it would be a good idea to split up the styling related properties (created with `Stylesheet.create`) and the other properties. What do you think?

PR #51, opened 2017-05-20
Implement default instance properties

jbovenschen:
This is my implementation of #47.
We need to figure out how we will call / trigger an update when the renderer triggers one, currently the renderer works only in one single loop.


packages/react-pdf/src/elements/PDFEntry.js
```diff
@@ -16,7 +16,7 @@ class PDFObject extends Base {
     root.addNode(this);
 
     if (props) {
-      this.applyProps(props);
+      this.applyProps(this.props);
```
diegomura: Should be `if (this.props)` also

PR #52, opened 2017-05-20
Reuse Document instance as the renderer root container

diegomura:
This is my suggestion for the issue that https://github.com/diegomura/react-pdf/pull/50 brings to light, or at least something similar to this.

@jbovenschen what do you think?


diegomura:
@jbovenschen shall we close https://github.com/diegomura/react-pdf/pull/50 then?

Issue #61, opened 2017-05-21
Fix TravisCI integration


diegomura:

jbovenschen:
This is caused by #61.

diegomura:
Travis is working fine now. I'm closing this

PR #66, opened 2017-05-21
Run Travis only on the osx image for now

jbovenschen:
This is because yoga-layout needs to compile itself first before we can use it, and the dependencies to do so are not default available on linux.

For now I only enabled osx, in the future we need to do the same for windows and linux.


jbovenschen:
This should fix #61

PR #67, opened 2017-05-21
Update yoga layout

jbovenschen:
Updated all the examples and tests accordingly.


packages/react-pdf/src/elements/Page.js
```diff
@@ -40,7 +40,7 @@ class Page extends PDFEntry {
   createsBackgroundColor() {
     const layout = this.layout.getComputedLayout();
 
-    return new Rectangle({ ...layout, style: this.style });
+    return new Rectangle({ ...layout, style: this.style || {} });
```
diegomura: I solved this by making use of defaultProps, so this should be discarded.

diegomura:
Closing this. Got obsolete

jbovenschen:
Why do you think this got obsolete? We still need to update though has some minor improvements for width and height calculations. Will make a separate PR with all the dependencies of our packages updated.

diegomura:
@elpic resolved most of the things you did on this PR: fixing travis, and updating the examples. The Rectangle fix I made it with defaultProps in another PR, and it's not needed anymore since we will use pdfkit. right? What width and height calculations are you talking about?

jbovenschen:
The 1.3 version did rounding for the values so it doesn't return any decimals, this is improved in the 1.5 version. Also we still need to use the yoga-layout stuff right? Else we would not have an API users know from React-native, react-sketchapp and others :) We can just feed pdfkit the positions and dimensions and it then should work.

PR #92, opened 2017-06-05
Adding more test 

elpic:
Use the way described on async testing document (
https://facebook.github.io/jest/docs/tutorial-async.html) to handle
async test. And adding more tests.


diegomura:
I liked using `async/await` on the tests. @elpic would you want to refactor the existing ones to be like there?

Issue #6, opened 2016-10-22
Add links support


diegomura:
For ref: https://github.com/devongovett/pdfkit

PR #100, opened 2017-06-23
Adding basic support for Link component

elpic:
Created Link component that could be added in anypart of the page,
except with a Text component. That requires extra work to calculate the
place in the text the Link should go.


diegomura:
This is freaking great!!
@elpic do you mind adding some spec for Links?

diegomura:
Fixes #6

PR #107, opened 2017-06-28
Embed Links inside Text

### LINKED ISSUE #104: Embed Links inside Text
Text rendering phase should support many children, and not just a text string as it is right now

diegomura:
Add the ability to render inline links element (or any other future element that we might add) inside text elements.

Fixes #104, #105


diegomura:
Haha, that's just the thumbnail image for the examples section. Not the pdf itself

Issue #111, opened 2017-06-29
Provide a default protocol for links (if not given)


diegomura:
Probably `http`

PR #112, opened 2017-06-30
Add protocol to links that doesn't have it

elpic:
When the src attribute doesn't have a protocol add `http` as default
protocol.


diegomura:
@elpic you rock. thanks!

diegomura:
Fixes #111

Issue #101, opened 2017-06-24
can't get DOM bindings to work


hyperborea:
I tried to make the package work in the browser, but ran into a bunch of issues along the way:
- the official example uses an outdated version of `@react-pdf/dom (0.1.6)` referencing the wrong core package (`react-pdf` instead of `@react-pdf/core`)
- browser environments don't provide fs, tls and net. create-react-app conveniently mocks them out, but I figured to still mention it for people using a different setup
- with above two points fixed I'm still getting below error, which I wasn't able to resolve yet
<img width="712" alt="screen shot 2017-06-24 at 12 31 46" src="https://user-images.githubusercontent.com/5477116/27507870-700dca60-58d9-11e7-9b33-4efc8ec56a7a.png">

Does anyone know what's the matter or has a local setup that works browser-side?

hyperborea:
Adding babel-polyfill to `@react-pdf/core` adds support for async generators, but then the problems continue: 
<img width="709" alt="screen shot 2017-06-24 at 13 59 13" src="https://user-images.githubusercontent.com/5477116/27508465-ba7202a4-58e5-11e7-97fd-513df07d22f2.png">

If anyone could provide me with a working example that would be great.

diegomura:
Hi @hyperborea .
Thanks for reporting all these issues.

We are still working on supporting react-pdf to the dom. A while ago we switch from using our own pdf bindings to use `pdfkit`. I'm not into this subject so much (@jbovenschen is, maybe he can give more info about it), but `pdfkit` has some issues working in the browser with webpack. @jbovenschen is on it, but I'm not sure if he had some progress yet. I was planning on helping him, since this is an important issue to the project. So I think the `fs` error is because of this.

About the other things, I will:
- Fix the examples dependencies issues.
- Add `babel-polyfill` to `@react-pdf/core`

I'll keep this issue opened to let other people know about this, and also to let people say if he/she wants to help on this.

hyperborea:
Thanks @diegomura, I'm having high hopes for this project and look forward to see how it develops. Fingers crossed that browser support is coming back soon, I'd ideally like to use it in both environments in parallel.

PS: adding babel-polyfill is probably overkill, best check out the last point under "Troubleshooting" here: https://www.npmjs.com/package/babel-watch.

jbovenschen:
[...]

Issue #57, opened 2017-05-21
Yoga does not compile in Linux


diegomura:

jbovenschen:
Emcscripten which is used to compile yoga-layout can not be used on Linux, this because of that it compile CLang from it's sources.

Check [this](https://github.com/facebook/yoga/blob/master/.travis.yml#L53-L61) for more info.

jbovenschen:
Yoga does work with the new release without needing a build step, but this will be triggered by scripts inside the package.json, does not know if we could ignore them for a single package? @diegomura you have any idea?

diegomura:
@jbovenschen is this still a problem? How this can impact us?

jbovenschen:
This is still an issue, and can be solved in two ways.

Main issue is that yoga-layout depends on emscripten which dependencies are not correctly installed on every OS or PC.

- `Yoga-layout` provides a precompiled version of the library
- We make a bundled release of `react-pdf` with an compiled version of `yoga-layout`

Sort of something I would like to do, providing a flat bundle like react is going to do in the future.

ferdinandsalis:
@diegomura  I just ran across this error when I deployed my app on now.
```
> micro: Error when importing /home/nowuser/src/build/index: Error: Error relocating /home/nowuser/src/node_modules/yoga-layout/build/Release/nbind.node: YGAlignToString: symbol not found
>     at Object.Module._extensions..node (module.js:598:18)
>     at Module.load (module.js:503:32)
>     at tryModuleLoad (module.js:466:12)
>     at Function.Module._load (module.js:458:3)
>     at Module.require (module.js:513:17)
>     at require (internal/module.js:11:18)
>     at initNode (/home/nowuser/src/node_modules/nbind/dist/nbind.js:141:15)
>     at /home/nowuser/src/node_modules/nbind/dist/nbind.js:115:13
>     at findCompiledModule (/home/nowuser/src/node_modules/nbind/dist/nbind.js:79:13)
>     at find (/home/nowuser/src/node_modules/nbind/dist/nbind.js:93:13)
>     at Object.init (/home/nowuser/src/node_modules/nbind/dist/nbind.js:104:5)
>     at Object.<anonymous> (/home/nowuser/src/node_modules/yoga-layout/sources/entry-node.js:11:17)
>     at Module._compile (module.js:569:30)
>     at Object.Module._extensions..js (module.js:580:10)
>     at Module.load (module.js:503:32)
[...]

PR #132, opened 2017-08-29
Add borders support

### LINKED ISSUE #127: Borders not being rendered?
Upon a cursory reading of the code I see that borders are supported. Do they only work on certain elements?
```
const CognitiveFunctionOverview = ({name}) => {
    style: {
      paddingRight: 10,
      borderWidth: 1,
      borderStyle: 'solid',
      borderColor: '#CCCCCC'
    }
  });

  return (
  <View style={styles.style}>
    <Text>
      CF1 {name}
    </Text>
  </View>
  )
};
```
No border is displayed.

### LINKED ISSUE #133: Image examples are not working
Image used in the example doesn't exist anymore. Maybe examples could be updated.

diegomura:
Fixes #127 
Fixes #133


KerenChandran:
Not sure if this was intended, but the borders are still not showing up for the other elements. I quickly looked into the code and it seems that drawBorders needs to be called on the other elements.

Not sure if you just forgot to add the call to the other elements, or if you had a different way to tackle the issue

diegomura:
Damn. You're right. I'll fix it asap

KerenChandran:
Thought you might be busy, so I just created the PR. Do you happen to have a roadmap for a 1.0 release?

diegomura:
Thank you man! I really appreciate it.
I don't have a fixed roadmap, but I think the two bigger issues we are having right now are:

- Yoga does not build in Linux (#57)
- Pdfkit not working in the browser (#101)

Once we have these two working, I will feel very confident on releasing the 1.0 😄

jmjpro:
I see the 1.0.0 milestone here: https://github.com/diegomura/react-pdf/milestone/1

Thanks for looking into this @KerenChandran. I was tackling the same problem just yesterday. I tried manually adding `this.drawBorders()` in the View component as you did I still didn't see any borders in the output. Do you have a working example?

diegomura:
@jmjpro yes. But the last two points can wait. Would be great to have them, but not it's mandatory. Feel free to contribute on the one you feel more comfortable 😄

KerenChandran:
@jmjpro @diegomura 

Something really odd is happening. On version 0.4.3, the borders still do not render. However, for my resume project (which I set to version 0.4.2 locally), if I apply `this.drawBorders()` change directly in `node_modules/@react-pdf/core/lib/elements/View.js`, the borders render as expected.

I'm going to spend the rest of today trying to understand what's going on and will report back my findings. I tested that on node 8, so not sure if there is some weird behaviour for node 8 vs node 7. Also the changes I made in 0.4.3 are not present in the build output, so something might be wrong with the build output or my last commit.

PR #142, opened 2017-09-11
Add precompiled yoga

### LINKED ISSUE #57: Yoga does not compile in Linux

diegomura:
Fixes #57


diegomura:
This is just an attempt to solve #57. Consist on shipping the library with the compiled version of Yoga. It has the build for version 6, 7 and 8 of Node, with some bindings to make it work.

To be honest, I don't fully understand everything of this topic. But this works at least. @jbovenschen  would love to have your input here

diegomura:
You can test this implementation using v0.4.5 beta version. I tried it in a blank project, using node 6, 7 and 8 and worked in each case

PR #143, opened 2017-09-12
Update travis.yml

jbovenschen:
I did test the work done in an previous PR on a local machine running linux, which made it work with yoga-layout.
So the test for linux could be re-enabled again. 

Also added node version 8


jbovenschen:
This seems to break, which is a little weird because locally I don't see it. Will investigate later what differs.

jbovenschen:
Need to disable the 'default' postinstall script for yoga-layout, will do so if I figure out the way how to do so.

jbovenschen:
It seems that I'm doing something wrong with the installation of a modern G++ library on linux, more info here; https://github.com/cbeck88/travis-cpp-assistant

PR #145, opened 2017-09-13
Update compileYoga.js

snettah:


jbovenschen:
Thank you for doing this, just curious which version of node did you use?

snettah:
7.5

Issue #152, opened 2017-10-18
Render to stream


rogierslag:
Internally react-pdf already uses readable streams to push the data into a file. For some use cases it might be beneficial to expose a regular NodeJS stream directly (instead of rendering to a file).

For our case, we are now writing to a file, reading it again and exporting it, then deleting the file. We could save a lot of overhead if `ReactPDF.render` had a `ReactPDF.renderStream` sister.

diegomura:
I like the idea!  I think the API you define is fine. Do you think you can submit a PR about it? 
I would do it, but I didn't have much time lately

PR #153, opened 2017-10-19
Support streaming output for nodejs

rogierslag:
Closes https://github.com/diegomura/react-pdf/issues/152

This adds support for streaming the PDF response. As a result it can be directly used by subsequent node calls, instead of having to write a file and reading it again.

I made the renderToStream a first class citizen, with the `renderToFile` effectively using that stream. To prevent future confusion, the `render` method directs to the `renderToFile` method. However since it may confuse people, that method might need deprecation/removal in the future.

### DIFF
--- a/packages/react-pdf-node/src/index.js
+++ b/packages/react-pdf-node/src/index.js
@@ -1,15 +1,18 @@
 import fs from 'fs';
 import path from 'path';
-import { PDFRenderer, createElement, pdf } from '@react-pdf/core';
+import { createElement, pdf, PDFRenderer } from '@react-pdf/core';
 
 export default {
-  async render(element, filePath, callback) {
+  async renderToStream(element) {
     const container = createElement('ROOT');
     const node = PDFRenderer.createContainer(container);
 
     PDFRenderer.updateContainer(element, node, null);
 
-    const output = await pdf(container).toBuffer();
+    return await pdf(container).toBuffer();
+  },
+  async renderToFile(element, filePath, callback) {
+    const output = await this.renderToStream(element);
     const stream = fs.createWriteStream(filePath);
     output.pipe(stream);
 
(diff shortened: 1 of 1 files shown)


rogierslag:
No idea why Travis does not start

packages/react-pdf-node/src/index.js
```diff
@@ -1,15 +1,18 @@
...
     const container = createElement('ROOT');
     const node = PDFRenderer.createContainer(container);
 
     PDFRenderer.updateContainer(element, node, null);
 
-    const output = await pdf(container).toBuffer();
+    return await pdf(container).toBuffer();
+  },
+  async renderToFile(element, filePath, callback) {
+    const output = await this.renderStream(element);
```
diegomura: Should be `renderToStream` here right?

rogierslag: Whoops, last minute change in a signature is never a good idea. Good catch!

rogierslag:
Amended my commit, so this should be fixed.

I removed the deprecation comment as well. For convenience I left the "main method" `renderToFile` for clarity, but everybody can also just use `render`

diegomura:
Agreed. I restarted the Travis build to see if works now. Thanks!

rclai:
I feel like this might be a stupid question, but if I'm using Express, how do I pipe this to the response object?

Doing `ReactPDF.renderToStream(MyPdf).pipe(res)` doesn't work. Is it because the return value is not actually a stream but a Buffer?

rogierslag:
You can do the following

```javascript
// When you can use async/await
const stream = await ReactPDF.renderToStream(MyPdf); // As it actually returns an async function
stream.pipe(res);

// When bound to promises with ES6
ReactPDF.renderToStream(MyPdf).then(stream => stream.pipe(res));
```

rclai:
Actually, I realized I was doing something wrong. I had to do:

```
await ReactPDF.renderToStream(<MyDocument />).pipe(res)
```

Issue #119, opened 2017-07-08
Support for local file in Image component


theobat:
There should be support for local filepath when using src attribute of the Image component. Althouth it seems very simple to do I didn't go straight to a PR because the core package is shared between node and the browser, so directly using node packages such as path and fs might not be an option as far as I understand.

We could go with something like:

```
// pseudo code
if (src.startsWith('file://')) {
  const fs = require('fs')
  return fs.read(src)
}

```
@diegomura what are your thoughts ?

diegomura:
Thanks @theobat !
You are right about we must have support for this feature.
Luckily for us, `pdfkit` already has support for local image files, so there is less to do on our side (http://pdfkit.org/docs/images.html).
I would avoid the `file://` prefix and just stick with urls and paths. If it's a url we fetch the image and add it to the doc (using `http` as the default protocol if not specified), and if it's a path, we pass the path to `pdfkit` and let him read it from the file system. What do you think? Would you like to upload a PR with this feature? That would be great

theobat:
All right I'm gonna do this

rogierslag:
Any updates on this?

PR #156, opened 2017-10-22
Add the ability to pass an image as a Buffer

rogierslag:
Closes https://github.com/diegomura/react-pdf/issues/119

This allows a consumer of the `react-pdf` API to directly use local files for images.

We use it in the following way:

```jsx
// outside of the component so it happens just once instead of on each render
const logo = fs.readFileSync(`${__dirname}/logo-6d10589dc3bb992cf77d8ede61131996.png`);

// inside the render method
const logoData = {data: logo, format: 'png'};
<Image src={logoData} style={styles.image} />
```

Apart from being really useful for node processes (maybe less in the browser), this significantly improves the speed of rendering. When rendering 1000 documents over 16 threads (which all have our logo), a average render might take around 1s to complete. Most of this is network IO to download the same logo over and over again from the CDN. It takes quite a number of sockets as well.

By moving to a local file, 1000 renders over 16 threads we got the following benchmark

In both cases: we'd warm the server by doing 100 identical requests. The server uses 16 threads. The benchmark is performed with 25 concurrent request. After the warm up, we give the server a few seconds, then send 1000 requests with 25 concurrent requests. We always use the production builds for both tests. The document contains a single image, weighing 22k

Over the network
```
Percentage of the requests served within a certain time (ms)
  50%    916
  66%   1054
  75%   1145
  80%   1283
  90%   1652
  95%   2028
  98%   2459
  99%   2847
 100%   4048 (longest request)
```

Locally
```
Percentage of the requests served within a certain time (ms)
  50%    488
  66%    563
  75%   1063
  80%   1310
  90%   2032
  95%   2684
  98%   3624
  99%   4196
 100%   4812 (longest request)
```

As one can see, in this case we achieve a 2x performance improvement (and its better to not depend on the external resources as well from a code quality standpoint)


diegomura:
This is just awesome! Thank you for tackling this, and also for taking the time to do a benchmark to illustrate the improvement.
Do you mind adding at least one spec that test this feature? I can add it if you want.

btw. I'd love to hear more about your usage of the lib. I love to hear about success cases!

rogierslag:
I've added a test. However when these run then change all the indenting (breaking it for travis probably) so I did not update the snapshots themselves (I'm using node 8.5 btw).

Could you push the correct snapshot in a separate commit?

rogierslag:
I'd love to fix the conflict here, but that makes the merging even worse

diegomura:
I screw it up. I had to do a manual merge. It was a nightmare haha. But these changes are now on master, which is the important thing. Sorry about that

PR #157, opened 2017-10-22
Allow metadata setting for a document

rogierslag:
This allow one to set the pdf metadata again. See the example how to.


diegomura:
Thanks man! I'm going to write some tests for this if you are ok and then merge it.

packages/react-pdf/src/elements/Document.js
```diff
@@ -20,6 +27,23 @@ class Document {
...
+    }
+    if (author) {
+      this.root.info.Author = author;
+    }
+    if (subject) {
+      this.root.info.Subject = subject;
+    }
+    if (keywords) {
+      this.root.info.Keywords = keywords;
+    }
```
diegomura: Are these 4 the only type of metadata possible? (I don't actually know 😅 )
Also. is there a problem if some metadata keeps set as`null` value? I'm asking if we can get rid of the `if`s statements, or maybe replacing them for `title || undefined`. Maybe I'm being a bit nitpicking here haha sorry

diegomura:
I added some comments 😄

packages/react-pdf/src/elements/Document.js
diegomura: Don't worry. I answer myself 😅 . Didn't worked with the way I suggested

diegomura:
@rogierslag I added some other changes. Hope you don't mind 😄

PR #161, opened 2017-11-23
Add .nvmrc file

### LINKED ISSUE #158: Add an `.nvmrc` file
Currently it is a bit unclear on which node version you develop. I just tried using 8.5 but that messes up all the indenting in snapshots.

Adding an `.nvmrc` file would allow everybody to run `nvm use` to select the correct version

diegomura:
Fixes #158


diegomura:
@jbovenschen I removed the engines keys from the `package.json`s because travis was complaining about something. It's not as critical right now. So it will be just the `.nvmrc` file right now, which I think is cool

jbovenschen:
Ok! Seems good to me.

Issue #139, opened 2017-09-06
enhancement idea: support css media queries


jmjpro:
Enhancement idea: the Page component should support the CSS Paged Media Draft Specification. See
- https://www.w3.org/TR/css3-page/
- http://caniuse.com/#feat=css-paged-media and -
- https://www.smashingmagazine.com/2015/01/designing-for-print-with-css/

diegomura:
@jmjpro I don't see how the `@page` css property should work for the `Page` component. `@page` is used to define how an html document should be printed right? In PDFs, what you print is what you have.

However, I do see great benefit on supporting `@media` of components in general.

Please let me know if there's something I'm missing

jbovenschen:
Could it beneficial to support a `onLayout` property like react-native does for `View`, so you could derive the width calculated by `yoga-layout`. If that works you could create your own abstractions for `@media` and `@page`. 

The hard part will be the node renderer, which should wait until the page is fully rendered.

diegomura:
I still don't see what we can do with `@page` 😅 .
`onLayout` would definitely be a cool feature, but do we need it to support `@media`? From what I can see so far, if we can somehow pass page attributes (width, height and orientation at least) down to the `Stylesheet` declaration (in layouting time), we can infer specific styles for a component using media-queries.

Last night I started this project https://github.com/diegomura/media-queries to have a media-queries engine in case we need it in the future

jbovenschen:
If we wan't to support the @media in the StyleSheet declarations, we don't need onLayout. We can just derive the "viewport" from the Page instance.

diegomura:
Exactly. That's why I'm building that "media-queries engine". With that, it should be pretty straightforward. 
I still don't see how `@page` (which as far as I know are styles for printing) can have any relevance in the project.

jbovenschen:
I neither see any need for the @page at this moment we should be ok, to just support @media and onLayout at this moment? Should I try to get onLayout implemented correctly.

diegomura:
That would be great. 
I'm not sure yet how the media-query API will work yet. Im not a big fan of defining object keys as strings (`{ '@media(whatever)': { styles }`). At least not as a public API. 

Any ideas?

jbovenschen:
Could maybe support nested values within `Stylesheet.create` thing is we should document it after we finish the documentation site. You could take a look within glamorous how it would look like.

diegomura:
[...]

PR #166, opened 2017-12-21
Page wrapping

### LINKED ISSUE #149: Page wrapping
Do I understand correctly that react-pdf currently does not attempt to do any page wrapping, that is, if the content overflows a page, I have to catch that myself and move it to a second page, correct? Any plans regarding this?

diegomura:
Fixes #149


packages/react-pdf/src/elements/Page.js
```diff
@@ -1,68 +1,63 @@
...
-        this.layout.setWidth(size[0]);
-      }
-    }
+    this.children.push(newSubpage);
   }
 
-  async render() {
-    const { orientation } = this.props;
+  addNewSubpage() {
+    const originalSubpage = this.children[0];
```
tpreusse: unify naming, `initial` sound better

packages/react-pdf/src/elements/Page.js
```diff
@@ -1,68 +1,63 @@
...
+    this.props = { ...Page.defaultProps, ...props };
+    this.children = [];
+
+    this.addInitialSubpage();
   }
 
-  applyProps(props) {
-    super.applyProps(props);
+  addInitialSubpage() {
+    const newSubpage = new SubPage(this.root, this.props);
```
tpreusse: I would put this on the instance instead of using `this.children[0]` everywhere. Makes the shared nature of the children array also a bit more transparent.
`this.initialSubpage = new SubPage(this.root, this.props`

packages/react-pdf/src/elements/Base.js
```diff
@@ -171,9 +191,40 @@ class Base {
...
+  async renderWrapChildren(page) {
+    const renderedChilds = [];
+    let availableHeight = this.parent.getHeight();
+
     for (let i = 0; i < this.children.length; i++) {
-      await this.children[i].render();
+      const child = this.children[i];
+      const childHeight = child.getHeight();
+
+      if (availableHeight >= childHeight) {
```
tpreusse: will we get a hook here? 😀

diegomura: yes!

flaurida:
Hi Diego,

Thanks for the fix! Is there any way to render something like `Page 1 of 4` at the bottom of wrapped pages now, if we do not know in advance how many pages there will be? Thanks again!

Sincerely,
Laura

diegomura:
No, there isn't. react-pdf current implementation does not know in advance how many pages the document will have either

flaurida:
No worries - is there possibility of this happening in the future? Thanks again!
 
Sincerely,
Laura

flaurida:
I also wonder if it is possible to do something with stylesheets like the below? Please let me know!

flaurida:
https://stackoverflow.com/questions/20050939/print-page-numbers-on-pages-when-printing-html

diegomura:
Not for the moment. With the current page wrapping implementation it would be a bit difficult to do, because the algorithm does not know how many pages will be in advance. Without changing too much, this should be done maybe in a second rendering run, but the implementation is not that straight forward. 

The tricky part is the `of x`. The most easy step we can do in that direction is rendering just the page number. I tried adding somethink like this, the renederer didn't like functions as childs, and then I never tried again:

```jsx
<Text>
  { ({ page }) => `page ${page.number}` } 
</Text>
```

About the `@page` metatag, that's also out of scope. We discuss about that these past days in here #139 . It doesn't make much sense `@page` in a PDF ecosystem, since that are styles for printing. In PDF, you print what you get. We cannot add things for printing on the fly. Not related, but there is some progress in start supporting `@media` thought 😉

flaurida:
Cool - down the road do you think it would be possible to support this functionality? Perhaps I can help work on it? Thanks again Diego 😊

diegomura:
@flaurida Do you mean having `Page x of y` or just `Page x` ??
Subtle, but the latter will be extremely easiest.

flaurida:
We can start with `Page x` perhaps then before having `Page x of y`? In addition, can I use the `fixed` property on a view or only text when creating my page headers and footers?

diegomura:
Cool. The API I would like for rendering dynamic content on the document is the one I already write:
```jsx
<Text>
  { ({ page }) => `page ${page.number}` } 
</Text>
```
However, I got some issues when trying to do it. Didn't do an extensive research though. You can start from there and see if for other API alternatives if that doesn't work.

Regarding `fixed`, you can use it, but not sure how the results will look like 😅

flaurida:
Is there any way to get nice wrapping besides using `fixed`? 😅

diegomura:
`fixed` is for re-rendering things in all pages. not sure what you want to do exactly in this case 😅

flaurida:
I have a header that includes an image and text that I would like to render on all pages, can I group the image and text together in a `View` and then apply `fixed`? Or is there a better way to do that?

flaurida:
In addition, I am having issues getting different view elements to wrap in a flexbox row when inside a page with the `wrap` component. Am I doing something wrong?

diegomura:
@flaurida That's the correct way of doing it!
Not sure about `wrap` issues. I did observe some obscure flexbox use cases that wasn't working as expected. If you like, you can create a new issue with a simple example of how to replicate it and we can discuss about that there

flaurida:
Yes - that is the issue - I cannot wrap something after it is inside a page. Will create a new issue. Thanks!

Sincerely,
Laura

flaurida:
In terms of the View with `fixed` though, it is not being rendered on every page. I am also seeing an extra page being generated.

flaurida:
Do I also need to apply `position: 'absolute'` to anything `fixed`?

flaurida:
I filed my issue here: https://github.com/diegomura/react-pdf/issues/206, thanks again!

diegomura:
That's odd. Please check the `pageWrap` example for how it's used. Page wrap is a quite new feature so it can fail, specially with complex layouts. It was originally for a simple top to bottom layout. For 2 columns layouts it may fail. The idea is to support it though.
I'll check your issue . thanks

flaurida:
Thanks - if at all possible can we make it support 2 columns? Perhaps there is a workaround? Can I help fix the code??

diegomura:
It's the idea, but I didn't estimate not know what involves. Wrapping however is on the top of my priority list. If you want to jump into the code, please go ahead! I can clarify any question you might have

flaurida:
As in wrapping and supporting multiple columns? Do you have a general idea of what needs to be done would be my first question?

diegomura:
Not something in particular. If you go to the code you will see that the algorithm tries always to fill in the remaining vertical space of the document, and add a new page when no more place is available.

I don't see clearly how your two columns layout is. Are they two separate columns? or you expect the wrapping algorithm to jump from column 1 to 2?

flaurida:
No, I just want to have a `View` with a `flexDirection` of row without the second item not rendering. Each column is half the width of the page, so it shouldn't need to wrap. Did you see the simple example I wrote in the issue? Thanks so much for being so responsive 🎉

diegomura:
Not yet! haha sorry. busy work day. You can try using fixed widths in the meanwhile, or percentage widths. I'll try to see the issue asap

flaurida:
No worries, I am already so grateful for your help! Yes, I am using percentage widths. If I have two `View` with percentage widths of say 30%, will they automatically go next to each other in a row? Or do I need to use flex? Thanks again Diego 😊

diegomura:
As far as I remember, they should just go next to each other in a row. Just like the web. Just in case, all elements in `react-pdf` has `display: flex` as default.

I'm glad to help! Thanks also for your ongoing effort of using and improving the library. Sorry if my answers are vague, but page wrapping and layouting are two very complex issues, and I'm barely supporting them. I plan to tackle them separately, and considering more use cases in the future

flaurida:
For some reason the two are not going next to each other in a row. I will try to play around with it. Hopefully together we can at least fix the immediate issue of items not going next to each other. Thanks again!

PR #170, opened 2018-01-17
Fix core and dom packages

langovoi:
This PR make `@react-pdf/dom` working again.

**How launch dom-bindings example?**

1. `yarn` (in the root of project)
2. `cd examples/dom-bindings`
3. `yarn run start`

**Key points to use `@react-pdf/dom`:**
1) transform `pdfkit` with Browserify (https://github.com/devongovett/pdfkit/issues/478#issuecomment-322664148);
2) use `babel-polyfil` to fix `regeneratorRuntime is not defined`, which causes by `@react-pdf/*;

To make `dom-bindings` working, I used `react-scripts eject`, because I was needs to customize webpack config (for first key point).

**Notes:**
Sure it is not good way to fix `@react-pdf/dom`, but I think progress searching better solution is in right way.
I found [fork of pdfkit](https://github.com/jerp/pdfkit), which in progress removing of CoffeeScript and rewriting to ES2015+. I think it will help to use pdfkit in browser in right way.


diegomura:
Thanks man! This is awesome!
I just merged a small PR and messed up this one. Going to test it and resolve those conflicts.
This solves #101 then right?

PR #175, opened 2018-01-18
Don't regenerate examples every time.

jbovenschen:
Because it causes always merge conflicts. When there are multiple PR's being merged.

Should we also add the output dirs to the .gitignore?


diegomura:
@jbovenschen this wouldn't be my first shot. What about finding the way of generating but if nothing changed on that example, don't be recognized as a conflict (same data)?

That's what I meant here: https://github.com/diegomura/react-pdf/pull/172#issuecomment-358642248

Examples are another way of ensuring you didn't break anything.
Maybe if we mock the date en examples run, the result will always be the same and it will not be recognized as a conflict. If we cannot do that, then I'll merge this

diegomura:
@jbovenschen I changed my mind. I try and couldn't achieve what I wanted. Let's take it away 😉 
Thanks man!

PR #194, opened 2018-02-06
Add fontkit as dependency

### LINKED ISSUE #193: Fontkit dependency?
In https://github.com/diegomura/react-pdf/compare/v0.5.1...v0.6.0 fontkit was removed as a dependency. However it seems required to properly load fonts into react-pdf with `Font.register`. Without fontkit the fonts are not loaded and the PDF is not rendered. A simple `yarn add fontkit` in out project fixed it though.

I could not find why it was removed, while it is still part of the webpack loaders.

diegomura:
Fixes #193


rogierslag:
Wow thanks for fixing this fast!

Issue #196, opened 2018-02-09
Warning when `styles` is passed instead of `style`


diegomura:
**OS:**
Any

**React-pdf version:**
Latest

**Description:**
We should display a warning when a prop `styles` is passed instead of `style`

**How to replicate issue (if applies):**
Pass a `style` prop and see the warning in the console

rogierslag:
I thought this was unnecessary but as it turned out I spend 40 minutes cursing on flexbox until I realized I did exactly this.

PR #197, opened 2018-02-11
Warn when styles is passed instead of style

rogierslag:
Closes https://github.com/diegomura/react-pdf/issues/196

So this is less likely to happen in the future ;)

Additionally I fixed two files which needed a reformat according to prettier


diegomura:
Thank you very much for this @rogierslag !! I really appreciate it. Going to merge this

Issue #199, opened 2018-02-19
Re-rendering PDF fails on commitTextUpdate


ottosichert:
**OS:**

all

**React-pdf version:**

latest, v0.7.1, c900af77fbf3c8cf682683c823387fd2af5f37e3

**Description:**

- Render any PDF, e.g. with
  ```
  import ReactPDF from '@react-pdf/node';
  // somewhere
  ReactPDF.renderToStream(app);
  ```
- Update props of any `<Text>` element
- Re-render PDF
- **=> crashes**

<details>
<summary>Stacktrace</summary>

```
The above error occurred in one of your React components:
    in TEXT (created by FooComponent)
    in VIEW (created by BarComponent)
    [...]
    in PAGE (created by BazComponent)
    in DOCUMENT (created by AnyComponent)
    [...]
    in App

Consider adding an error boundary to your tree to customize error handling behavior.
Visit https://fb.me/react-error-boundaries to learn more about error boundaries.

  TypeError: Cannot create property 'chidren' on string '<updated string>'
      at commitTextUpdate (/home/ottosichert/<project_path>/node_modules/@react-pdf/core/lib/renderer.js:98:28)
      at commitWork (/home/ottosichert/<project_path>/node_modules/react-reconciler/cjs/react-reconciler.development.js:5059:11)
      at commitAllHostEffects (/home/ottosichert/<project_path>/node_modules/react-reconciler/cjs/react-reconciler.development.js:5706:13)
      at Object.invokeGuardedCallback$1 (/home/ottosichert/<project_path>/node_modules/react-reconciler/cjs/react-reconciler.development.js:1562:10)
      at invokeGuardedCallback (/home/ottosichert/<project_path>/node_modules/react-reconciler/cjs/react-reconciler.development.js:1509:29)
      at commitRoot (/home/ottosichert/<project_path>/node_modules/react-reconciler/cjs/react-reconciler.development.js:5801:9)
      at performWorkOnRoot (/home/ottosichert/<project_path>/node_modules/react-reconciler/cjs/react-reconciler.development.js:6803:42)
      at performWork (/home/ottosichert/<project_path>/node_modules/react-reconciler/cjs/react-reconciler.development.js:6753:7)
      at requestWork (/home/ottosichert/<project_path>/node_modules/react-reconciler/cjs/react-reconciler.development.js:6664:7)
[...]

PR #200, opened 2018-02-19
Fix typo in commitTextUpdate()

ottosichert:
Fixes diegomura/react-pdf#199


diegomura:
Thanks. Yes, make sence since the `createTextInstance` method returns the string, and not an instance of something.

PR #205, opened 2018-03-01
Better border support

maxaggedon:
This PR brings a small improvement for border support.

When there was no border radius, the different borders of a `<View />` would not join correctly and leave blank space in the corner (see [before.pdf](https://github.com/diegomura/react-pdf/files/1771332/before.pdf)). This now fixed (result [after.pdf](https://github.com/diegomura/react-pdf/files/1771333/after.pdf)).

There are still some problems with border radius support I guess (different widths on adjacent sides, clipping content...). I will try to work on those later if I need to use border radius.

### DIFF
--- a/packages/react-pdf/src/mixins/borders.js
+++ b/packages/react-pdf/src/mixins/borders.js
@@ -64,8 +64,14 @@ const Borders = {
 
     // border top
     this.drawHorizontalBorder(
-      [left + borderTopWidth / 2, top + borderTopWidth / 2],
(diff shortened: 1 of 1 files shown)


diegomura:
Thanks @maxaggedon !!
Yes,  I wish I could put more effort in the borders feature, specially when rendering radius. I was aware of issues in the radius, but not this one. Thanks for fixing it!

maxaggedon:
Thank you for the quick merge ! `react-pdf` is a really good initiative.

PR #207, opened 2018-03-07
Check image extension with buffer

YozhikM:
The previous check was weak and did not check the actual extension (format) of the file. Now the test is done using a buffer.

This solves the problem when the image extension specified in the title does not match the actual format.


diegomura:
Thanks @YozhikM ! 
Sorry for the delay. I didn't have much time to test it before merging it, but I will this week 😄

YozhikM:
TX! @diegomura

diegomura:
Thanks to you!

PR #214, opened 2018-03-29
Popup mode for dom

bkniffler:
This will open the pdf in a new window and call onDone. This could also be a different package react-pdf-popup or something.

### DIFF
--- a/packages/react-pdf-dom/src/Document.js
+++ b/packages/react-pdf-dom/src/Document.js
@@ -21,7 +21,7 @@ class Container extends Component {
     this.mountNode = PDFRenderer.createContainer(this.container);
 
     PDFRenderer.updateContainer(
-      <Document {...omit(['height', 'width', 'children'], this.props)}>
+      <Document {...omit(['height', 'width', 'children', 'popup', 'onDone'], this.props)}>
         {this.props.children}
       </Document>,
       this.mountNode,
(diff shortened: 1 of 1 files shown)


bkniffler:
The point is, writing pdf with react is really cool, but currently, on the client side of a business application (offline-first with pouchdb, electronjs), I need to ensure the pdf is rendered somewhere where it takes a good amount of screensize (e.g. mobile), that it can be closed, printed, etc.

Sometimes its just easier to pop the document open into a new browser window (which works nice on mobile, electron and web) instead of overlaying a modal or creating a specific pdf view on your app. Electronjs has got known issues with embedded pdf rendering (no print option). Embedding pdf on mobile is difficult UX-wise.

I could of course create an API /pdf on my server that renders everything server side, but then I need to find a way to fetch the data needed for rendering, also it doesn't work offline.

jbovenschen:
My thoughts on this particular issues;

It is nice to have some way to create the pdf's within the background without rendering it directly to the dom, use-cases for this would be a download button for example.

I'm into changing the Document Component in something which would look like this.
```js
<Document
   render={({ blob }) => (
     <a target="_blank" href={URL.createObjectURL(blob)}>Download</a>
  )}
>
   {/* PDF specific components */}
</Document>
```

This would solve your use-case right?

PR #222, opened 2018-04-24
Change text function API

diegomura:


jbovenschen:
Little bit late, but seems reasonable, so hyped for the 1.0.0 release

PR #256, opened 2018-07-15
Fixes rendering of nested Text elements.

johanventer:
Nested Text elements (to apply different formatting, i.e. a single bold word) were rendering on top of each other due to the Text element not respecting the value of `inline`. This PR resolves the issue for me.

### DIFF
--- a/packages/react-pdf/src/elements/Text.js
+++ b/packages/react-pdf/src/elements/Text.js
@@ -132,8 +132,9 @@ class Text extends Base {
   //   return this.children.length;
   // }
 
-  async renderText(text, isFirstNode) {
+  async renderText(text) {
     const {
+      color = 'black',
       textAlign = 'left',
       align,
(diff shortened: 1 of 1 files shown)


diegomura:
Thanks @johanventer ! 
PRs are always very appreciated.
Is it mandatory for you to merge this? Let me explain why I ask for this: I'm putting lot of effort on releasing a `v1.0.0` very soon, that will ship plenty of new features, and the whole text implementation changes. I don't have much time to test this PR and ensure it does not break any use case, and the few I have I prefer to use it in this new version. 
So if it is ok for you, I'll keep this open for the moment. Unless something unpredictable happens with the new version, I'll have it ready pretty soon

johanventer:
@diegomura No problem at all - I linked to this PR from the relevant Issue so anyone looking for a solution can find it. Looking forward to 1.0.0!

PR #257, opened 2018-07-19
1.0.0

diegomura:


examples/images/index.js
```diff
@@ -3,52 +3,53 @@
 
 import React from 'react';
 import ReactPDF from '@react-pdf/node';
-import { Document, Page, Text, Image, StyleSheet } from '@react-pdf/core';
+import { Document, Page, Text, Image, StyleSheet, Font } from '@react-pdf/core';
```
bkoltai: Should this be imported from `@react-pdf/react-pdf`?

diegomura: Yes. Old /core and /dom packages are obsolete now

PR #262, opened 2018-07-24
Prepare react-native build

### LINKED ISSUE #34: Add react-pdf-native
Make it possible to render pdf's inside react-native.
Same way as the react-pdf-dom package works.

diegomura:
Fixes #34


geoffreym74:
I'm curious about the status of this PR.  I am interested in this capability and wondering what it would take to get it working.

diegomura:
Hi @geoffreym74 !

To be honest there is not much progress about this. I remember having conflicts with some of the object react-pdf uses (not sure if it was Buffer, or with streams). It's definitely on the roadmap, but had to put pause it in favor of other more critical stuff.

Help would be very welcomed if you feel to give it a shot! Besides that the rest should just run in native

ybbond:
> To be honest there is not much progress about this. I remember having conflicts with some of the object react-pdf uses (not sure if it was Buffer, or with streams). It's definitely on the roadmap, but had to put pause it in favor of other more critical stuff.

I think the problem is with `blob-stream` dependency, as this failure happens to me.
screen shot 2019-02-11 at 05 21 19

luizcieslak:
@diegomura I want to try solving this by using [rn-fetch-blob](https://github.com/joltup/rn-fetch-blob). Are you ok with this approach?

ybbond:
> @diegomura I want to try solving this by using [rn-fetch-blob](https://github.com/joltup/rn-fetch-blob). Are you ok with this approach?

goodluck!
and also ... bump

mitchellmcm27:
This is because the core libraries for Node aren't available in React Native. I was able to get around these errors using [node-libs-react-native](https://github.com/parshap/node-libs-react-native), which provides `stream`, `Buffer`, etc. Maybe this will push the PR forward?

PR #274, opened 2018-08-05
Fixed module path in package.json

maxsbelt:
If you import package as `es` module you'll receive `cjs` module version instead of `es`. Most likely it's just a typo in module path definition in `package.json`. This PR solves this issue:

screen shot 2018-08-05 at 16 01 18


diegomura:
Thank you! My bad 😅

PR #279, opened 2018-08-08
Update Resume Text and Use .map for Rendering ExperienceEntry Components

mattstates:
I made some spelling changes to fit with the theme for this mock resume. I also extracted the ExperienceEntry prop info into an array of objects and used .map to to render the components.

### DIFF
--- a/examples/resume/Experience.js
+++ b/examples/resume/Experience.js
@@ -75,49 +75,64 @@ const ExperienceEntry = ({ company, details, position, date }) => {
   );
 };
 
+const experienceData = [
+  {
+    company: 'Jedi Temple, Coruseant',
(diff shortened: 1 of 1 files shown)


mattstates:
As a Star Wars fan and someone looking to make more open source contributions, I thought I could help out here. Please let me know if I need to address anything. Thanks!

diegomura:
If you ever want to contribute to react-pdf just ping me that we might find something for you to do!

mattstates:
Thanks Diego. I will let you know. I want to be careful about committing to more than I could deliver on.

diegomura:
Makes sense. It's a clever thought haha. I've been in the other side and things can get stressful. Just let me know!

PR #285, opened 2018-08-11
Shallow rendering and live-reloading

### LINKED ISSUE #84: Generate PDF on the fly
Thanks for a great library I was playing around with something similar recently.

This is what I am trying to solve currently and wondering if it is/will be possible with react-pdf:
I need to be able to generate loads of different PDFs in our platform and would love to do that fully on the frontend. The user clicks `download` button which allows him to download a generated PDF by something like `react-pdf`?

### LINKED ISSUE #208: Unable to use `Document` `title` prop for filename of saved PDF
I am using react-pdf to generate PDF files directly on the client to avoid unnecessary network requests. I am using `@react-pdf/dom` and the PDFs are rendered successfully in an `iframe`. However, I want my users to be able to download the PDF and I want to be able to specify the name of the file when it is downloaded. I couldn't figure out how to setup the PDF such that when they click the download link in the `iframe`, the save dialog uses a defined filename. Am I missing something?

I was able to get my desired output by bypassing the iframe rendering and instead using the `blob` [here](https://github.com/diegomura/react-pdf/blob/master/packages/react-pdf-dom/src/Document.js#L33) and used it to trigger an `a` tag with `src` and `download` properties set. This does the trick:

```javascript
pdf(this.container)
      .toBlob()
      .then(blob => {
        const data = URL.createObjectURL(blob);
        const link = document.createElement('a');
        link.href = data;
        link.download = this.props.title || "output.pdf";
        link.click();
        setTimeout(() => {
          // For Firefox it is necessary to delay revoking the ObjectURL
          URL.revokeObjectURL(data)
        }, 100)
      });
```

I'm hoping to contribute this feature to the library, as I think others would find it useful to have the option to access the PDF blob directly so they can perform this type of download.

My thought is that `@react-pdf/dom` might be better of exposing a `DocumentViewer` object (which is what the current `Document` element is in my mind) and maybe a `DocumentDownloader` that works as I described above. To me, it's a bit surprising that the `Document` element is actually an `iframe`. It might make more sense for it to be a `Blob` that represents the generated PDF file.

Thoughts?

### LINKED ISSUE #236: Trying to implement live reloading
Hello! Thanks for the good work on this lib.

We try to implement the live reload in our project. We got inspiration from the repl site of React PDF, but we are stuck in the last step which is showing the blob file in the browser.
When we opened the blob url (for instance one of them is blob:http://localhost:8080/24c0232a-88f3-4042-b746-2e69072bad73) in the browser, we see the expected PDF displaying, but in our component when we use :
```javascript
<Document file={this.state.document} />
```
with document the blob url file, it just fail silently, no warning or error and checking the DOM of our page, we see an empty div : 
```html
<div class="ReactPDF__Document"></div>
```
When we try the same code on the cloned repository of your site, it works perfectly.
What do you think is the problem in our project? Does it have to do with the webpack config to load the blob? We also tried to install `url-loader` and add the following rule :
```javascript
{
    test: /\.pdf$/,
    loader: 'url-loader',
}
```
without success (which is not surprising since the blob url doesn't finish with .pdf)

Thank you!

### LINKED ISSUE #253: <View fixed> is not rendered properly
**OS:**
macOS High Sierra 10.13.5

**React-pdf version:**
0.7.6
1.0.0

**Description:**
_0.7.6_
`<View fixed>...</View>` does not have any effect. The element is only visual on the first page.
_1.0.0_
`<View fixed>...</View>` is displayed on every page, however, on every page the fixed content is indented further.

1st page
screen shot 2018-07-03 at 13 37 50

2nd page
screen shot 2018-07-03 at 13 37 54

3rd page
screen shot 2018-07-03 at 13 37 57

4th page
screen shot 2018-07-03 at 13 38 00

```
...
  <Document>
    <Page style={styles.body} wrap>
      <View fixed>
        <Text style={styles.header}>
          ~ Created with react-pdf ~
        </Text>
      </View>
      ...
    </Page>
  </Document>
...
```

### LINKED ISSUE #292: dynamic View?
I've been playing around with react-pdf and so far I'm quite happy 😊 One thing that would have been useful is a `render` prop on `View` like it's available for `Text` elements for dynamic rendering based on the current page. As far as I can see, it's not possible now to have a picture or box with borders that doesn't appear on the first page, or did I miss something?

### LINKED ISSUE #296: Footer Text gets a margin when put inside a view and left aligned.
**OS:** Ubuntu Linux 16.04

**React-pdf version:** 1.0.0alpha15

**Description:** When designing a more complex footer i tried using a `View` with `Text` inside. I noticed that starting from page 2, the text gets a margin that grows according to the page its displayed.

Go to https://react-pdf.org/repl change the `pageNumber` style to this:
```js
pageNumber: {
    position: 'absolute',
    fontSize: 12,
    bottom: 15,
    left: 20,
    right: 20,
    textAlign: 'left',
    color: 'grey',
    paddingLeft: 2
  },
```
And change the page indicator to this: 
```js
<View style={styles.pageNumber} fixed>
      	<Text  render={({ pageNumber, totalPages }) => (
        `${pageNumber} / ${totalPages}`
      	)} />
      </View>
```

Page 2 and beyond the text goes more to the right of the page

### LINKED ISSUE #300: Page wrap not working landscape
Hi, I'm trying to buil a pdf with dynamics pages in orientation="landscape", the problem is that the tag wrap doesn't work. Here is my code:

`      <Page orientation="landscape" wrap>
                <View style={{padding: 10}}>
                {this.state.tableHeaderPdf}
                <View style={{flexDirection: 'row'}}>
                  {this.state.tablePdfOre}
                  {this.state.tablePdf}
                </View>
                </View>
            </Page>
`

diegomura:
## Roadmap
- [x] Text layout 
- [x] Make `pdf` method part of public API (see [here](https://github.com/diegomura/react-pdf/issues/84#issuecomment-419526265))
- [x] Add documentation of new API
- [x] Add new node and dom api tests
- [x] Add background tests
- [x] Add border tests
- [x] Fix resume example
- [x] Show error when font missing

## Fixes
Fixes #84 
Fixes #208
Fixes #296 
Fixes #253 
Fixes #236 
Fixes #292 
Fixes #300


bkoltai:
Hi @diegomura, excited to see you starting on this feature; I was actually about to start an API proposal, but you beat me to it here! I'll sketch it here instead to hear and could work on a POC if you're onboard. 

How about using render props to pass the blob to children who can decide what to do with it? Something like (untested, very-pseudo code 😄 )

```javascript
import React from "react"
import {BlobProvider, Document} from "@react-pdf/renderer"

const MyPdf = () => (
  <Document>
    ...
  </Document>
)

const DownloadLink = () => {
  return (
      <BlobProvider document={<MyPdf />}>
        {blob => {
          const url = URL.createObjectURL(blob)
          return <a src={url}>Download</a>
        }}
      </BlobProvider>
  )
}
```
Or even
```javascript
import { DownloadLink } from "@react-pdf/renderer

const DownloadLink = () => {
  return (
      <BlobProvider document={<MyPdf />}>
        {DownloadLink}
      </BlobProvider>
  )
}
```

The `shallow` functionality could be implemented with this pattern with something like
```javascript
import { IFrameViewer } from "@react-pdf/renderer"

const PdfModal = ({isOpen}) => {
  return (
      <BlobProvider document={<MyPdf />}>
        {blob => isOpen ? <IFrameViewer blob={blob}/> : null}
      </BlobProvider>
  )
}
```

Essentially, the `BlobProvider` would handle the actual PDF rendering, while the `Document` handles building the actual PDF document. As I work with the library, I typically have that idea in my head but then get surprised when the Document is also the `iframe` once actually rendered. In my mind, this is the `Document` doing more than one thing, which I believe may have caused some challenges with the API. If we make more explicit and SRP components, it may make things easier to use.

I'd love to hear your thoughts and apologies if this is on a completely different track from the project, I am likely missing lots of context 🙈 

ping #236

bkoltai:
To extend this to the use case [here](https://github.com/diegomura/react-pdf/issues/84#issuecomment-412300860)

```javascript
import { DownloadLink, IFrameViewer, BlobProvider } from "@react-pdf/renderer"

const App = () => (
  <BlobProvider>
    {({ blob }) => (
      <div>
        <IFrameViewer blob={blob} />
        <DownloadLink blob={blob} title="document.pdf" />
      </div>
    )}
  </BlobProvider>
)
```

diegomura:
Thanks @bkoltai . I'll answer in the PR so anybody can see it

diegomura:
Above I added some changes needed to update the doc on the reconciler, as suggested by @phryneas.  I think it's the way to go, specially because this will also solve #236. Even though it's working quite well, it breaks page wrapping so still needs to be improved. This will probably involve changing the page wrapping algorithm itself, and maybe the element data structure, which can be a big disruptive change. Going to iterate on this idea these next days

diegomura:
So... this is (almost) ready to go!

All I have to do is write docs for the new API bits, and or course, to be tested by anyone how can.
I did A LOT of testing myself on different scenarios and layouts, in addition of writing a handful of new automated tests for all elements and interfaces.

This PR escalated more than wanted initially, but was mostly because live-reloading and rendering on the fly were tightly related. Now layout and page wrapping relies more in Yoga, which is awesome since it's super fast and reliable.

I'm merging this tomorrow probably when I have more time to write docs.

PR #288, opened 2018-08-18
ensure backward compatibility with height and width props

ybbond:
in this PR, I ensure users from previous version of `@react-pdf/renderer` (from version alpha 14 and below) to be able to keep their code that has `width` and `height` props to their `<Document />` component.

I put `width` and `height` before `style`, because users from previous package version (by my assumption) would not have `style` props defined. So that if migrating users defined `style` after they upgraded, the defined `style` will have its effect, the spread operator will replace the `width` and `height` from props.

###### **p.s.** this is my first contribution to github community, please let me know if my PR's quality standard is not enough or something went wrong. thank you :) ######

### DIFF
--- a/src/dom.js
+++ b/src/dom.js
@@ -61,7 +61,7 @@ export class Document extends Component {
   }
 
   render() {
-    const { className, width, height, style } = this.props;
+    const { className, width = null, height = null, style } = this.props;
 
     return (
       <iframe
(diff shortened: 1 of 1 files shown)


diegomura:
Thanks @ybbond! I’m not in front of the computer now, but I’ll review it asap

diegomura:
This is going to be erased after #285 is merged, but let's merge it temporarily. Thanks!

PR #291, opened 2018-08-23
Update README.md

timgivois:


diegomura:
@timgivois wait. I rollback the merge. I read it quickly and I thought you were proposing the opposite change.

`@react-pdf/node` is obsolete now. `@react-pdf/renderer` is now the only package you need for all environments.

timgivois:
Oh then we have to create a PR to the react-pdf.org documentation
image

PR #308, opened 2018-09-14
Support CSS transformations

### LINKED ISSUE #306: Support CSS transformations
**Description:**
Add CSS transformations support to react-pdf. Pdfkit already provides some transformation primitives we can use: http://pdfkit.org/docs/vector.html

Available transformations:
- Scale
- Rotate
- Translate

refs: https://css-tricks.com/almanac/properties/t/transform/

diegomura:
- [x] Add tests
- [x] Enable rotation in not just view
- [x] Enable translation
- [x] Enable scale
- [x] Add new style to docs list
- [x] Support negative values

Fixes #306


lukepighetti:
Hi there, do we have an example of this working? I am using styled components and haven't had any luck getting `transform: rotate(-45deg);` to work on a View component

PR #344, opened 2018-10-20
Fix base64 images in Node

### LINKED ISSUE #342: Image: Base64 as src is broken
**OS:**
MacOS 10.14

**React-pdf version:**
1.0.0-alpha.18

**Description:**
Base64 as `Image` `src` is broken and throws `Couldn't fetch image: <string>`. `isomorphic-fetch`, which used `node-fetch` under the hood throws `Error: only http(s) protocols are supported`.

I suspect this commit, that implemented `isomorphic-fetch`: https://github.com/diegomura/react-pdf/commit/72d0454c0876c24386b7a37b53f39293ef3540d0, as it's younger than #216, which seems to conclude that this should be working as expected and documented in the docs.

**Suggested actions**
- [x] Somehow detect and handle base64 strings as `src`, as `node-fetch` simply does not support this. Maybe something like `Buffer.from(<string>, 'base64')`.
- [x] Properly output `node-fetch` errors for future debugging, instead of just `Couldn't fetch image: <string>`.
- [x] Add a base64 test case in `image.test.js`.

Alternatively, remove the instructions from the documentation, if this functionality is unwanted.

https://runkit.com/embed/uj7thrhoup3u

diegomura:
Fixes #342


src/utils/image.js
```diff
@@ -2,7 +2,12 @@ import fetch from 'isomorphic-fetch';
 import PNG from '@react-pdf/png-js';
 import JPEG from './jpeg';
 
-const Buffer = require('buffer/').Buffer;
+const isValidFormat = format => {
+  const lower = format.toLowerCase();
+  return lower === 'jpg' || lower === 'jpeg' || lower === 'png';
+};
+
+const isBase64 = src => /data:image\/[a-zA-Z]*;base64,[^\"]*/g.test(src);
```
ostrgard: `"` doesn't need escaping, as your linter suggests.

src/utils/image.js
```diff
@@ -16,31 +21,41 @@ function getImage(body, extension) {
...
-      // Local file given
-      return new Promise((resolve, reject) =>
-        resolve(getImage(src.data, src.format)),
-      );
-    }
-    throw new Error(
-      `Invalid data given for local file: ${JSON.stringify(src)}`,
-    );
+const resolveBase64Image = src => {
+  const match = /data:image\/([a-zA-Z]*);base64,([^\"]*)/g.exec(src);
```
ostrgard: Same as above.

src/utils/image.js
ostrgard: I'm no expert on this, but reading up on matching for base64 it does seem a bit more complex than this suggests. I understand that separating data and filetype is the objective here, so it makes sense to keep it simple. I would suggest renaming the function to `isCompatibleBase64`, just to make this distinction clear.

src/utils/image.js
ostrgard: This regex is identical to the `isBase64` regex besides the catching groups, right? If so, they should be able to use the same, extracted to a single `const`, just for simplicity.

diegomura:
Thanks for the review @ostrgard ! Merging this and publishing very soon. Thanks for your patience

Issue #363, opened 2018-10-31
yoga-layout does not compile on Amazon Linux 1


DuncanMacWeb:
**OS:** Amazon Linux 1

**React-pdf version:** 1.0.0-alpha.18

**Description:** We have been unable to push react-pdf to our deployment pipeline, because the pipeline runs on Amazon Linux 1 which has restrictions that make it difficult/impossible to update GCC to a version required to compile Yoga ([4.9.3+, I believe](https://gitlab.com/sse/sophos/issues/1#note_16948539)). The outdated GCC produced the error `g++: error: unrecognized command line option '-std=c++14'` (more output below). We are unable to upgrade the base images we are using right now.

**How to replicate issue including code snippet (if applies):** installing react-pdf fails with the following error:

```
> yoga-layout@1.9.3 install /tmp/app/node_modules/yoga-layout
> npm run build:node


> yoga-layout@1.9.3 build:node /tmp/app/node_modules/yoga-layout
> npm run copy-sources && autogypi && node-gyp configure build


> yoga-layout@1.9.3 copy-sources /tmp/app/node_modules/yoga-layout
> ! npm -s run is-monolithic || (rsync -r --checksum --delete ../yoga/ sources/yoga/)

make: Entering directory `/tmp/app/node_modules/yoga-layout/build'
  COPY Release/obj.target/nbind/geni/symbols.txt
  CXX(target) Release/obj.target/nbind/sources/yoga/Utils.o
g++: error: unrecognized command line option '-std=c++14'
make: *** [Release/obj.target/nbind/sources/yoga/Utils.o] Error 1
make: Leaving directory `/tmp/app/node_modules/yoga-layout/build'
gyp ERR! build error 
gyp ERR! stack Error: `make` failed with exit code: 2
gyp ERR! stack     at ChildProcess.onExit (/usr/lib/node_modules/npm/node_modules/node-gyp/lib/build.js:262:23)
gyp ERR! stack     at emitTwo (events.js:126:13)
gyp ERR! stack     at ChildProcess.emit (events.js:214:7)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:198:12)
gyp ERR! System Linux 4.14.72-68.55.amzn1.x86_64
gyp ERR! command "/usr/bin/node" "/usr/lib/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js" "configure" "build"
gyp ERR! cwd /tmp/app/node_modules/yoga-layout
[...]

Issue #337, opened 2018-10-09
Error installing package


jayanthnd:
**OS: Mac OS High Sierra**

**React-pdf version: 1.0.0-alpha.18**

**Description: Unable to yarn/npm install @react-pdf/renderer**

**How to replicate issue including code snippet (if applies): Tried running `yarn install @react-pdf/renderer` and get the following error:
fatal error: too many errors emitted, stopping now [-ferror-limit=]
20 errors generated.
make: *** [Release/obj.target/nbind/sources/yoga/Utils.o] Error 1
gyp ERR! build error
gyp ERR! stack Error: `make` failed with exit code: 2
gyp ERR! stack     at ChildProcess.onExit (/usr/local/lib/node_modules/npm/node_modules/node-gyp/lib/build.js:262:23)
gyp ERR! stack     at ChildProcess.emit (events.js:182:13)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:240:12)
gyp ERR! System Darwin 17.6.0
gyp ERR! command "/usr/local/Cellar/node/10.11.0/bin/node" "/usr/local/lib/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js" "configure" "build"
gyp ERR! cwd /Users/node_modules/yoga-layout
gyp ERR! node -v v10.11.0
gyp ERR! node-gyp -v v3.8.0
gyp ERR! not ok
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! yoga-layout@1.9.3 build:node: `npm run copy-sources && autogypi && node-gyp configure build`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the yoga-layout@1.9.3 build:node script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
**

_You can make use of [react-pdf REPL](https://react-pdf/repl) to share the snippet_

I've seen a couple of other threads with similar issues, just wondering if there's a workaround for it.

diegomura:
That’s an issue about Yoga installation on your system. I would reccomend you to look for more information in the yoga-layout repo, or the ones that were created here (such as #327)

I’m not available right now and I’ll try to take a closer look when I got back from vacations

jayanthnd:
Thanks for the reply @diegomura. I fixed the issue, turned out it had to do with how the header files for the yoga library were getting referenced in my local setup.

[...]

Issue #251, opened 2018-07-01
Installing dependencies end up with 'yoga-layout' error


mitkpl:
**OS:** Windows

**React-pdf version:** 0.7.6

**Description:** Installing first dependencies occuring with error

**How to replicate issue including code snippet (if applies):**
1. Create react app with Create-react-app
2. Go to the folder
3. Open some CMD and `yarn add @react-pdf/core`
4. ERROR
`...\node_modules\yoga-layout: Command failed.`. And info about details in error log:
```
12 info lifecycle yoga-layout@1.9.3~build:node: Failed to exec build:node script
13 verbose stack Error: yoga-layout@1.9.3 build:node: `npm run copy-sources && autogypi && node-gyp configure build`
13 verbose stack Exit status 1
```

another-guy:
I confirm that `yarn add @react-pdf/core` fails for me as well with a similar error:

```sh

igor@DESKTOP-MRND0KT MINGW64 /c/GitHub/pdf-demo (master)
$ yarn add @react-pdf/core
yarn add v1.6.0
(node:16400) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
[1/4] Resolving packages...
[2/4] Fetching packages...
info fsevents@1.2.4: The platform "win32" is incompatible with this module.
info "fsevents@1.2.4" is an optional dependency and failed compatibility check. Excluding it from installation.
[3/4] Linking dependencies...
[4/4] Building fresh packages...
[1/1] ⠠ yoga-layout: npm ERR!     C:\Users\igor\AppData\Roaming\npm-cache\_logs\2018-07-02T07_47_24_790Z-debug.log
[-/1] ⠠ waiting...
[-/1] ⠠ waiting...
[-/1] ⠠ waiting...
error C:\GitHub\pdf-demo\node_modules\yoga-layout: Command failed.
Exit code: 1
Command: npm run build:node
Arguments:
Directory: C:\GitHub\pdf-demo\node_modules\yoga-layout
Output:
> yoga-layout@1.9.3 build:node C:\GitHub\pdf-demo\node_modules\yoga-layout
> npm run copy-sources && autogypi && node-gyp configure build


> yoga-layout@1.9.3 copy-sources C:\GitHub\pdf-demo\node_modules\yoga-layout
[...]

Issue #225, opened 2018-05-02
Using react-pdf in a lambda fails to install


dberringer:
**OS:**
Linux 4.9.85-38.58.amzn1.x86_64
**React-pdf version:**
0.7.6
**Description:**
Does anyone know how to run react-pdf in a lambda?  I'm trying to install it(yarn add @react-pdf/core) but I get the following error:

yarn add v1.6.0
[1/5] Validating package.json...
[2/5] Resolving packages...
[3/5] Fetching packages...
[4/5] Linking dependencies...
[5/5] Building fresh packages...
[1/1] ⠈ yoga-layout: NVM_NODEJS_ORG_MIRROR is deprecated and will be removed in node-gyp v4, please use NODEJS_ORG_MIRROR
[-/1] ⠈ waiting...
[-/1] ⠈ waiting...
[-/1] ⠈ waiting...
error /home/ec2-user/environment/Commerce/node_modules/yoga-layout: Command failed.
Exit code: 1
Command: npm run build:node
Arguments: 
Directory: /home/ec2-user/environment/Commerce/node_modules/yoga-layout
Output:
> yoga-layout@1.9.3 build:node /home/ec2-user/environment/Commerce/node_modules/yoga-layout
> npm run copy-sources && autogypi && node-gyp configure build

> yoga-layout@1.9.3 copy-sources /home/ec2-user/environment/Commerce/node_modules/yoga-layout
> ! npm -s run is-monolithic || (rsync -r --checksum --delete ../yoga/ sources/yoga/)

gyp WARN download NVM_NODEJS_ORG_MIRROR is deprecated and will be removed in node-gyp v4, please use NODEJS_ORG_MIRROR
gyp WARN download NVM_NODEJS_ORG_MIRROR is deprecated and will be removed in node-gyp v4, please use NODEJS_ORG_MIRROR
gyp WARN download NVM_NODEJS_ORG_MIRROR is deprecated and will be removed in node-gyp v4, please use NODEJS_ORG_MIRROR
make: Entering directory `/home/ec2-user/environment/Commerce/node_modules/yoga-layout/build'
  COPY Release/obj.target/nbind/geni/symbols.txt
  CXX(target) Release/obj.target/nbind/sources/yoga/Utils.o
g++: error: unrecognized command line option ‘-std=c++14’
make: *** [Release/obj.target/nbind/sources/yoga/Utils.o] Error 1
[...]

PR #364, opened 2018-10-31
Depend on yoga-layout-prebuilt instead of yoga-layout

DuncanMacWeb:
Amazon Linux 1 ships with an outdated version of GCC. In order to deploy react-pdf successfully on Amazon Linux 1, we depend on [yoga-layout-prebuilt](https://github.com/vadimdemedes/yoga-layout-prebuilt), which works really well, instead of yoga-layout, which does not compile.

(Aside: to make yoga-layout(-prebuilt) run at an acceptable speed, we minified production code using [terser](https://github.com/terser-js/terser) instead of UglifyJS; it produced an almost magical improvement in execution speed, probably due to not stripping the `"use asm";` magic string from the asm.js output.)

References: #57, #142, #169
Resolves: #363 #337 #225 (would also address #211, #251, etc.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "page-wrapping": "^1.0.2",
     "react": "^16.4.1",
     "react-reconciler": "^0.12.0",
-    "yoga-layout": "1.9.3"
+    "yoga-layout-prebuilt": "1.9.3"
   },
   "devDependencies": {
     "babel-cli": "^6.22.2",
--- a/src/elements/Image.js
+++ b/src/elements/Image.js
@@ -1,4 +1,4 @@
-import Yoga from 'yoga-layout';
+import Yoga from 'yoga-layout-prebuilt';
 import warning from 'fbjs/lib/warning';
 import Base from './Base';
 import { resolveImage } from '../utils/image';
--- a/src/elements/Node.js
+++ b/src/elements/Node.js
@@ -1,4 +1,4 @@
-import Yoga from 'yoga-layout';
+import Yoga from 'yoga-layout-prebuilt';
 import upperFirst from 'lodash.upperfirst';
 
 const PERCENT = /^(\d+)?%$/g;
--- a/src/elements/Text.js
+++ b/src/elements/Text.js
@@ -1,4 +1,4 @@
-import Yoga from 'yoga-layout';
+import Yoga from 'yoga-layout-prebuilt';
 import createPDFRenderer from '@textkit/pdf-renderer';
 import Base from './Base';
 import Font from '../font';
--- a/src/stylesheet/yogaValue.js
+++ b/src/stylesheet/yogaValue.js
@@ -1,4 +1,4 @@
-import Yoga from 'yoga-layout';
+import Yoga from 'yoga-layout-prebuilt';
 
 const yogaValue = (prop, value) => {
   const isAlignType = prop =>
--- a/tests/stylesheet.test.js
+++ b/tests/stylesheet.test.js
@@ -1,4 +1,4 @@
-import Yoga from 'yoga-layout';
+import Yoga from 'yoga-layout-prebuilt';
 import { StyleSheet } from '../src';
 
 describe('StyleSheet', () => {
--- a/yarn.lock
+++ b/yarn.lock
@@ -19,14 +19,12 @@
     js-tokens "^4.0.0"
 
 "@react-pdf/fontkit@^1.11.0":
-  version "1.11.0"
-  resolved "https://registry.yarnpkg.com/@react-pdf/fontkit/-/fontkit-1.11.0.tgz#2492ffca9454be2a52eb2aba01eacdf2a11ce9fd"
-  integrity sha1-JJL/ypRUvipS6yq6AerN8qEc6f0=
+  version "1.13.0"
+  resolved "https://registry.yarnpkg.com/@react-pdf/fontkit/-/fontkit-1.13.0.tgz#ec14cc61120e814c1d48cf276771684ec615be9e"
+  integrity sha512-g4rxtkSkEbIwDoqZc6ZM2yHq/imqmgGXQuMDnrkM2yI9TpTGhQfJGnz9IrCadrRgOrEx9orxRuTALOxzAce7Kg==
   dependencies:
     "@react-pdf/unicode-properties" "^2.2.0"
-    brfs "^1.4.0"
     brotli "^1.2.0"
-    browserify-optional "^1.0.0"
     clone "^1.0.1"
     deep-equal "^1.0.0"
     dfa "^1.0.0"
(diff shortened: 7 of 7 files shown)


diegomura:
I tried to prebuild `yoga-layout` before publishing in the past, and the issue I had is that it gets build for the current node version I was using. If someone else tried to use the library with another node version, yoga complained. Do you think this is going to be an issue with a solution such as this?

diegomura:
@DuncanMacWeb btw, where is the terser parsing happening? Couldn't find it, but I'm very intrigued it that really makes such a big performance improvenent

diegomura:
Ok. I tried to make the build in react-pdf project but the result was a bit messy. I'm ok to use `yoga-layout-prebuilt` straight away. But I would like to know those two points I mentioned: 

- Is this prebuilt going to work on all Node versions?
- Where's the terser parsing happening?

Thanks in advance for this effort!

DuncanMacWeb:
Hey @diegomura, no problem! Thanks for creating this amazing library :)

- As far as I know, yoga-layout-prebuilt is compiled down to simple asm.js, so it will work in any JavaScript environment that supports asm.js (any that supports Typed Arrays, basically). I think [all versions of Node](https://node.green/#ES2015-built-ins-typed-arrays) support the [Typed Array features that asm.js depends on](http://asmjs.org/spec/latest/#heap-view-types). How were you doing your previous pre-compilation?
- The terser parsing is not happening in this PR :) I just mentioned as an aside that we are using Terser in our own Webpack build, because other people might also have problems with UglifyJS stripping out the `"use asm";` magic string. It doesn’t really have anything to do with this PR, sorry!

If you want to build yoga-layout automatically and include it in react-pdf, I think the best way I can think of would be the `prepublishOnly` hook, the same as yoga-layout-prebuilt uses. But they deliberately don’t have yoga-layout as am explicit dependency, so that npm doesn’t try to compile it.

diegomura:
Thanks for your clarification!

I don't exactly remember what I was doing when I tried include precompiled yoga before, but I've been trying this solution and seems to work fine in each env I tested.

Sorry I'm a bit lost in the `asm` world, but does that means that everyone using `UglyfyJS` to build their projects (which are a bunch of people) will encounter problems building react-pdf if we use this precompiled version? I don't want to forced users to use `terser` if they want to use react-pdf. Sorry if the question is silly!

About building yoga-layout in the project, I came to a solution using `prepublish`, and adding `yoga-layout` as a devDependency, including the yoga build into `package.json files` so it get published along with the lib. However, I experienced issues when running tests, and the overall result wasn't that nice as just importing `yoga-layout-prebuilt`. I don't think the project is ever going to be published on yoga's site, so I think it's just better to follow this approach anyways 😂

vadimdemedes:
> But @vadimdemedes deliberately doesn’t have yoga-layout as am explicit dependency, so that npm doesn’t try to compile it on npm install.

That's correct. If package depends on `yoga-layout`, it will try to build it on install. I've ran into this problem when trying to set up Travis CI builds and it lacked some of the required native libs to build it. That's why I built `yoga-layout-prebuilt` to avoid this problem.

diegomura:
Thanks @vadimdemedes !
We are having similar issues with react-pdf users. Not on Travis, but when yoga gets built on certain envs that does not have those native libs. Is there any downside you can think of using `yoga-layout-prebuilt` here we might be missing?

vadimdemedes:
Can't think of any, really. How `yoga-layout-prebuilt` works is it just copies files from the official build, nothing more. I assume this the same bundle used in the official Yoga Layout website (https://yogalayout.com/playground), so I think it's safe.

DuncanMacWeb:
To clarify: asm.js actually requires an optimised compilation pipeline to run fast; this is why `"use asm";` is important, as it activates the pipeline. Asm.js experienced a [>100% performance boost in Chrome 28](https://blog.chromium.org/2013/05/chrome-28-beta-more-immersive-web.html) (V8 revisions 14306, 14448, 14534) and the latest Chrome 28 release contained V8 revision

DuncanMacWeb:
To clarify: asm.js actually requires an optimised compilation pipeline to run fast; this is why `"use asm";` is important, as it activates the pipeline. Asm.js experienced a [>100% performance boost in Chrome 28](https://blog.chromium.org/2013/05/chrome-28-beta-more-immersive-web.html) (V8 revisions 14306, 14448, 14534) and the latest Chrome 28 release [contained V8 revision 14510](https://src.chromium.org/viewvc/chrome/releases/28.0.1501.1/DEPS?revision=198854) but I’m not sure which Node release corresponded to those V8 revisions. Anyway, it was 5 years ago :)

I’m not sure about the UglifyJS issue. `webpack@4.25.1` [depends on `uglifyjs-webpack-plugin@1.2.4`](https://github.com/webpack/webpack/blob/v4.25.1/package.json), which depends on [`uglify-es@3.3.4`](https://github.com/webpack-contrib/uglifyjs-webpack-plugin/blob/v1.2.4/package.json). It looks like this was addressed in [two](https://github.com/mishoo/UglifyJS2/pull/818) [threads](https://github.com/mishoo/UglifyJS2/issues/2456) over there. I’m not sure why we experienced this issue; I just switched to terser and noticed the parse time went down dramatically, but it could have been a non-standard part of our build process or some non-standard options we set, too. I guess, if no-one else has an issue with UglifyJS, it’s all good? If you have tried this on multiple environments (ours was Webpack 4, Node 8, Chrome 70) and it works... maybe it’s OK :shrug:

DuncanMacWeb:
Sorry @diegomura, I have _no idea_ why GitHub closed this. I must have said the magic incantation by mistake.

diegomura:
Thank you very much @DuncanMacWeb !! I'm pretty excited about not building yoga every time and the benefits this will have to users!
I just published a `1.0.0-alpha.25` with these changes

DuncanMacWeb:
Amazing, thank you @diegomura. I will switch to `1.0.0-alpha.25` with UglifyJS and report any issues 👍

DuncanMacWeb:
Preliminary results with `1.0.0-alpha.25`: total download+parse+render time is hovering around 8–9s with UglifyJS and around 6–7s with terser for a sample document in Chrome 70, so the difference is noticeable but not as dramatic as previously. Not sure why. I’d definitely recommend people to use terser to minify react-pdf/yoga-layout!

By the way... I think it would be good if we could get yoga-layout to load as WebAssembly in environments which support it (Node 8+ and [80% of browsers](https://caniuse.com/#feat=wasm)), as I think it would be significantly smaller and probably faster...

diegomura:
There is [yoga-dom])(https://github.com/vincentriemer/yoga-dom) which already builds yoga in wa. However, there is a slight but important difference on that lib: it exports a promise that resolves on the Yoga objects. I tried to setup this on the lib but I wasn't able to do it without changing the API.

However, does it make that _much_ difference? Something I've been wanting to do for awhile is a performance test, and identify where are our greatest bottle necks. From what I saw those are page wrapping (that handles yoga layout calculation) and text layout (in which I'm working right now to make it more fast) and assets loading (there's no much we can do on that). However, I'm not sure on what percentages

PR #365, opened 2018-11-04
Fix loading available emojis

agarant:
The string on the TextInstance node is on the value property. As a result, emojis where not loaded before the rendering.


diegomura:
Thanks @agarant !!

PR #399, opened 2018-11-30
important things

shinriyo:
It is for server-side only.


diegomura:
Thanks!
I always thought that was clear. After all, you cannot create a file directly to the file-system from the browser.
I appreciate your PR, but I think it's more clear in each title (as it's now).

Issue #394, opened 2018-11-29
Documentation: starting from create-react-app


lukepighetti:
Hey all,

I'm trying to get up and running with `react-pdf` starting with `create-react-app` and I have reached a limit in my understanding of how packages like this work.

This is my index.js file

```
import React from "react";
import ReactDOM from "react-dom";
import ReactPDF, { PDFViewer } from "@react-pdf/renderer";

import { MyDocument } from "./sections/Test";

// ReactPDF.render(<MyDocument />, `${__dirname}/test.pdf`); // I have to uncomment this line

const App = () => (
  <PDFViewer>
    <MyDocument />
  </PDFViewer>
);

ReactDOM.render(<App />, document.getElementById("root"));
```

Now, I am trying to render PDF previews in the browser so I can render-on-save (for productivity) and then finally build a PDF as part of my `npm run build` script. This is my understanding of how `react-pdf` can be put to work.

The above shows the PDF in the PDF Viewer, but only after removing the build line due to error

```
TypeError: undefined is not a function (near '..._react_pdf_renderer__WEBPACK_IMPORTED_MODULE_2__["default"].renderToFile...')
```

So, I am sure that my setup is wrong, but I haven't been able to find documentation as to how to get this render function to work in Node. When I put the function in its own file and try to run it with Node I predictably get an error due to not first using Babel or whatever to compile everything over.

So I am requesting an addendum to the docs that assume the user has no knowledge of React outside of Create React App to get started building PDFs with `react-pdf`

Cheers!

lukepighetti:
tried to install babel-node and I'm getting nowhere.

```
{
  "presets": [
    [
      "@babel/preset-react",
      {
        "pragma": "dom", 
        "pragmaFrag": "DomFrag",
        "throwIfNamespace": false 
      }
    ]
  ]
}
```


```
  "scripts": {
    "start": "react-scripts start",
    "build": "react-scripts build",
    "test": "react-scripts test",
    "eject": "react-scripts eject",
    "pdf": "babel-node ./src/render.js"
  },
  "devDependencies": {
    "@babel/cli": "^7.1.5",
    "@babel/core": "^7.1.6",
    "@babel/node": "^7.0.0",
    "@babel/preset-react": "^7.0.0"
  }

```

lukepighetti:
Hi, sorry about the confusion, I probably could have done a better job of explaining myself.

[...]

PR #400, opened 2018-11-30
add babel instructions

lukepighetti:
This pull request adds basic instructions for setting up Babel for `react-pdf`

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,25 @@ Then, run `yarn example -- <example-name>`
 yarn example -- fractals
 ```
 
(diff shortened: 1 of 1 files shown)


lukepighetti:
addresses https://github.com/diegomura/react-pdf/issues/394

diegomura:
Hey sorry for not addressing this. 
To be honest I don't think this is something we should include in the README, for the same reason any other React library doesn't include how you should setup Babel for React. 
I understand it's necessary to setup Babel, but it is for using React itself, not react-pdf.

PR #408, opened 2018-12-01
Add insertBefore method to renderer

### LINKED ISSUE #396: Iteration of Dynamic Data Arrays
I am trying to iterate of an array of data to generate my PDF.

I have been able to make everything work, except for when the Array is dynamic.
When the array is hard coded, it will work just fine. but things break when the array comes from props that change after the component has mounted.
It will break with the error : 
```
TypeError
insertBefore is not a function
```

This sandbox kinda shows what I am trying to do.
https://codesandbox.io/s/my3mlmwjv8

My app is using props from redux instead of state.

Thank you for any help.

diegomura:
Fixes #396

### DIFF
--- a/src/elements/Base.js
+++ b/src/elements/Base.js
@@ -54,6 +54,31 @@ class Base extends Node {
     this.props.break = value;
   }
 
+  appendChild(child) {
+    super.appendChild(child);
+    this.root.markDirty();
+  }
+
+  appendChildBefore(child, beforeChild) {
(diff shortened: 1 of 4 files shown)


AirborneEagle:
Thank you for doing that so quickly!!  Your awesome.
How does the whole release thing work for react-pdf? ie: How often do features like this get release?

thanks again!!

diegomura:
Very soon! The next version will also include some deep changes in the text layout engine and I have to be careful that it does not break something.

PR #412, opened 2018-12-06
Canvas node

### LINKED ISSUE #410: Expose API to render graphics
**Description:**
It would be useful to be able to render vector graphics using react-pdf. 

One way could be to expose a moveTo-lineTo-type API (like PDFKit):

```
  <Document>
    <Page size="A4" style={styles.page}>
      <Canvas>
        {painter => painter
            .moveTo(100, 150)
            .lineTo(100, 250)
            .lineTo(200, 250)
            .fill("#FF3300")}
      </Canvas>
    </Page>
  </Document>
```

Another, to provide graphical primitives:
```
  <Document>
    <Page size="A4" style={styles.page}>
      <Path points={[[100, 150], [100, 250], [200, 250]]} fill="#FF3300"/>
    </Page>
  </Document>
```

markusjohnsson:
This is work in progress to resolve #410. This PR introduces a `Canvas` element to enable drawing of vector graphics into a PDF document. It does this by exposing a drawing API based on a limited pdfkit instance. 

Example: Draw a triangle
```jsx
  <Document>
    <Page size="A4" style={styles.page}>
      <Canvas paint={painter => painter
            .moveTo(100, 150)
            .lineTo(100, 250)
            .lineTo(200, 250)
            .fill("#FF3300")}/>
    </Page>
  </Document>
```

This is work in progress. Remaining tasks:

- [x] Finalize API
- [X] Surprising warning: "Functions are not valid as a React child." - consider attribute prop instead
- [x] Adjust left, top, width and height to content area (respect borders, padding etc)
- [x] DOM rendering (?)
- [x] Show warning if canvas does not have width or height
- [x] Canvas should not wrap
- [x] Clip canvas area
- [x] Tests


markusjohnsson:
> Be careful with this file. `react-pdf` uses Yarn

Yes, this was created (and the committed 🙄) by mistake when I used `npm link` to test functionality. Will remove.

diegomura:
`Surprising warning: "Functions are not valid as a React child." - consider attribute prop instead`

Oh 😓 Forgot about that. I remember having that issue when implementing dynamic rendering. React renderer does not support function children. The solution here would be changing the `Canvas` API a bit:

```jsx
<Canvas render={painter => {
  painter
    .moveTo(100, 150)
    .lineTo(100, 250)
    .lineTo(200, 250
    .fill("#FF3300")
}} />

**Comment by markusjohnsson** [comment]
> The solution here would be changing the `Canvas` API a bit

Yeah. It turns out `render` conflicted with dynamic nodes so I opted for `paint` instead:

```jsx
  <Document>
    <Page size="A4" style={styles.page}>
      <Canvas paint={painter => painter
            .moveTo(100, 150)
            .lineTo(100, 250)
            .lineTo(200, 250)
            .fill("#FF3300")}/>
    </Page>
  </Document>
```

diegomura:
I figured out that since we are exposing `font` and `text` primitives, why not just passing by the pdf instance and allow rendering all sorts of things. Of course, the user will have to use it own it's own risk, since it will have total access to the document, but this way we won't have to deal with future requests petitions

PR #426, opened 2018-12-20
Add TS typings

yenbekbay:
First of all, thanks for the awesome work on the project!

Here's my attempt to type most of the public interfaces. I could submit the typings to [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/), but it'd be nicer to have the typings bundled with the library to avoid adding an extra dependency.


yenbekbay:
@diegomura Hi! I updated the typings for `Document`.

diegomura:
Thanks! It will be published in next version

Issue #431, opened 2018-12-27
Custom fonts cause the page to break and leak resources


puregarlic:
**OS:** macOS Mojave 10.14.3 Beta

**React-pdf version:** v1.1.0

**Description:**

Applying a custom font in the browser seems to cause some sort of resources leak, ultimately preventing the document from rendering.

**How to replicate issue including code snippet (if applies):**

I put together a reproduction repo: https://github.com/puregarlic/react-pdf-repro

This is all the code needed to break it:

```jsx
import React from 'react'
import { PDFViewer, Document, Font, Page, Text } from '@react-pdf/renderer'

// font included in the example repository
import PTSansCaption from './fonts/PT_Sans-Caption-Web-Regular.ttf'

Font.register(PTSansCaption, { family: 'PT Sans Caption' })

const App = () => (
  <PDFViewer width="100%" height="100%">
    <Document>
      <Page>
        {/* Removing the style prop fixes the page */}
        <Text style={{ fontFamily: 'PT Sans Caption' }}>Hello, world</Text>
      </Page>
    </Document>
  </PDFViewer>
)

export default App
```

maxsbelt:
Hi @puregarlic. I opened PR that fixes infinite loop problem but it's not the root of your problem. To fix the actual problem you need to update `Font.register` to use absolute url:

```js
// example for create react app in dev mode
Font.register(`http://localhost:3000${PTSansCaption}`, { family: 'PT Sans Caption' })
```

Jacob33123:
I'm having this issue as well. Up until about a month or so, I was registering the fonts like so:

`export const Roboto = Font.register(`${__dirname}/Roboto-Regular`, {
  family: 'Roboto',
})

export const RobotoBold = Font.register(`${__dirname}/Roboto-Bold`, {
  family: 'Roboto Bold',
})

export const RobotoMono = Font.register(`${__dirname}/RobotoMono-Regular`, {
  family: 'Roboto Mono',
})

export const RobotoMonoBold = Font.register(`${__dirname}/RobotoMono-Bold`, {
  family: 'Roboto Mono Bold',
})`

and importing in the index files for the PDFs i'm generating like so: 
`import ../fonts`
 
[...]

PR #434, opened 2019-01-02
Fixed page hangs on error

maxsbelt:
## Description

Fixed https://github.com/diegomura/react-pdf/issues/431:
- Fixed problem with infinite rerendering on error
- Added better error message for invalid font url in browser environment


diegomura:
Thanks @maxsbelt !
Unfortunately I’m out of town for some days without access to my computer, so I cannot publish a new version with this change. But I’ll merge it in the meantime and publish as soon as I can 😊

PR #438, opened 2019-01-09
Open local fonts asynchronously

DuncanMacWeb:
This loads local fonts asynchronously to improve server-side performance by no longer blocking the main thread while font files are loaded.

~This depends on:~

- ~react-pdf/fontkit#4 being merged~
- ~this PR being updated with the new version of `react-pdf/fontkit`~

### DIFF
--- a/src/font/index.js
+++ b/src/font/index.js
@@ -36,7 +36,7 @@ const getHyphenationCallback = () => hyphenationCallback;
 const load = async function(fontFamily, doc) {
   const font = fonts[fontFamily];
 
-  // We cache the font to avoid fetching it many time
+  // We cache the font to avoid fetching it many times
   if (font && !font.data && !font.loading) {
     font.loading = true;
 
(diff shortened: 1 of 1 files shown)


diegomura:
Thanks @DuncanMacWeb !
Wouldn't be also possible to fix this only from the `react-pdf` side, by creating a promise wrapper around `openSync`?
I truly don't mind merging changes in both repos, but I usually try to leave the forks as much unchanged as possible. In this case the change is not big at all, but just wanted to throw this in the table before merging 😄

DuncanMacWeb:
Good call @diegomura — that simplifies this significantly.

Sorry, a change I made locally to `package.json` seems to have slipped in to the commit and I wasn’t able to remove it. But you might be happy to have it in there anyway, it’s quite useful ;)

Also, I’m not sure why a couple of image tests are failing, but they were also failing for me on the `master` branch, so I think that’s unrelated.

diegomura:
Thanks for updating this PR @DuncanMacWeb !
I'm testing it right now and I'll merge it soon

PR #449, opened 2019-01-18
fix export of PDFViewer, BlobProvider and PDFDownloadLink

william726:


src/index.js
```diff
@@ -12,6 +12,9 @@ const Page = 'PAGE';
 const Note = 'NOTE';
 const Image = 'IMAGE';
 const Document = 'DOCUMENT';
+const PDFViewer = 'PDFViewer';
+const BlobProvider = 'BlobProvider';
+const PDFDownloadLink = 'PDFDownloadLink';
```
diegomura: Sorry. But I don't get this.
These variables are supposed to be React Components, not strings.
I understand this may solve the issues you are having with typings, but it doesn't make much sense for the library itself. Specially considering that the entry points for both Node and DOM bindings are `node.js` and `dom.js` files respectively. Those two files import `index.js`
Again, I know nothing about TypeScript, so I cannot say exactly

william726: Agree. And I tested my code will work if I only add the changes in `index.d.ts` (the export statements of the components). Removed the changes in `index.js`.

diegomura: Awesome! Thanks

PR #457, opened 2019-01-21
Allow local images to be specified by path

DuncanMacWeb:
This allows local images to be specified by file path. It recognises paths specified using the `file:` protocol as well as relative paths with no protocol. There is a safeguard in place (which can be overridden) that prevents files being accessed from outside a “safe” path, which defaults to `./public`.

As well as enabling the more user-friendly `src={string}` interface for local files, this also means that local files will be added to react-pdf images cache just like remote ones.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -91,6 +91,8 @@ declare module '@react-pdf/renderer' {
       debug?: boolean;
       src: string | {data: Buffer; format: 'png' | 'jpg'};
       cache?: boolean;
+      safePath?: string;
+      allowDangerousPaths?: boolean;
     }
 
     /**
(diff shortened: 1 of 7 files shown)


DuncanMacWeb:
Absolutely—particularly important to test for vulnerabilities with this one.

diegomura:
Hey @DuncanMacWeb . Sorry for taking so long.
I just tried this PR, and everything seems to be working fine. I'm not sure if I tested all scenarios on Node, but as far as I saw everything is ok. 
Thank you very much for taking the time to make this contribution. Going to merge this now, but do you mind also submitting another PR to the [site project](https://github.com/react-pdf/site) adding the new props to the docs?

PR #467, opened 2019-01-26
add type string to PDFViewerProps width and height

william726:
the official docs mention width and height of PDFViewer support both number and string


diegomura:
Thanks!

PR #469, opened 2019-01-28
fix(Text): dont add additional x value for lines

### LINKED ISSUE #361: Children of fixed View offsets unexpectedly
Hello!
At first I wanted to say thank you for so cool library! It's exactly what I looked for!

But unfortunately, I encountered with a disappointing problem.

**OS:**
Mac OSX and Ubuntu
**React-pdf version:**
@react-pdf/renderer@1.0.0-alpha.24

**Description:**
I found that children of View with _fixed_ flag moving to the right for each page > 1.
It's possible to see if you render whole document at a time and it consist of > 1 page.
At every page starts from second you can notice some offset for children of fixed View.

The root of the problem (as I understand it) is View-wrapper with fixed property
`
<View fixed>
  <Text>...</Text>
</View>
`
because 
`
<Text fixed>...</Text>
`
works ok.

I've created example repo. There are two already rendered pdf which shows the issue. But I've already setup everything for experiments and you could try it by yourself. Just follow readme, please.
[https://github.com/rederteph/react-pdf-issue](https://github.com/rederteph/react-pdf-issue)
Hope it will help to reproduce and understand issue.

If you have any further questions, please don't hesitate to contact me!

arahansen:
When Text nodes are cloned, the container is copied over, including x offsets for each line. For `fixed` Text elements, this means that the x value was being accumulated on top of the original rect offset which resulted in this shifting effect that gradually grows worse for each page the `fixed` element shows up on.

I don't have an exact repro jsx but it would look roughly like this:

```jsx
<Document>
  <Page>
    {/* Foo shows up fine on the first page, but is shifted to the right 
         8 pixels on subsequent pages 
    */}
    <Text style={{ padding: 8 }} fixed>Foo</Text>
     <LotsOfOtherContent />
  </Page>
</Document>
```

Before fix:
image

After fix:
image

Fixes #361

### DIFF
--- a/tests/text.test.js
+++ b/tests/text.test.js
@@ -4,6 +4,12 @@ import Text from '../src/elements/Text';
 import TextInstance from '../src/elements/TextInstance';
 import root from './utils/dummyRoot';
 
+jest.mock('@textkit/pdf-renderer', () => () =>
+  class PDFRendererMock {
+    render() {}
+  },
+);
+
 let dummyRoot;
 
 describe('Text', () => {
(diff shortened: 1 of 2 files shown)


tests/text.test.js
```diff
@@ -28,4 +34,27 @@ describe('Text', () => {
     expect(text.layoutEngine.layout.mock.calls).toHaveLength(1);
     expect(text.layoutEngine.layout.mock.calls[0][0].string).toBe('');
   });
+
+  test('Should render the same rect x value for clones', async () => {
```
arahansen: I dont love how this test is constructed. Let me know if you have a better means to test this!

diegomura: I think is fine! Maybe there's a simpler way, but this is not bad either. Thank you very much!

arahansen:
would it be possible to get a release cut that includes this change? Would love to integrate this back into my project 🙂

diegomura:
Definitely! I will release a new fix today. Sorry about the delay

arahansen:
No worries, I really appreciate your responsiveness and help getting started with the project!

PR #472, opened 2019-02-02
Run render prop twice in non-fixed elements

### LINKED ISSUE #470: page-wrapping does not provide totalPages
**Describe the bug**
When using `render` prop on `Text`, the `totalPages` value is not provided until the second pass.

This would usually be fine (and is appropriately called out in the docs to watch out for), however, if the `fixed` prop is not defined on `Text` then `render` is only called once and `1/undefined` gets rendered to the page.

Looks like the second pass render is [deliberately skipped](https://github.com/diegomura/react-pdf/blob/master/src/elements/Page.js#L147) if `fixed` is not supplied. Assuming as a performance optimization. But it seems maybe this is not a valid optimization that can be made.

**To Reproduce**
```
<Document>
  <Page>
    <Text render={({ totalPages }) => /* totalPages is undefined */} />
</Document>
```

**Expected behavior**
Either totalPages should be defined on the first pass, or `render` should always be called twice.

**Screenshots**

 - OS: MacOS
 - Browser: N/A (node)
 - React-pdf version: 1.1.1

diegomura:
Fixes #470

### DIFF
--- a/src/elements/Page.js
+++ b/src/elements/Page.js
@@ -144,7 +144,6 @@ class Page extends Base {
         node.removeAllChilds();
         const elements = node.props.render(props);
         await this.addDynamicChild(node, elements);
-        if (!node.fixed) node.props.render = null;
         continue;
       }
 


arahansen:
Looks like tests failed due to excessive memory?

```
Failed: "abort(\"Cannot enlarge memory arrays. Either (1) compile with  -s TOTAL_MEMORY=X  with X higher than the current value 134217728, (2) compile with  -s ALLOW_MEMORY_GROWTH=1  which allows increasing the size at runtime but prevents some optimizations, (3) set Module.TOTAL_MEMORY to a higher value before the program runs, or (4) if you want malloc to return NULL (0) instead of this abort, compile with  -s ABORTING_MALLOC=0 \")
```

This is something I’ve hit in different scenarios, and it’s pretty concerning from a performance point of view? Any thoughts on how to make this less of an issue in general for react-pdf?

diegomura:
In this case that error get's triggered because (_I think_) i's wrapping pages on and on, causing infinite yoga nodes to be added.

To be honest I don't have a fix yet to that. By what I saw this is caused by the pre-compiled Yoga project when there are many nodes in memory (which can grow very fast when wrapping pages). The error itself suggests some solutions which I couldn't test yet, but maybe we can fork the `precompile-yoga-layout` project and try compiling with a higher `TOTAL_MEMORY` or `ALLOW_MEMORY_GROWTH` flag

arahansen:
Yeah that would make sense. On the other hand, I saw the imposed limit as a sensible default, and going above that could strain some systems more than what’s reasonable.

And as you described, sort of acts as a safety guard against infinite nodes being loaded in memory

Issue #314, opened 2018-09-23
"Cannot enlarge memory arrays" error when document is big.


binchik:
**OS:**
MacOS 10.13.6

**React-pdf version:**
1.0.0-alpha.17

**Description:**
Hello! I'm using react-pdf to generate payment reports for a website I'm working on. The reports can often get really big, 400+ pages long. But I can only generate about 450 pages on my machine. If I try to render more, it hangs and finally crashes with the following error:
```
Cannot enlarge memory arrays. Either (1) compile with -s TOTAL_MEMORY=X with X higher than the current value 134217728, (2) compile with -s ALLOW_MEMORY_GROWTH=1 which allows increasing the size at runtime but prevents some optimizations, (3) set Module.TOTAL_MEMORY to a higher value before the program runs, or (4) if you want malloc to return NULL (0) instead of this abort, compile with -s ABORTING_MALLOC=0
```

I'm rendering the pdf reports in browser with PDFViewer.

**How to replicate issue including code snippet (if applies):**

I made a reproducible demo here (almost same code as in pageWrap example but rendering the same Don Quijote page content 1600 times):
https://github.com/binchik/react-pdf-crash-example

It loads for about 5-10 minutes then crashes with the "Cannot enlarge memory arrays" error.

Neil-G:
@diegomura hey, just wanted to let you know that I've reproduced this as well, would appreciate an update if you're able to help- thanks!

diegomura:
Thanks for reporting it. Never experienced this with the new version because I didn't test documents that big.What page size are you using? Did you discover from what page threshold this happens? Providing that data would be helpful

Neil-G:
@diegomura , I'm using the default page size.  Or do you mean wrapping chunks of content in a `<Page>` component?  I think the largest pdf I was able to print without crashing was ~40 pages.  Here's a screenshot of the error message in the browser:
 
<img width="838" alt="screen shot 2018-09-25 at 6 18 39 pm" src="https://user-images.githubusercontent.com/8787807/46090949-c41c5a80-c17f-11e8-95fc-4c1d4c72601f.png">

Thanks and let me know how I can help you address this

diegomura:
Hey @Neil-G . Sorry for the absence. I was on vacations for awhile.
Can you test the last version to check if this still happens?
If it does, I would appreciate if you can provide me a way to reproduce this
Thanks 😄

AirborneEagle:
I am also now running into this issue.
[...]

PR #475, opened 2019-02-05
Alias yoga-layout-prebuilt to yoga-layout

DuncanMacWeb:
This seems a cleaner way of using yoga-layout-prebuilt — it’s _really_ yoga-layout so it would be nicer to import yoga-layout but have the build system take care of turning that into yoga-layout-prebuilt for us. Should have thought of this for #364 really.


diegomura:
Thank you @DuncanMacWeb !
I really like this change. In the future if we don't depend on `yoga-layout-prebuilt` anymore (ex. building yoga ourselves, setting some memory flags to fix #314 ) we wouldn't have to change the imports.

Issue #461, opened 2019-01-23
Add Ellipsis to Text


arahansen:
**Is your feature request related to a problem? Please describe.**
I'm looking to add ellipses to my text rendering. It looks like pdfkit handles this, but react-pdf uses a custom layout engine that (from what I can tell) does not call out to pdfkit. 

Is this a feature that react-pdf could support? I started digging in a little, but couldn't really find where to get started.

**Describe the solution you'd like**
An API that would be nice for this use case would be to handle it in the stylesheet ie:
```
const styles = StyleSheet.create({
  myText: {
    textWrap: 'nowrap',
    textOverflow: 'ellipsis'
  }
})
```

**Describe alternatives you've considered**
I suppose this could also be done as a prop on the Text component as well, but this seems a styling concern.

Alternatively, would it be possible to use pdfkit for an alternative Text layout? (for example, `SimpleText` or something along those lines). I'm assuming react-pdf is not using pdfkit because the layout there is not powerful enough, or not compatible with yoga.

**Additional context**

arahansen:
Seems like `createInstance` may be the escape hatch I'm looking for to implement my own Text component. Given the amount of heavy lifting the existing Text component is doing, would it be a futile effort to implement an alternate Text component that works well with the other react-pdf components?

arahansen:
[Here](https://github.com/diegomura/react-pdf/compare/master...arahansen:max-lines?expand=1#diff-98bcf02a8ea4fc975fe3929f67ac0265) is a (pretty hacky) prototype of something that I may be able to leverage what I'm looking for.

Basically, since textkit is already doing the heavy lifting to truncate the string, I would be able to use the first line found to determine the cutoff point. I still would need to be able to customize the hyphenization token, which would be a change in textkit afaict.

diegomura:
Thanks @arahansen ! 
I agree this is a cool feature to have, but unfortunately react-pdf does not support it yet. 

It's true we don't rely on `pdfkit` text and line breaking logic. We use [`textkit`](https://github.com/react-pdf/textkit) to perform these tasks, which was coded to produce much more higher quality text, plus Knuth & Plass line breaking algorithm and many other cool features. Imo this is something that should be implemented in textkit itself, since involves mutating the glyphs when text ran out of space.

Hope I can tackle this soon, but if you would like to give it a shot I'll be happy to help on the onboarding!

arahansen:
[...]

PR #482, opened 2019-02-08
feat(Text): add support for maxLines

arahansen:
adds maxLines as a prop to Text component. This isn't quite what I'm looking for from #461 but it would at least partially solve the following issue I'm seeing:

image

I want the text within my box to cutoff if it is going to overflow its container. With `maxLines` I could at least hard-code the height and compute the number of lines based on the text line height.

If you have any other suggestions for how to handle this, I'm open to them!

I also noticed that textkit used to support a maxLines attribute [here](https://github.com/react-pdf/textkit/pull/2/files#diff-cf1a9d87c4324bf6accbbbf1606a89f0L93) but the functionality was removed. I'm not sure if this was intentional since `maxLines` is still parsed for when adding default style attributes [here](https://github.com/react-pdf/textkit/pull/2/files#diff-e5d36af9f865022076b349138be46af1R30)


diegomura:
Hi @arahansen!

Thanks for looking at this. Sorry for removing the `maxLines` support from textkit. I didn't do it on purpose, but I had to almost rewrite textkit some month ago because it had some issues on it core that I couldn't solve otherwise. 

Ideally this should be something that's built in in textkit, in order to prevent layout text that then is no longer visible on the document, but for the moment I don't mind tackling this on the react-pdf layer and then adding it to textkit.

I'll publish a new version with this change now, and some others that were not published yet. Thanks!

Issue #481, opened 2019-02-08
Ideas for images: multiple resolutions, upgrade fetch library


DuncanMacWeb:
Here are some ideas to improve how we handle images, that I had while thinking about #478:

### Handling multiple image resolutions

The web has a standard that we could draw on to handle multiple image resolutions: [the `srcset` and `sizes` attributes](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images#How_do_you_create_responsive_images). Might this be useful to, for example, output a smaller PDF for screen than for print?

### Upgrade from `isomorphic-fetch` and alias `fetch`

Also it might be good to take the opportunity to upgrade to [`cross-fetch`](https://github.com/lquixada/cross-fetch) which seems to be better maintained than `isomorphic-fetch`. We could alias `fetch` to `cross-fetch` at the same time:

- Change [import references from `'isomorphic-fetch'`](https://github.com/diegomura/react-pdf/search?q=%22import+fetch+from+%27isomorphic-fetch%27%22) to `'fetch'`
- Install `cross-fetch` and alias it the same way we aliased `yoga-layout-prebuilt` in [this commit](https://github.com/diegomura/react-pdf/commit/0dfcc618e5b3bec45e8f3658a6fd5bb39752dde3)

diegomura:
Thanks for the ideas @DuncanMacWeb !
Both sound good to me. The second one should be really straight forward. The first one can be a bit more tricky to implement, but we can maybe use the same [media-engine](https://github.com/diegomura/media-engine) lib that we already use fo media queries.
Would be happy to make these changes but if you want feel free to do it!

DuncanMacWeb:
No worries! I can definitely upgrade the library but don’t have much bandwidth to tackle the full responsive images spec right now... will give it some further thought though :)

diegomura:
I just did the `cross-fetch` changes in a separate PR. Hope you don't mind.

That change is great, specially because `cross-fetch` solves some issues un React Native that `isomorphic-fetch`, for future RN integration. 

I kind of needed this change before moving forward with #485, so I can now have better (and fastest) tests using [jest-fetch-mock](https://github.com/jefflau/jest-fetch-mock)

diegomura:
[...]

PR #486, opened 2019-02-11
Use cross fetch

diegomura:
#481


DuncanMacWeb:
@diegomura you [said](https://github.com/diegomura/react-pdf/issues/481#issuecomment-462532509):

> My only concern here is that `isomorphic-fetch` is a `fbjs` dependency, which is also a `react-reconciler` dependency. So there's no really a way of getting rid of `isomorphic-fetch` completely right? what do you think @DuncanMacWeb ?

Good spot! I compared [`isomorphic-fetch`](https://npm.anvaka.com/#/view/2d/isomorphic-fetch) with [`cross-fetch`](https://npm.anvaka.com/#/view/2d/cross-fetch) over at npm.anvaka.com and it appears that the browser-side dependency whatwg-fetch happens to be pinned at `whatwg-fetch@3.0.0` for both... at present. So _right now_ I think both libraries would resolve to `whatwg-fetch@3.0.0` in the browser, and there would be negligible additional weight. But in future, particularly if a new version of `whatwg-fetch` is released, that could change... worth keeping an eye on 🤔

diegomura:
Thanks for your input and quick response @DuncanMacWeb !

Looking further to this issue, I realized that `fbjs` is no longer a dependency of `react-reconciler`. It was on the version we were using, but by just upgrading it we can completely get rid of `isomorphic-fetch` (plus some more changes fo remove `fbjs` as a `react-pdf` dependency).

In addition, _I think_ these changes will have a positive impact on `react-pdf` size itself, since before `node-fetch` required `iconv-lite`, and as you can see [here](https://bundlephobia.com/result?p=@react-pdf/renderer@1.3.0) it was a big part of this library size. I won't be sure if this is the case until I see the updated graph though 😅 .

diegomura:
All tests are green, and I did several manual checks (all good). 
I will merge this PR and release a new patch version 😄

bartzy:
@diegomura I'm still getting `iconv-lite` as a depedency because of `cross-fetch` that depends on `node-fetch`. Any idea if this was eventually added to a release?

PR #491, opened 2019-02-15
Fix type for toBlob to Promise

### LINKED ISSUE #490: toBlob type should be a Promise-wrapped Blob
**Describe the bug**
The Typescript type for `toBlob` should be a `Promise<Blob>` rather than a `Blob`

**To Reproduce**
Use the `toBlob` function in Typescript

**Expected behavior**
It should be typed as returning a promise

bkoltai:
Fixes #490


diegomura:
Thanks!! Will merge once test are green

PR #494, opened 2019-02-18
Fix wrong emoji urls.

agarant:
Some emojis like: ❤️,✈️ and 🛰️have a length > 1 but their url should only have one part. Because of it, some urls were incorrectly built and therefore not rendering in the final pdf document.

This PR:
- Add tests for fetch emojis
- Fix this case


src/utils/emoji.js
```diff
@@ -10,10 +10,11 @@ const regex = emojiRegex();
 
 const reflect = promise => (...args) => promise(...args).then(v => v, e => e);
 
-const fetchEmojiImage = reflect(resolveImage);
+const makeFetchEmojiImage = () => reflect(resolveImage);
```
diegomura: Can you explain the reason of this change?

src/utils/emoji.js
```diff
@@ -10,10 +10,11 @@ const regex = emojiRegex();
...
 
 const getCodePoints = string =>
   Array.from(string)
+    .filter(x => x !== '️')
```
diegomura: And this 😅

diegomura:
Thanks!! I wasn’t aware of this issue. I appreciate the PR. I made some questions but everything looks good for me

src/utils/emoji.js
agarant: Yes, this is not really obvious, I will add a comment. This is the fix. The picture below should make it clearer.

image

src/utils/emoji.js
agarant: When it is a constant, it is not possible to mock resolveImage as it uses the real version before jest is able to mock it. I will add a comment.

diegomura:
Thanks! Will appreciate those comments. But overall it looks great!

Just so you know, you can now mock `fetch` for testing cases such as these (as [here](https://github.com/diegomura/react-pdf/blob/master/tests/image.test.js#L76)). Your approach is good for me though!

I'm not sure if the last headers changes will affect how emojis are being loaded, but I'll take a look after this get's merged and fix it in case it's needed

agarant:
@diegomura Good to know! Thank you for this library and being so quick at responding to PRs! 👍

diegomura:
Thank you for taking care of this! All contributions are very welcomed 😄

PR #497, opened 2019-02-19
Update index.d.ts

### LINKED ISSUE #496: Compilation error due to stray semicolon.
Hey folks. Thanks for your great work!

It appears that the most recent release (v1.3.3) introduced a bug that is causing the following TypeScript compilation error:

```
/Users/path/to/project/node_modules/@react-pdf/renderer/index.d.ts
(97,9): The right-hand side of an arithmetic operation must be of type 'any', 'number', 'bigint' or an enum type.
```

Following this error message leads to the following bit of code in `index.d.ts`:

```typescript
    type SourceObject =
      | string
      | { data: Buffer; format: 'png' | 'jpg' };
      | { uri: string; method: HTTPMethod, body: any, headers: any }
```

Which seems to include an errant semicolon:

screen shot 2019-02-18 at 5 02 49 pm

Looks like this was merged in #485.

I think we're going to have to lock in v1.3.2 until this is patched.

Cheers,
Nick

diegomura:
Fixes #496

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -93,7 +93,7 @@ declare module '@react-pdf/renderer' {
 
     type SourceObject =
       | string
-      | { data: Buffer; format: 'png' | 'jpg' };
+      | { data: Buffer; format: 'png' | 'jpg' }
       | { uri: string; method: HTTPMethod, body: any, headers: any }
 
     interface BaseImageProps extends NodeProps {


diegomura:
Sorry @nmccrea-db for this inconvenience 😓 
Is that all right?

nmccrea-db:
No worries @diegomura. Looks good to me. Thanks!

diegomura:
I’m going to publish a new version today!

PR #504, opened 2019-02-22
Switch to hyphen

### LINKED ISSUE #489: Strange underline character is added at the end of each words that have "İ" (Turkish language character)
**Describe the bug**
Strange underline character is added at the end of each words that have "İ" (Turkish language character)

**To Reproduce**
Put "İ" character in Text. ex. "İçin". You can check that in https://react-pdf.org/repl?code=3187b0760ce02e00408a057025803c450298c0bc300500943807cf805030c00f0022230080b6198531165d400a02180e659a004f003619b006f61622003a0046200099080beed3a72a005430a58d3c54a8a231c8016187928c009d54c0066a83128d9b3a0064000e748c074d2a007a5d7d774a60de017760ba0666567602006e323200317028595b0c3e24683b3c0e0072732828000708002e20a0e02530002b39601110042547111e5cd95046209e669e14209124790820bf006b20806659004655800620c72c99802510452810007d50301b485720db03f47a700b8857005a4511256780770c796dd6590aa384a001a0e0489c3c4612044421a8c04afb438804a3055282526933b406086080e06000651318809960c36580b91e260f0120e2285470ba6692ad6251f8f8da10254e18b75bad4194344712cd63b132024e2c81290002f0c1c2002c02c9630faf93000085d047461c2004c2acd261f40041499f0c07092b0058985b08325a00ead9ad7c5c9091d9a3f941d290e86c2004488bba0638c2d5062c8bb2b300a05c1a3a4722c1b2d8f054442a1d05820b1031400000 

**Expected behavior**
To not render underline at the end of the words which has "İ" character

**Screenshots**

 - OS: MacOS
 - Browser [chrome and safari
 - React-pdf version v1.3.1

diegomura:
Fixes #489


oskarkook:
Hi! This change is slightly problematic because of this line in `hyphen`:
https://github.com/ytiurin/hyphen/blob/master/hyphen.js#L100

Since this line uses a literal unicode character, instead of matching by [unicode code point](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp#Regular_expression_and_Unicode_characters), it can cause issues. This should either be documented or possibly changed in `hyphen` itself.

oskarkook:
For info about my situation:
In my build process somewhere, `/\s|[\!-\@\[-\`\{-\¿]/` gets turned into `/\s|[\!-\@\[-\`\{-\\xbf]/`. I'm not too sure in which step it happens, as I am using Create React App, which uses Webpack under the hood.

`¿` can be mapped to `/xbf` code point in JS, however `/\s|[\!-\@\[-\`\{-\\xbf]/` is not a valid regex range. If `hyphen` were to match by unicode code point `\u00bf`, this issue would not surface.

oskarkook:
Further investigation leads me to: https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/config/webpack.config.js#L218

So this package will be buggy with CRA default configuration.

Issue #499, opened 2019-02-21
Support multiple font families within an element


DuncanMacWeb:
**Is your feature request related to a problem? Please describe.**
Approximately 2,140 characters are used in less than 2% of the content on our pages that we need to render to PDF. The remaining 98% of pages use just 200 characters. We don’t want to send the unneeded characters to render the 98% of PDFs that don’t need them.

**Describe the solution you'd like**
Improve the font fallback mechanism to pull characters from fallback fonts specified using the CSS `font-family: foo, bar, baz` syntax (or the equivalent in React Native style objects)

**Describe alternatives you've considered**
We have considered putting all the characters any of our content could possibly need into a single font file, but it would just be too large and a waste of network bandwidth.

Relates to #377, #402

SachinRupani:
Any update on supporting multiple font families / font files?

diegomura:
We already have multiple font support

PR #508, opened 2019-02-25
Resolve fonts for text runs against font-family (font stack) and @font-face rules

### LINKED ISSUE #402: Adding font weight support
Hi there,

I'd like to add font weight support to `react-pdf` and I was hoping a contributor could recommend an efficient strategy or inform me of any existing work before I dive in. 

Cheers
Luke

DuncanMacWeb:
Would resolve #402 and #499


DuncanMacWeb:
@diegomura just had a dig around in the repo and textkit, and this is what I found — do you think these would be good places to start with this PR? Anything I’m missing?

src/layout/fontSubstitution/index.js
```diff
@@ -20,6 +20,7 @@ export default () => ({ Run }) =>
     }
 
     shouldFallbackToFont(codePoint, font) {
+      // return the correct font for this code point based on font-face and font stack
```
diegomura: This seems right

src/stylesheet/index.js
```diff
@@ -45,6 +45,7 @@ const resolve = (styles, container) => {
 
   styles = flatten(styles);
   styles = resolveMediaQueries(styles, container);
+  // something here to resolve @font-face styles?
```
diegomura: I would leave `@font-face` css support for a separate issue, and stick with the `Font.register` API for the moment

diegomura:
I wrote some questions above, but I'm not sure if this is the correct way to handle this.

This is me thinking out loud, but let me detail how I think this would work.

**In order to support multiple font-families**

1. Make `style.fontFamily` an array instead of a string by splitting the original `fontFamily` value by `,`. Then on the [async loadFonts](https://github.com/diegomura/react-pdf/pull/508/files#diff-76514a85d16a6c21824a4b486dce5ec4R58) resolve all the needed fonts in these arrays (fonts should be previously registered by `Font.register`). Note that this process does not means that you are loading the fonts to the document, you are just downloading or resolving fonts for future uses.

2. In the [font substitution phase](https://github.com/diegomura/react-pdf/pull/508/files/72f1861f82fe68f8539d8acdaa3c891abebd16db#diff-f502aa66e72c2a4a5221974bf1966357R22) fallback to next font in the list instead of always to the default font. Hard to tell beforehand if this will require some change in `textkit` also, but it shouldn't be that much.

3. Only the resolved fonts will be included in the final document. It there was a font that the text engine didn't fallback to, it won't be included since this is not one until the very last rendering part by `@textkit/pdf-renderer` ([see here](https://github.com/foliojs/textkit/blob/master/packages/pdf-renderer/index.js#L94))

**Avoid shipping non-used characters**

Changes above can already mean a performance boost, since you can have the _200 most used chars_ in a separate font and the extra ones will only be included in the doc if a special char is present (as explained above). However, you will always ship some unused chars, specially if you fallback to several fonts.

Fortunately there might be something we can do to fix that. `fontkit` makes possible to create a [subset of a specific font](https://github.com/foliojs/fontkit#fontcreatesubset). So we can make use of that to only ship fonts with exactly the used glyphs, and no more. 

To be honest I never used that API of fontkit, and it's still unclear to me where to create the subsets. Maybe it can be in the very same `fontSubstitution` phase, or maybe it has to be done inside textkit itself. I won't be sure until I test it.

**font-face and font-width**

I see these points as separate things from what I just described. Adding `@font-face` support will basically mean to inspect text styles and make inner calls to `Font.register`. `font-width` I think would be a bit more complicated, and it's unclear to me yet how to support it.


Hope this helped in some way. Please share your thoughts about it and I will try to make some tests soon. I guess I will have to prioritize issues a bit because I feel I'm trying to work on several things at the same time, unable to finish any of them 😅

DuncanMacWeb:
Thanks so much for these thoughts and pointers, they are really helpful. I’m working on this right now so should be able to update this PR based on your comments :+1:

DuncanMacWeb:
On the prioritisation, maybe it’s possible to link GitHub issues/pulls to Trello or similar, and prioritise things there?

jeremyhon:
+1 for this! Our use case is we have to support a mix of latin and non-latin characters that might appear in the same string, so having a chain of fallback fonts is absolutely crucial. Any chance for this to go in? @diegomura

index.d.ts
```diff
@@ -47,7 +47,7 @@ declare module '@react-pdf/renderer' {
       // Text?:never,
 
       fontSize?: number,
-      fontFamily?: string,
+      fontFamily?: string | array,
```
jeremyhon:
```suggestion
      fontFamily?: string | string[],
```

index.d.ts
```diff
@@ -352,7 +352,7 @@ declare module '@react-pdf/renderer' {
 
     interface FontSource {
       src: string;
-      fontFamily: string;
+      fontFamily: string | array;
```
jeremyhon:
```suggestion
      fontFamily: string | string[];
```

src/layout/fontSubstitution/index.js
```diff
@@ -1,32 +1,78 @@
...
-    getFallbackFont().hasGlyphForCodePoint(codePoint)
-  );
+const toFontNameStack = (...fontFamilyObjects) =>
+  fontFamilyObjects
+    .map(fontFamilies =>
+      typeof fontFamilies === 'string'
+        ? fontFamilies.split(',').map(f => f.trim())
+        : Array.from(fontFamilies || []),
+    )
+    .flat()
```
jeremyhon: This broke for me on nodejs. Is it possible to use `flatten` from `ramda` instead?

jeremyhon:
Apologies if I overstepped boundaries, just realized I probably could have just left more comments instead of a review

diegomura:
Would be happy to take a look after those changes are applied!

jeremyhon:
Hi @DuncanMacWeb, are you still working on this? Otherwise I'd like to take over the work that you've done already, if that's at all possible. I've got some work-in-progress at this branch (building on your work that has been great): https://github.com/jeremyhon/react-pdf

@diegomura could I have a crack at this issue?

DuncanMacWeb:
@jeremyhon Thank you for the review, it’s very welcome. I’m working on this but alongside other priorities; sorry progress has been slow. Very happy if you’d like to build on this, please go ahead! Which branch are you using in your repo?

I may be able to let you push to our branch so your changes can be included in this PR; I’ll look in to that. Thank you!

Note that this PR also depends on changes to `@react-pdf/textkit` (as I’m sure you’ve seen).

DuncanMacWeb:
@jeremyhon @diegomura I’ve added you both as collaborators to our repo so you can push to the branch if you want.

lpww:
@jeremyhon Any updates on this? 

The `TypeError: Cannot read property 'glyphForCodePoint' of undefined` error that you saw is an existing issue that may not be caused by the changes in this pr. See https://github.com/diegomura/react-pdf/issues/667

You may want to test with Keifont, as that definitely works right now. I posted a link to a working Keifont example [here](https://github.com/diegomura/react-pdf/issues/667#issuecomment-536788254)

jeremyhon:
@lpww I've had to put this issue on the backburner for now. My branch is here: https://github.com/SDFE/react-pdf/tree/jeremyhon/font-face-fallback

I'm stuck right now because of two main issues:
1. A lot of the code appears to intertwine tightly with the textkit engine, which I'm not sure where it is and I only have access by modifying my `node_modules`
2. The textkit engine tries to treat a `FontSource` like an actual Font object, calling random methods that don't exist (but that exist on the underlying Font implementation inside FontSource.data). I got a little further ahead by monkey patching all the methods being called into the FontSource class, but then the PDF generated ended being unreadable.

@diegomura do you have any ideas?

lpww:
@jeremyhon Thanks for the update!

Just to clarify, I wasn't suggesting that Keifont may solve your use case, I was suggesting that it may help unblock your work on this specific pr. I think `TypeError: Cannot read property 'glyphForCodePoint' of undefined` is a different issue that also needs to be resolved

I have pretty much the same use case as you (supporting multiple languages in a single document) and I think there are 2 steps: 

1. Fix the `TypeError: Cannot read property 'glyphForCodePoint' of undefined` bug
2. Add fallback font support (this pr)

jeremyhon:
Would like to clarify - @diegomura do we use a forked implementation of pdfkit? or do we use the open-source one?

spik3s:
Hi guys! What's the status of this PR?

daihuaye:
@DuncanMacWeb Do you need help on the PR to move forward?

diegomura:
This PR is for 1.0. @DuncanMacWeb so sorry this got postponed. I'm not sure if this is something it's already fixed on 2.0 or if someone wants to keep working on this. If not I think I'll close it in a couple of days to keep the project clean. Again, thank you very much and I encourage you to keep working on the 2.0. Your contributions to this project were great :)

PR #513, opened 2019-03-02
Fix email links

ralphholzmann:
This fixes using `mailto` src's on links. The mailto protocol does not contain `://`, so the previous Regex was causing mailto link srcs like `mailto:test@example.com` to show up as `http://mailto:test@example.com`


diegomura:
Thanks!!

ralphholzmann:
Thanks so much! If you could release a patch or minor version at your convenience, that would be great!

Thanks again for such a great module.

diegomura:
I will tomorrow! Thanks again

PR #515, opened 2019-03-04
Add skype links

kuzivany:
Allow links such as `skype:username`

### DIFF
--- a/src/renderer.js
+++ b/src/renderer.js
@@ -12,6 +12,7 @@ const emptyObject = {};
 const shouldReplaceLink = (type, props) =>
   type === 'LINK' &&
   (typeof props.children === 'string' ||
+    typeof props.children === 'number' ||
     Array.isArray(props.children) ||
     props.render);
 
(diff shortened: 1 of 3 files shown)


diegomura:
Thanks. First time I see a link such as that. I'll have to run some tests before merging

kuzivany:
Learning never stops =D These types of links open the Skype app if it's installed

diegomura:
Hey! sorry for not responding soon. It's been some crazy days. I was thinking... do you mind changing this to support any type of protocols? There are also `whatsapp://` or `twitter://` and stuff like that thay may work on PDFs also. Never tried it, but I guess that if `skype://` works those should also work

kuzivany:
Hi,

I know what those days are like, haha.

I'm not sure whether `whatsapp://` or `twitter://` work, I've never used them myself but I could just make it pick up on a set of characters followed by the colon and foward slashes (`://`)

kuzivany:
Instead, or in addition to creating a long regular expression containing different strings like you have here:

```javascript
const PROTOCOL_REGEXP = /^(http|https|ftp|ftps|mailto|skype)/i;
```

possibly consider creating a regex like:

```javascript
const PROTOCOL_REGEXP_GENERIC = /^([a-z]+\:\/\/)/i;
```

I think a bit of research into some popular protocols or app links would lead down the right path. I don't have time at the moment but this is what I'd look into. Hope it's helpful

diegomura:
Yes! Exactly

PR #530, opened 2019-03-13
Apply objectFit styling correctly

DuncanMacWeb:
Components styled with `styled-components`, for reasons I don’t currently know 😉, seem to have their `style` prop passed in inside an array. To resolve this I have made the `Image` component take its `objectFit` property from `this.style` instead of `this.props.style`.


diegomura:
Thank you @DuncanMacWeb . LGTM also

kalimantos:
Could this change be released on npm please?

PR #538, opened 2019-03-19
Enable dynamic imports

DuncanMacWeb:
This could enable us to dynamically load Yoga, PNG.js, svgkit, or any other module that is heavy and should only be loaded if it’s needed.

WIP — needs tests, and doesn’t yet generate separate chunks, but `yarn run example:fractals` just about works ;)


diegomura:
This is great!
I've been smashing my head to the wall these last few days thinking about this. So thank you!

I do think this can be awesome for `PNG.js` and `svgkit`, but I have serious doubts about implementing this for Yoga. After all, you *cannot* run react-pdf without Yoga, so there's not much point of dynamically render it, don't you think? Th boilerplate this solution adds doesn't seem to be much needed for Yoga

DuncanMacWeb:
Well [Yoga is 25% of the total module weight](https://bundlephobia.com/result?p=@react-pdf/renderer@1.4.1), and compiled to WebAssembly it would probably be about half of that so the benefit is pretty significant. Making react-pdf work with asynchronous `import('yoga-layout')` means we can choose to import the WebAssembly or asm.js versions of Yoga depending on what the browser supports. I think [Wasm modules stream/parse in parallel](https://webassembly.org/docs/faq/), so the browser could maybe parse all react-pdf’s JavaScript at the same time as it parses the WebAssembly, making it even faster. Of course, we don’t yet have a working WebAssembly version of Yoga, so I guess we have a couple of options, as well as not doing anything:

- keep the async yoga support and bundle the asm.js version in our package, then the “async” import should be no different from the current behaviour and it would be easy to load the WebAssembly version in future
- separate out the async yoga functionality from rollup’s dynamic loading configuration, and move it to a separate PR

Happy with either one — what are your thoughts?

diegomura:
I'll repeat what I wrote in #694 , but it's also true for this one.

The real reason why I didn't merge this was because 2.0 was already in progress and could't keep up with everything to be honest. Adding this to v1 would meant also having to add it to v2 and after an entire year of working I couldn't think about adding new features. Sorry again :(

Would you be interested on doing this to v2? The implementation might be different though but it would be awesome to have this feature. I would understand if not.

diegomura:
@DuncanMacWeb thanks again for all your help! This PR points to 1.0. I'll close it but let me know if it's something you would like to work on the future 😄

PR #556, opened 2019-04-09
Add text decoration line-through support

### LINKED ISSUE #519: Add textDecoration line-through support
Hello,

I tried to use the textDescoration attribute as shown in the following example.
Line-through isn't working but is very important for invoices.
Am I doing something wrong or is there another solution?

Thank you in advance.

```
lineThrough : {
textDecoration: 'line-through',
textDecorationStyle: 'solid',
textDecorationColor: 'grey',
}, ... })

<Text style={styles.lineThrough}>{priceBeforeDiscount}</Text>
```

diegomura:
Fixes #519

### DIFF
--- a/src/utils/attributedString.js
+++ b/src/utils/attributedString.js
@@ -62,6 +62,9 @@ export const getFragments = instance => {
           underlineStyle: textDecorationStyle,
           underline: textDecoration === 'underline',
           underlineColor: textDecorationColor || color,
+          strike: textDecoration === 'line-through',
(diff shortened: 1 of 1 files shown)


mieszko4:
@diegomura This works when using `style` prop of `Text` but it seems to me that it is missing bindings to `@react-pdf/styled-components`.

For example the following does not work for me:
```es
const StrikedText = styled.Text`
  text-decoration: line-through;
`;
```

PR #560, opened 2019-04-11
Update typings

### LINKED ISSUE #551: Canvas: interface is outdated
Version: 1.4.2
Stack: React App with Typescript

Hello, I tried to use the new feature of `react-pdf` called Canvas but interface included in `node_modules/@react-pdf/renderer/index.d.ts` seems to be outdated. There is no declarations for Canvas module and Typescript doesn't see this one. Could you please update the interface?

This is what I see: https://i.imgur.com/GD5xIGW.png

diegomura:
Fixes #551

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -166,6 +166,17 @@ declare module '@react-pdf/renderer' {
 
     class Note extends React.Component<NoteProps> {}
 
+    interface CanvasProps extends NodeProps {
+      debug?: boolean;
+      paint: (
+        painter: any,
+        availableWidth: number,
+        availableHeight: number
+      ) => null;
+    }
+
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -273,6 +284,12 @@ declare module '@react-pdf/renderer' {
       register: (options: {
         family: string;
         src: string;
+        fonts?: [font: {
```
CodeWitchBella: This is syntax error. Assuming you want array of fonts

```suggestion
        fonts?: {
          src: string;
          fontStyle: string;
          fontWeight: string | number;
          [key: string]: any;
        }[];
```

this is correct syntax

diegomura: Thanks. I’ll fix it asap

PR #585, opened 2019-04-26
Image fetching middleware

maxaggedon:
I use this to do some image processing. It is not intended to be merged as is, but rather to discuss if such a feature should be added to the package.

Let me know what you think !

### DIFF
--- a/src/elements/Image.js
+++ b/src/elements/Image.js
@@ -105,7 +105,8 @@ class Image extends Base {
     }
 
     try {
-      this.image = await resolveImage(this.src, {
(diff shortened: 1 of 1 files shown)


DASPRiD:
I'd have a need for exactly that right now. I'm currently converting an SVG to a JPEG on the fly via canvas, and trying to figure out a good way to wrap the Image component to do that asynchronously.

diegomura:
Hey guys! I would be happy to merge this. Could someone update the docs accordingly? https://github.com/react-pdf/site/blob/master/docs/image.md. I promise I'll merge and publish after that

Issue #378, opened 2018-11-15
Memory Leak


coreyweidenhammer:
Before I get into the issue, I just wanted to thank @diegomura and others for your work and responsiveness on this wonderful library! Your efforts are greatly appreciated.

**OS:**
- Amazon Linux aws-elasticbeanstalk-amzn-2018.03.0.x86_64-nodejs-hvm-201809210902 (ami-00d787dad6adc69a1) on AWS Elastic Beanstalk
- Also, macOS 10.13.6

**React-pdf version:**
1.0.0-alpha.24, 1.0.0-alpha.25
Node v8.9.3

**Description:**
We are consistently observing a memory leak, with the memory footprint of our node process growing linearly with each generated PDF until eventually exhausting the resources on our servers.

My team saw https://github.com/diegomura/react-pdf/issues/165 (now closed), but we are uncertain as to whether what we are experiencing is related or not. As noted above, we have tried using both alpha.24 and alpha.25 with the same results.

We first noticed the memory leak on servers running in AWS via ElasticBeanstalk (Amazon Linux) but have since been able to recreate it on our local development environments running macOS 10.13.6. We attempted isolating the source of the leak ourselves but have not yet come up with anything conclusive. We're not even sure if the memory leak is (as before) related to a dependency vs the react-pdf code itself, but we are hoping you might be able to identify something more readily.

**How to replicate issue:**
We created a small project with limited dependencies that should allow you to recreate the issue as we see it. Zip attached here:
[pdfGenerator.zip](https://github.com/diegomura/react-pdf/files/2586375/pdfGenerator.zip)

There are instructions in server.js, repeated below:

RUN $ yarn / npm install and then run: $ npm run start-with-chrome-inspect
1. In a Chrome browser window, go to http://localhost:3000/.
2. In a new tab go to: chrome://inspect/#devices and click inspect under the target.
In the devtools window, select the Memory tab at the top.
  Select 'Heap snapshot'.
  Click 'Take snapshot'.
4. Open http://localhost:3000/pdf in a new tab.
5. Go back the devtools window and click the dark circle icon at top-left above the Profiles pane to generate another snap.
6. Repeat steps 4 and 5 repeatedly and you can see the retained memory increase. (JSArrayBufferData especially)

An alternate path to viewing the memory leak at a higher level is to monitor via pm2:
 $ yarn / npm install pm2
 $ pm2 start server.js
 $ pm2 monit
 -- proceed to access http://localhost:3000/pdf repeatedly
[...]

PR #591, opened 2019-05-06
Fix memory leak

### LINKED ISSUE #314: "Cannot enlarge memory arrays" error when document is big.

### LINKED ISSUE #590: toBuffer doesn't handle error in render()
`toBuffer` doesn't call `await` on render... thus any exception will be unhandled... such as "Font family not registered".

https://github.com/diegomura/react-pdf/blob/1bc45ae3da2740eb008df8b1486454cd0753fe65/src/index.js#L63

```js
{"level":30,"time":1556912171434,"msg":"unhandledRejection caught","pid":78,"hostname":"6f3e3bc080d6","name":"print-react","v":1}
{"level":60,"time":1556912171434,"msg":"error caused exit","pid":78,"hostname":"6f3e3bc080d6","name":"print-react","err":{"type":"Error","message":"Font family not registered: Open Sans. Please register it calling Font.register() method.","stack":"Error: Font family not registered: Open Sans. Please register it calling Font.register() method.\n    at getFont (/usr/src/app/node_modules/@react-pdf/renderer/dist/react-pdf.cjs.js:2536:11)\n    at Object.load (/usr/src/app/node_modules/@react-pdf/renderer/dist/react-pdf.cjs.js:2548:16)\n    at Document.loadFonts (/usr/src/app/node_modules/@react-pdf/renderer/dist/react-pdf.cjs.js:3926:30)\n    at Document.loadAssets (/usr/src/app/node_modules/@react-pdf/renderer/dist/react-pdf.cjs.js:3982:29)\n    at Document.render (/usr/src/app/node_modules/@react-pdf/renderer/dist/react-pdf.cjs.js:4039:18)\n    at Root.render (/usr/src/app/node_modules/@react-pdf/renderer/dist/react-pdf.cjs.js:60:25)\n    at Object.toBuffer (/usr/src/app/node_modules/@react-pdf/renderer/dist/react-pdf.cjs.js:4318:15)\n    at renderToStream (/usr/src/app/node_modules/@react-pdf/renderer/dist/react-pdf.cjs.js:4352:23)\n    at Promise (/usr/src/app/src/print-react/publish.js:147:22)\n    at new Promise (<anonymous>)"},"v":1}

```

bhunjadi:
Memory fix implementation based on findings and discussion in issue #378.
This should not be seen as a definite solution, but it should be a place to start.

Each node has a cleanup function which performs two things:
1. calls `unsetMeasureFunc` - this solves the problem with leaked instances (Text, Image)
2. calls `Yoga.Node.destroy(this)` - after testing export of a larger document with 1000 images with the same `src` I always got the error "Cannot enlarge memory arrays" and adding `Yoga.Node.destroy` solved this error. This is something that happened in our app after generating around 70 pdfs (server side) with size of 15-20 pages. 

So it is possible this might fix #314. Also fixes #590

### Implementation
Idea is that on each `remove()` called from wrapping algorithm we clean the removed child.
And after rendering, root instance calls `cleanup()` on the children:
```
async render() {
    this.instance = new PDFDocument({ autoFirstPage: false });
    await this.document.render();
    **this.cleanup();**
    this.isDirty = false;
}
```


bhunjadi:
Argh, thanks. 
Looks like permissions were changed because of samba mount.

diegomura:
Awesome! Thank you very much! @n6g7 if you can test it that would be great. I'll try to test it today also 😄

n6g7:
So this PR does fix the leak in [the reproduction gist](https://gist.github.com/n6g7/4d9d9af8eb5b943b61c89c1c6fa29a3b) but not in [my original project](https://github.com/n6g7/httpdf). 🤔
I might be overlooking something, I'll look into it more tomorrow.

bhunjadi:
I tried with your project and to me it looks like it is ok.
I used `yarn dev`, generated 1 pdf (/demo), made a heapdump, then generated another 10 docs and then made a heapdump again. I see no leaks there regarding leaked Root, Text, View etc.

Could you post your heapdumps when you do the tests again? Thanks

n6g7:
Yep you're right, it looks like it's working! :D
I actually was still using `@react-pdf/node` instead `@react-pdf/renderer` for some reason so was 
running unpatched code unknowingly.

diegomura:
So, I tried this today (sorry for taking that much!) and I found issues when the document re renders based on a changing state:

Kapture 2019-05-11 at 16 05 03

This is a very simple app that renders an image and a counter (that increases when clicking the button). First renders work, but afterwards it doesn't and it breaks the layout process.

What we should do is only cleaning the [subpages](https://github.com/diegomura/react-pdf/pull/591/files#diff-76514a85d16a6c21824a4b486dce5ec4R187) nodes, not the children since these are reused on each render to calculate the next subpages. I tried this approach and fixed this. I'll push it here. Could you please try if the gist and your projects to see that the memory leak is still gone?

n6g7:
Hey @diegomura, I applied your commit on both the gist and my project and the leak was back. :(

diegomura:
@n6g7 got it. We should definitely cleanup children in node (since those nodes are also used once in that env), but not on the web. Will appreciate a lot if you can test again in the gist and your project!

bhunjadi:
Looks to me that with latest commit everything is OK - no memory leak in the gist nor my project!

n6g7:
Yes, same results here, no leak in gist or my project with this latest commit. :)
Well done!

diegomura:
Thank you! Merging this now and I’ll publish a new version later

ameya-synerzip:
This happening again on re-rendering the pdf I have.
I am not sure how the disposeOnUnmount works here.

PR #605, opened 2019-05-17
Page percentage units (vh/vw)

olistic:
This PR adds support for `vh` and `vw` units. These units are useful to define length values relative to the size of the document's page.


diegomura:
Docs already updated for this change: https://github.com/react-pdf/site/commit/f4b756c6cb106772ae49b0f061f80455b8eb6a7e (Not published yet)

diegomura:
Added support for borders, margin, padding, transform-origin and position using new units. 

I'm aware we are repeating some logic between different regex throughout the project, but they are not always exactly the same (some should support `%`, others not. Some have suffixes, others not). I don't think that's too painful since it's very rare to add new units

PR #618, opened 2019-05-29
Updated Font.Register src param in examples

dfdumaresq:
I've updated the params in Font.Register as recommended by the error message, which was pretty annoying since I wanted to quickly test the App examples and could not get any output.

The examples 'just work' now.


diegomura:
Thank you!

dfdumaresq:
And thank you and the rest of the team for your important work!

PR #656, opened 2019-06-27
Fix the HyphenationCallback Typescript definition

GabeAtWork:
According to the doc, the hyphenation callback is given the word as a
`string` and must return an `array` of `string` (`string[]` in TS).
That's how the features works according to my tests as well.

This commit fixes the type definition so as to accept the right type for
`word` (`string`).


diegomura:
Thank you very much! I have to run now but I'll take a look at the other PR asap!

PR #665, opened 2019-07-15
Add inline image unit test

kacperturon:


kacperturon:
This is still work in progress, need to be able to distinguish inline emojis from inline images.

diegomura:
Be careful with the `package-lock.json` file. We use yarn and already have that lock file

PR #671, opened 2019-07-26
[typescript] Style typings

rzymek:
Complete, I hope, typings for Style interface. I went through the code to find the actually used and supported values. Some where not that easy so spot, as they span across yoga-layouts and pdfkit. Thus the `// ?`.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -4,8 +4,117 @@ declare module '@react-pdf/renderer' {
   import * as React from 'react';
 
   namespace ReactPDF {
-    interface Style {
-      [property: string]: any;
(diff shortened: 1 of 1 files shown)


diegomura:
Wow this is so great! Thank you very much. There's also a list of supported CSS properties in [the site](https://react-pdf.org/styling#valid-css-properties). Let me know if the list is complete and I
ll merge this!

rzymek:
Yes, the interface contains all the properties from the site docs. That's
what I've started with.

diegomura:
Awesome. Thanks

PR #694, opened 2019-08-17
Allow user to supply their own Yoga implementation

CodeWitchBella:
# What
This would allow to use yoga-dom instead of yoga-layout-prebuilt.
It currently does not work because yoga-dom does not expose all the things react-pdf needs but it's a start.

# How
Remove dependence on yoga from all except entrypoint files. Entrypoint files import yoga and pass it to root node from where other nodes load it.

# How to use
There is new build browser-custom. Import `@react-pdf/renderer/dist/react-pdf.browser-custom.es.js` that instead of `@react-pdf/renderer` (you can probably set this as alias in webpack)

This build exposes single default export - function which takes Yoga and returns everything that is ordinarily exported from `@react-pdf/renderer`

# Tests
I did not add new tests but made sure that old ones pass.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -153,9 +153,30 @@ const browserProdConfig = {
   plugins: browserConfig.plugins.concat(terser()),
 };
 
+const browserCustomConfig = {
+  ...configBase,
+  input: './src/custom-dom.js',
+  output: [
+    getESM({ file: 'dist/react-pdf.browser-custom.es.js' }),
+    getCJS({ file: 'dist/react-pdf.browser-custom.cjs.js' }),
+  ],
+  plugins: [...getPlugins({ browser: true }), ignore(['fs', 'path', 'url'])],
+};
+
+const browserCustomProdConfig = {
+  ...browserConfig,
+  output: [
+    getESM({ file: 'dist/react-pdf.browser-custom.es.min.js' }),
+    getCJS({ file: 'dist/react-pdf.browser-custom.cjs.min.js' }),
+  ],
+  plugins: browserConfig.plugins.concat(terser()),
+};
+
 export default [
   serverConfig,
   serverProdConfig,
   browserConfig,
   browserProdConfig,
+  browserCustomConfig,
+  browserCustomProdConfig,
 ];
--- a/src/custom-dom.js
+++ b/src/custom-dom.js
@@ -1,0 +1,163 @@
+/* eslint-disable no-unused-vars */
+import React from 'react';
+
+import warning from './utils/warning';
+import createReactPDFIndex from './index';
+
+export default function createReactPDFDom(Yoga) {
+  const {
+    pdf,
+    View,
+    Text,
+    Link,
+    Page,
+    Font,
+    Note,
+    Image,
+    Canvas,
+    version,
+    StyleSheet,
+    PDFRenderer,
+    createInstance,
+    Document: PDFDocument,
+  } = createReactPDFIndex(Yoga);
+
+  const flatStyles = stylesArray =>
+    stylesArray.reduce((acc, style) => ({ ...acc, ...style }), {});
+
+  const Document = ({ children, ...props }) => {
+    return <PDFDocument {...props}>{children}</PDFDocument>;
+  };
+
+  class InternalBlobProvider extends React.PureComponent {
+    state = { blob: null, url: null, loading: true, error: null };
+
+    constructor(props) {
+      super(props);
+
+      // Create new root container for this render
+      this.instance = pdf();
+    }
+
+    componentDidMount() {
+      this.renderDocument();
+      this.onDocumentUpdate();
+    }
+
+    componentDidUpdate() {
+      this.renderDocument();
+
+      if (this.instance.isDirty() && !this.state.error) {
+        this.onDocumentUpdate();
+      }
+    }
+
+    renderDocument() {
+      this.instance.updateContainer(this.props.document);
+    }
+
+    onDocumentUpdate() {
+      const oldBlobUrl = this.state.url;
+
+      this.instance
+        .toBlob()
+        .then(blob => {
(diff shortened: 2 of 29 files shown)


DuncanMacWeb:
Ooh interesting @CodeWitchBella. I drafted something quite like this in #538 but you’ve taken it a step further with the custom bundle — allowing current or custom Yoga to be used without requiring the consumer’s code bundler to understand dynamic imports 👍 

I think this would be easier for @diegomura to accept if we could create a working WebAssembly build of Yoga that demonstrates the benefit... have you taken steps towards that at all?

CodeWitchBella:
@DuncanMacWeb I have proof of concept which is able to (almost correctly) render the PDF I use this for. Will try to clean it up and publish it somewhere. 

It is based on [yoga-dom](https://github.com/vincentriemer/yoga-dom) and I did two things:
1) exported missing methods
2) made a shim which adapts yoga-dom's bindings to those which are exported from yoga-layout
   - main difference here is that yoga-dom uses getters and setters which is IMO better than yoga-layout but also somehow deals incosistently with units...

I implemented maybe third of the API but it is mostly manual and hard to do without proper documentation on how it should work (both from yoga-layout and from yoga-dom) so I have to read sources and create reproduction cases where it is used.

CodeWitchBella:
@DuncanMacWeb Uploaded my (very work in progress) version of wasm bindings to [github repo](https://github.com/CodeWitchBella/yoga-wasm) and published to npm as [@codewitchbella/yoga-wasm](https://www.npmjs.com/package/@codewitchbella/yoga-wasm)

It exports promise (same as yoga-dom) but adds methods used in react-pdf in my PDFs.

Also commit above allows to accept promise or function returning promise (so it accepts `() => import(...)`) instead of only plain yoga. This makes it simpler to use.

<details><summary>Usage (click to expand)</summary><p>

```js
import mk from '@react-pdf/renderer/dist/react-pdf.browser-custom.es.js'
import pYoga from '@codewitchbella/yoga-wasm'

const rpdf = mk(pYoga)

const {
  pdf,
  View,
  Text,
  Link,
  Page,
  Font,
  Note,
  Image,
  Canvas,
  version,
  Document,
  PDFViewer,
  StyleSheet,
  PDFRenderer,
  BlobProvider,
  createInstance,
  PDFDownloadLink,
} = rpdf
export default rpdf

export {
  pdf,
  View,
  Text,
  Link,
  Page,
  Font,
  Note,
  Image,
  Canvas,
  version,
  Document,
  PDFViewer,
  StyleSheet,
  PDFRenderer,
  BlobProvider,
  createInstance,
  PDFDownloadLink,
}
```

and then you just import this module instead of `@react-pdf/renderer`
</p></details>

diegomura:
Hey! Sorry for being unresponsive. Like this change! I eventually plan to do something like this for all other heavy dependencies also in case a user wants to reduce bundle size if, for instance, knows beforehand that he won't need PNG images.

I'll try to find the time to test it soon. For what I understood, it does not change the way the library is consumed right? It just adds a new build you can use if you want to inject your own yoga

CodeWitchBella:
> Hey! Sorry for being unresponsive. Like this change! I eventually plan to do something like this for all other heavy dependencies also in case a user wants to reduce bundle size if, for instance, knows beforehand that he won't need PNG images.
That would be very cool

> 
> For what I understood, it does not change the way the library is consumed right? It just adds a new build you can use if you want to inject your own yoga
Correct

CodeWitchBella:
The [@CodeWitchBella/yoga-wasm](https://github.com/CodeWitchBella/yoga-wasm) package is now at stage when I think that it should be fully compatible with upstream yoga-layout.

Also got another idea how this could be implemented - we can remove the custom build and then user just puts alias `'yoga-layout-prebuilt': '@codewitchbella/yoga-wasm'` to their webpack config and if react-pdf awaits the promise everything should just work™

Skywalker13:
@diegomura I hope that you will merge this stuff because the yoga-layout prebuilt package has a too low memory limit (unusable with large pdf), the original yoga-layout package is very annoying because it needs to build the sources (and it's broken with too recent version of node)... the @CodeWitchBella way looks like the best way. WASM is the way to go..

diegomura:
@CodeWitchBella hey. Sorry I never merged this. Really, I feel bad. 
The real reason was because 2.0 was already in progress and could't keep up with everything to be honest. Adding this to v1 would meant also having to add it to v2 and after an entire year of working I couldn't think about adding new features. Sorry again :(

Would you be interested on doing this to v2? The implementation might be different though but it would be awesome to have this feature. I would understand if not.

CodeWitchBella:
> @CodeWitchBella hey. Sorry I never merged this. Really, I feel bad.
> The real reason was because 2.0 was already in progress and could't keep up with everything to be honest. Adding this to v1 would meant also having to add it to v2 and after an entire year of working I couldn't think about adding new features. Sorry again :(
> 
> Would you be interested on doing this to v2? The implementation might be different though but it would be awesome to have this feature. I would understand if not.

No problem. I have seen that you recently release v2, great work as always :slightly_smiling_face: If I find time I would love to redo this for v2, but no promises, since between school work and money work I am pretty busy too :slightly_smiling_face: 

But this approach was overly complicated anyway and in combination with some features of new emscripten versions this could be way simpler, but that would also mean porting yoga to wasm again, since last time I tried it was difficult to update to new upstream yoga :slightly_frowning_face: 

Anyway, thanks for your hard work, since I don't thing the app I use this on would be possible without this - at least not as streamlined as it is currently :+1:

diegomura:
Thanks @CodeWitchBella !! You're definitely more experience with emscripten than I am so any future help would be much appreciated. I do understand it's hard to keep up with everything, so of course no preasure. Thanks again for showing that much enthusiasm for the project and glad it's useful for your app 😄

PR #708, opened 2019-09-06
typing improvemnt: allow string in most number field (e.g. 1mm)

rzymek:


diegomura:
All merged! Thanks for your effort. I’ve been traveling these past few days and couldn’t dedicate that much time to the project but I’ll soon!

Issue #212, opened 2018-03-22
Implement Goto links ( ref links )


crubier:
Hi, I saw that pdf kit supports setting « goto » références and inks to these references. This feature would be very useful for react pdf. The lack of this feature is actually quite blocking for us at the moment. Thank you very much for the good work !

diegomura:
Hi @crubier !
Thanks for reporting this.
I think this is not a huge feature, and I might be able to work on it in these days. I imagine something like this:

```jsx
<Link src={{ page: 5 }}>Click me</Link>
```

What do you think?
This API enables us to extend this functionality in the future if we want to. 
Let me know your thoughts!

crubier:
I think we need to have something a bit closer to the pdf kit api, which allows to add names anchors (or refs) in the document, and add links to these refs. This allows to provide links to precise elements whose page might be unknown (due to page wrap for example)

mike-down:
Hi @diegomura 
thank you for the quick respond 
I think it will be more simple and natural. like this :
```html
<Link src={5}>Click me</Link>
```
if `src` is a number then we can say it's a page number.

diegomura:
Good point @crubier . I didn't know that pdfkit enables you to do that.
Either way, accepting an object as `src` will enable to link to a page number (maybe we should call it `pageNumber` instead of  just `page`), or a particular `ref` inside the document. 
Implementing refs might be a bit more difficult (all depends on what pdfkit already does) than just linking to a page number.

@mike-down I don't find passing just the page number more natural 🤔 . It's not clear at a first glance what are you trying to link. We can support this however (accepting string, number and an object), but I wouldn't promote this as the official API.

diegomura:
I was looking at `pdfkit` code, and I couldn't find any hint that linking by ref internally is doable: https://github.com/devongovett/pdfkit/blob/master/lib/mixins/annotations.coffee#L28

It seems to me that we can link to a page number, or a URI. Am I missing something @crubier ?

javierfernandes:
Hi guys, 

I guess that this PR needs to be merged in pdfkit to build some "reactish" support on top of it.
Right ?

https://github.com/devongovett/pdfkit/pull/720

Seems like they call them "named destinations" in pdf (I'm not an expert pdf format :P)

Cheers

sanohin:
@diegomura  What do you think about switching to pdfkit fork to make this feature possible to implement? seems like the main repo isn't actively maintained

diegomura:
[...]

PR #744, opened 2019-10-28
GoTo functionality

RuslanNikolov:
I thought I'd share with you some slight changes to @react-pdf that would enable the GoTo functionality referenced in issue 212 -> https://github.com/diegomura/react-pdf/issues/212.
Also, in order for it to work I had to change @react-pdf/pdfkit a little bit as well. I had to just apply the changes from this commit -> https://github.com/foliojs/pdfkit/pull/947, that has been merged into foliojs/pdfkit some months ago.

Can you look at the src/layout/fontSubstitution/standardFont.js. Instead of PDFFont I exported PDFFontFactory, because I could not see a method ‘open’ on PDFFont.


kmooney517:
I have found that the goTo functionality works as described above when the generated pdf is viewed in browser - but the internal linking is not working when the pdf is viewed in adobe or preview. Any thoughts?

PR #803, opened 2019-12-26
Children should inherit text transform

omgovich:
I found out that `text-transform` isn't inheritable propery in React PDF. But it should be (See [W3C info](https://www.w3.org/TR/CSS21/propidx.html)).

I updated `resume` example to debug this error (moved `textTranform` from each child to whole `detailColumn`):

### Before:
before

### After:
Xnip2019-12-26_13-35-36


diegomura:
Thanks!

PR #811, opened 2020-01-16
Call onRender again in v2

tomaswitek:
@diegomura  It looks like we forgot to call `onRender` callback.
I've left out following line, not sure how important it was:
```
const layoutData = container.document.getLayoutData();
```

Here is a `master` version for comparison: https://github.com/diegomura/react-pdf/blob/master/src/index.js#L25


diegomura:
Thanks Tomas! Let's leave `layoutData` out for the moment, it's not important

PR #815, opened 2020-01-22
Add scope definition to read.me

smithywick:
Adding the scope definition from react-pdf.org to the repository. This will help avoid confusion about how the library is used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,8 @@
   </p>
 </p>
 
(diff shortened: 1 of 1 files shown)


diegomura:
Hey @smithywick . Thank you for this, but I'm not convinced to include it. 
I think the scope is pretty clear, and also _render things_ sounds a bit odd to me. 
What caused confusion that makes you think this would be helpful? Happy to find another way

PR #817, opened 2020-01-28
Empty the render queue immediately on unmount

tomaswitek:
This PR helps to clean the queue on blob and canvas unmounts.
Without it I was getting console warnings:

screenshot

`Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.`

### DIFF
--- a/src/canvas.js
+++ b/src/canvas.js
@@ -163,6 +163,10 @@ export class CanvasViewer extends React.Component {
     }
   }
 
+  componentWillUnmount() {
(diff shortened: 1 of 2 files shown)


diegomura:
Thank youuu!

tomaswitek:
Hey @diegomura could you pls release a new beta version which includes this PR? Thank you!

PR #846, opened 2020-03-13
fix issue 777

vincepich:


yuchi:
I’m not sure the correct approach is to change the access to `this.family`. Since `FontSource` and all other properties use the `font___` prefix, probably it’s the property name that should change.

PR #864, opened 2020-04-03
Remove default underline for links

MariuszKogut:
Note: This PR is based on V2 branch.

I noticed that setting text-decoration: none for links is not working in v2 any more. So I removed the default underline text-decoration from the link styles.

### DIFF
--- a/src/layout/resolveStyles.js
+++ b/src/layout/resolveStyles.js
@@ -10,7 +10,6 @@ import resolveMediaQueries from '../stylesheet/resolveMediaQueries';
 
 const LINK_STYLES = {
   color: 'blue',
-  textDecoration: 'underline',
 };
 
 /**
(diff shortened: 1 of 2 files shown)


tsnobip:
I think that's related to this from https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration: 
> Text decorations are drawn across descendant text elements. This means that if an element specifies a text decoration, then a child element can't remove the decoration. For example, in the markup <p>This text has <em>some emphasized words</em> in it.</p>, the style rule p { text-decoration: underline; } would cause the entire paragraph to be underlined. The style rule em { text-decoration: none; } would not cause any change; the entire paragraph would still be underlined.

To remove the underline, simply use:
```js
textDecoration: "unset"
```

diegomura:
v2 links do not have styles anymore. Sorry for not merging this before!

PR #884, opened 2020-05-02
Added the ability to register a font by passing the font as a buffer

dberringer:
Currently users can register fonts with a url or a file path.  This simple pull request adds the additional capability for users to register a font by passing a buffer to source.  This is essentially the same thing that happens behind the scenes when the user passes a url to source, however, it gives the user additional flexibility for how the font is fetched.

-Added additional test for this method of registering fonts
-Updated text example document to demo this functionality


dberringer:
Upon further consideration I think this could be done a little cleaner.  I will resubmit a new pull request.

PR #885, opened 2020-05-03
Added the ability to register a font by passing the font as a data url

dberringer:
Currently users can register fonts with a url or a file path. This simple pull request adds the additional capability for users to register a font by passing a data url to source. This is essentially the same thing that happens behind the scenes when the user passes a url to source, however, it gives the user additional flexibility for how the font is fetched.

-Added additional test for this method of registering fonts
-Updated text example document to demo this functionality

Please let me know if you need any other updates to approve this.  Thanks @diegomura


dberringer:
Hey @diegomura.  Have you had a chance to take a look at my pull request?  Can get this into a 1.6.9 release?

Issue #899, opened 2020-05-12
Support multiple layout passes (Pull request included)


rmmr:
I wanted to measure the height and width of my components through React refs and useEffect hook. Further more I wanted to modify the layout according to these measurements. This wasn't supported by react-pdf (v2).

I modified the reconciler and layout renderer to perform multiple passes, running at least 2 times, and at most N times comparing the previous layout to the final layout. Once these are equal it will render the actual PDF.

I also changed the Fractals example to showcase this usecase:

examples/fractals/Fractals.js

```
...
 const [bounds, setBounds] = useState(null);
  const ref = useRef();
  useEffect(() => {
    if (ref.current.prevBox !== undefined) {
      const { width, height } = ref.current.prevBox;
      if (bounds === null) {
        setBounds({ width, height });
      }
    }
  }, []);

  return (
    <View ref={ref} style={styles[direction]}>
      <Fractal direction={toggle(direction)} steps={steps - 1} />
      <View style={fractalStyle}>
        <Text style={styles.text}>
          {steps}
          {bounds !== null &&
            ` (${Math.round(bounds.width)}x${Math.round(bounds.height)})`}
        </Text>
      </View>
    </View>
  );

```
<img width="542" alt="Screenshot 2020-05-12 at 12 42 18" src="https://user-images.githubusercontent.com/1424733/81674839-1e018f00-944e-11ea-930b-f56c019f4e0f.png">

The included pull request is the best implementation I could come up with. One major problem is the async nature of the layout renderer. I did not find a way to sync it with the reconciler in order to run it before every mount effect. That is why this solution creates a new container on every pass.

puchm:
What is the state of this? It would be really helpful if it was possible to measure the components.

PR #900, opened 2020-05-12
Support multiple render passes

rmmr:
See issue https://github.com/diegomura/react-pdf/issues/899


diegomura:
Will take a deeper look this asap but looks promising! Thanks

rmmr:
That would be great. In the meantime I have made more improvements to this new render process.  The latest state (including support for concurrent mode) can be found in the v2 branch.

pcraciunoiu:
This looks like an interesting approach. I have an issue now where I'd like to repeat a header when a section (table) wraps to the next page. It seems like this would work, though I tried the branch and it threw an error for the PDF I have working on v2-beta6.

Calculating the heights of the elements also seems like it would work, but I don't see a way to programmatically retrieve the height either until after render (in v1 using `layoutData`, but not in v2). Or by manually looking at the debug measurements.

Can either of you think of another way to insert (repeat) a header when an element goes to the next page? Seems like a common use case, but I didn't find any easy way of doing it.

Issue #905, opened 2020-05-15
Support React Concurrent Mode (Pull request available)


rmmr:
First of all this issue's pull request is a continuation of my work as described in issue #899. 

Why support concurrent mode? Let me explain my initial use-case:

I want to use jimp to fetch images and transform them to greyscale. Since image buffers are supported I can prepare these buffers outside of the render process and pass these buffers to the root component and use them further on.

But this is a messy solution (imo), and I want to use React hooks for this. This way I can create awesome re-usable components for fetching and altering images. Just like you would for fetching data, etc. (also not supported previously).

I modified the "images" example to showcase this:

examples/images/index.js
```
const doc = () => {
  const imageBuffer = useTransformImage(
    'https://upload.wikimedia.org/wikipedia/commons/thumb/4/4b/Borah_Peak_ID_2-22-15.JPG/1280px-Borah_Peak_ID_2-22-15.JPG',
  );
  return (
    <Document>
      <Page style={styles.page} size={[1500, 600]} page>
        <View style={styles.image}>
          <Image src={imageBuffer} />
        </View>
....
```

examples/images/useTransformImage.js

```
const cache = {};

import Jimp from 'jimp';

export default function(url) {
  if (cache[url] === undefined) {
    throw new Promise(resolve => {
      Jimp.read(url).then(image => {
        image
          .greyscale()
          .getBufferAsync('image/jpeg')
          .then(buffer => {
            cache[url] = buffer;
            resolve();
          });
      });
    });
  }

  return cache[url];
}
```

This should also enable integration with future concurrent data fetching libraries (such as URQL), etc. 

I will explain the implementation details in the pull request.

rmmr:
Disregard the term **concurrent mode**, i meant **suspense**!

PR #906, opened 2020-05-15
Suspense support

rmmr:
Motivation in issue #905 . 

Rendering process works as follows:

1. The render function wraps the document in a React.Suspense component.
2. This suspense component creates a Fallback element of type "SUSPENDED"
3. While React renders "SUSPENDED" instead of "DOCUMENT" set a small timeout and loop.
4. Continue with layout phase.
5. (pull request #900) Perform another pass.
6. Compare layout against previous pass, if equal then render the pdf with the given layout.


rmmr:
@diegomura could you have a look at this? I'm currently running this with useQuery hooks (from URQL), etc. Enabling a much smoother development experience opposed to fetching data before rendering. I believe this will greatly improve this library!

PR #918, opened 2020-05-26
Split into packages using lerna

diegomura:


ozturkcangokkaya:
Hi Diego, thank you for the great repo! 

I was wondering if there is an ETA for v2 to come out?

diegomura:
@ozturkcangokkaya Unfortunately because of other duties I cannot say a precise ETA. This lerna refactor is almost done (running last tests these days and writing some docs) and then if I cannot find any big bug I'll merg eit to master. There is already a 2.0 version published in npm though if you would like to use it

PR #954, opened 2020-07-20
expose InternalBlobProvider

toddtarsi:
This is useful for custom download operations, like doing file writes in the cordova environment


toddtarsi:
Oh my gosh, I'm so dumb, I didn't realize BlobProvider was there >_<

PR #974, opened 2020-08-12
Added possibility to override hyphenationCallback

alevid99:
Added the possibility to override hyphenationCallback with a prop in Text component. This can be useful to avoid to split some text like peoples names


diegomura:
Oh sorry. Just merged this same change. Closing this, but thank you very much!

PR #992, opened 2020-08-31
TypeScript: Add ID property to ViewProps & TextProps interfaces

LucasMatuszewski:
Hi @diegomura - sorry that I didn't add it in the previous Pull Request, but I've realized that I need ID props also on View and Page components to Link to them internally.


diegomura:
Mm sorry. What are these `id` props for?

diegomura:
My bad. It's for link destinations. Thanks!

PR #1051, opened 2020-11-05
multi page wrap

kneczaj:
This is an example which should render on more than two pages, but it fails. See `THE END` is not at the pdf.


alew140:
it is a issue not is a commit

kneczaj:
Examples show various usage of the library, and they are also a good way to visually test if everything displays correctly.

If you look at this PR in this context then this should be merged.

diegomura:
This is not what PRs are for. Please submit an issue with this error

PR #1070, opened 2020-11-27
fix(v2): Fix adding custom fonts of different weights

gmaclennan:
This bug meant that it was not possible to add fonts with different fontWeight, all fonts were registered with fontWeight 400.


diegomura:
Thank you!

PR #1141, opened 2021-03-10
update typescript typings

tkreis:
It seems like the typings were out of date. In the code bases, I was only able to find references to:

overflow: 'hidden'
and textOverflow: 'ellipsis'
Not sure, if there are other options allowed as well.

**Sorry, about that I closed the PR, thinking I could just reopen it again.**


diegomura (approved):

diegomura:
Thanks!

PR #1142, opened 2021-03-10
Allow hyphenationCallback on Text component to be configured by prop

tkreis:
Typescript defs already defined a way to define the prop, but it did not have any affect. 

With this change, it would be easily possible to use different hyphenations for different parts of the PDF.


diegomura (approved):

diegomura:
Thanks! This is great

Issue #1148, opened 2021-03-15
Support for core-js3 & React 17!


ivaylopetrovdev:
To Reproduce
Steps to reproduce the behavior including code snippet (if applies):
npm install @react-pdf/renderer
npm install core-js

Expected behavior

NOT to receive such error:
ERROR in ./node_modules/babel-runtime/core-js/promise.js
Module not found: Error: Can't resolve 'core-js/library/fn/promise' in 'project_path/node_modules/babel-runtime/core-js'
@ ./node_modules/babel-runtime/core-js/promise.js 1:30-67
@ ./node_modules/babel-runtime/helpers/asyncToGenerator.js
@ ./node_modules/page-wrapping/dist/page-wrapping.es.js
@ ./node_modules/@react-pdf/renderer/dist/react-pdf.browser.es.js
@ ./src/app/util/pdf-generator.worker.js

ivaylopetrovdev:
@diegomura please address this with highest priority :)

princefishthrower:
@ivaylopetrovdev Made a pull request to bump react version here: https://github.com/diegomura/react-pdf/pull/1157
You can meanwhile just fork this and issue `npm update react`, it works fine and then all tests pass, not sure where any maintainers are on this one, the original React 17 version issue was raised last October...

ivaylopetrovdev:
> @ivaylopetrovdev Made a pull request to bump react version here: #1157
> You can meanwhile just fork this and issue `npm update react`, it works fine and then all tests pass, not sure where any maintainers are on this one, the original React 17 version issue was raised last October...

@princefishthrower lovely, can't wait to see your PR merged so we can start using this :) Appreciate your efforts!

KolbySisk:
Any updates on this? I'd like to use this with a new CRA project, but I'm unable to until React 17 is supported.

diegomura:
Keeping #1062. It's confusing to keep the conversation in both sorry

Issue #1123, opened 2021-02-19
This Library is Incompatible with React 17


osimuka:
**To Reproduce**
Steps to reproduce the behavior including code snippet (if applies):
```npm install @react-pdf/renderer```

**Expected behavior**
```
Found: react@17.0.1
node_modules/react
react@"^17.0.1" from the root project
Could not resolve dependency:
peer react@"^16.0.0" from @react-pdf/renderer@1.6.13
```

vineethsagar:
@diegomura Hey, can you please make the library compatible with React 17

princefishthrower:
Made a pull request to bump react version here: https://github.com/diegomura/react-pdf/pull/1157
You can meanwhile just fork this and issue npm update react, it works fine and then all tests pass, not sure where any maintainers are on this one, the original React 17 version issue was raised last October...

diegomura:
Keeping just #1062 to track this

Issue #1062, opened 2020-11-18
Support React 17


mquandalle:
NPM v7 throws the following error when using react-pdf with the latest version of React 

```
npm ERR! node_modules/react
npm ERR!   react@"^17.0.0" from the root project
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peer react@"^16.0.0" from @react-pdf/renderer@1.6.12
npm ERR! node_modules/@react-pdf/renderer
npm ERR!   @react-pdf/renderer@"^1.6.10" from the root project
```

I'm not sure if there are any code migration required or if only the `package.json` needs to me modified.

arobert93:
Any update on this?

xavierraffin:
Interested too

arobert93:
> it might help you `npm install @react-pdf/renderer --legacy-peer-deps`

This is a quick fix that solves the situation temporary, but if you remove the lock file, you have to use it again. It is not a clean solution...

princefishthrower:
Made a pull request to bump react version here: https://github.com/diegomura/react-pdf/pull/1157
You can meanwhile just fork this and issue `npm update react`, it works fine and then all tests pass, not sure where any maintainers are on this one, the original React 17 version issue was raised last October...

diegomura:
Hi! Sorry for the delay!
What would happen to projects using React 16?

arobert93:
@diegomura Maybe you can add both, with version range, until you test it.

diegomura:
So I tried upgrading react's peer dependency from `^16.8.6` to `>16.8.6` that matches new React 17 version, but NPM v7 stills throws peer dependency errors due to `react-reconciler@0.23.0` which has `react@16` as peer dependency as well... 

The newest `react-reconciler@0.26.2` has `react@^17.0.0` as peer dependency, but first it breaks react-pdf (still have to see why) and second, if I'm not mistaken will break react-pdf to anyone using `react@16`, which is still a lot of people.

It seems that the cause of this is NPMv7 now [installs peer dependencies](https://github.com/npm/rfcs/blob/latest/implemented/0025-install-peer-deps.md) and breaks things (this is apparently breaking the react library ecosystem, not just this package). I'm open to suggestions here but I do not see I can keep supporting all react versions and making NPMv7 to work, unless you use the `--legacy-peer-dep` flag as recommended above

diegomura:
[...]

PR #1157, opened 2021-03-27
bump peer dependency react to 17.0.0, all tests passed, no issues!

princefishthrower:
As many open issues state (https://github.com/diegomura/react-pdf/issues/1148) and (https://github.com/diegomura/react-pdf/issues/1123) for starters, this package currently will throw errors when installed in a project using react >= 17.0.0. I've bumped this version and all tests seemed to pass fine. hoping this can be merged ASAP as, for example, new framework versions like Gatsby V3 cannot use this awesome library! cheers!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,20881 @@
+{
+  "name": "@react-pdf/renderer",
+  "version": "1.6.15",
+  "lockfileVersion": 2,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "@react-pdf/renderer",
+      "version": "1.6.15",
+      "license": "MIT",
+      "dependencies": {
+        "@babel/runtime": "^7.6.2",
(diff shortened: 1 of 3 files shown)


daihuaye:
The issue to bump the peer dependency to React 17.0.0, all the project with React ^16.0.0 will need to upgrade too. I think it's shouldn't force them to upgrade to use this library. 

Could you elaborate: what's the root cause to upgrade to React 17 that is solved by the pull request?

diegomura:
Maybe  setting it to `^16.0.0 || ^17.0.0` or `>16.0.0` would be the safer thing to do here? They both resolve to the same versions

Screen Shot 2021-03-29 at 4 19 56 PM

diegomura:
This is not enough (please read https://github.com/diegomura/react-pdf/issues/1062) and also adds the package-lock.json which is not desired

PR #1183, opened 2021-04-08
Duplicate Id's removed from typings

builat:
There are duplicate keys `id` in `ViewProps` & `TextProps`. This PR resolves it.


diegomura:
Thanks @builat !

PR #1192, opened 2021-04-09
feat(add-href-support): add `href` support for a <Link>

BimBoss:
href is already supported but we can't use it due to missed property in the interface


diegomura:
Mmm this is not correct, since implies that both `src` and `href` should be present, right?

PR #1196, opened 2021-04-09
Fix renderToString blank document

### LINKED ISSUE #1067: ReactPdf.renderToString generates an empty white pdf
**Describe the bug**

It seems like there is something wrong with encoding while calling `renderToString` because `renderToFile` and `renderToStream` seems to works ok with the same code.

**File comparison:**
react-pdf-next

The stream part seems to be different.
Maybe there is a just a problem in my `next.js` code.
Although I tried every possible solution I could find.
```
IndexPage.getInitialProps = async ({res}) => {
  const filename = "file.pdf";
  await ReactPdf.renderToFile(<PdfDocument />, filename);
  const string = await ReactPdf.renderToString(<PdfDocument />);
  res.setHeader("Content-disposition", `attachment; filename="${filename}"`);
  res.setHeader("Content-Type", "application/pdf");
  res.end(string);
};
``` 
Here is a repo https://github.com/tomaswitek/react-pdf-next to reproduce it.

**To Reproduce**
1. `git clone https://github.com/tomaswitek/react-pdf-next`
2. `yarn install && yarn dev`
3. Open [http://localhost:3000](http://localhost:3000)
4. An empty PDF will be downloaded automatically

**Expected behavior**
A valid PDF should be downloaded

**Screenshots**

 - OS: [MacOS]
 - Browser [chrome]
 - React-pdf version [2]

diegomura:
Fixes #1067

For some reason if we disable `compress` on the pdfkit instance, the document renders correctly. Pushing this at least as an intermediate fix, but couldn't get to the bottom of why this was happening. Not sure I know enough about PDF internals to figure it out

### DIFF
--- a/packages/renderer/src/index.js
+++ b/packages/renderer/src/index.js
@@ -32,9 +32,13 @@ const pdf = initialValue => {
 
   if (initialValue) updateContainer(initialValue);
 
-  const render = async () => {
+  const render = async (compress = true) => {
     const ctx = new PDFDocument({ autoFirstPage: false });
     const layout = await layoutDocument(container.document, fontStore);
+
+    // For some reason, when rendering to string if compress=true the document is blank
+    ctx.compress = compress;
+
     return renderPDF(ctx, layout);
   };
 
@@ -70,7 +74,7 @@ const pdf = initialValue => {
 
   const toString = async () => {
     let result = '';
-    const instance = await render();
+    const instance = await render(false);
 
     return new Promise((resolve, reject) => {
       try {


hhprogram:
@diegomura This might be related to an issue I'm running into. Looking at this issue #1067 . I'm basically trying to collect the stream that results from calling `renderToStream` to a string so I can pass it between worker threads, and the resulting string comes out as gibberish / it renders a blank PDF.

If I'm understanding this correctly, this change aims to fix that?

diegomura:
Hard to say but It might. I'll publish a new version today so you can test it

PR #1203, opened 2021-04-11
Fix Page style array support

diegomura:


gabrielvincent:
This 100% resolved all the issues I was having on V2. Thank you so much, Diego!

PR #1248, opened 2021-04-22
docs: bring back access to Style type

meeroslav:
After migration to version 2, the access to the `Style` type is missing.

If we want to reference it we have to now use a hack, such as `typeof myStyles.myStyle` which is not ideal.

This PR exposes `Style` as `ReactPDF.Style`.

### DIFF
--- a/packages/renderer/index.d.ts
+++ b/packages/renderer/index.d.ts
@@ -15,6 +15,8 @@ declare module '@react-pdf/renderer' {
     interface Styles {
       [key: string]: Style;
     }
+    
+    type Style = Style;
 
     interface OnRenderProps {
       blob?: Blob;


diegomura:
Mmm wait... what is `ReactPDF.Style`? Do you mean `StyleSheet`?

meeroslav:
In version 1, the `Style` type was available. The simple use case for this would be e.g.

```typescript
const MyStyles: ReactPDF.Styles = StyleSheet.create({
  myStyle: {
    //... some styles here
  },
  someOtherStyle: {
    //... some styles here
  }
});

const paintRed = (style: Style) => Style {
   return { ...style, backgroundColor: '#ff0000'
}

const myRedStyle = paintRed(MyStyles.myStyle);
```

I have component methods that modify styles based on dynamic data - widths, colors, etc. In order to avoid my IDE and compiler screaming at me, I need to use the proper types. The type for Style already exists, it's just not exported.

diegomura:
Thanks for this!

diegomura:
Hey apparently this introduced a bug https://github.com/diegomura/react-pdf/issues/1269 .I'm not as familiarized with TS so I don't know exactly how to fix it. I recommend installing `@react-pdf/types` to use the `Styles` type so we can solve this and the other ticket as well. Thanks!

Issue #777, opened 2019-12-05
Unhandled Rejection (TypeError): Cannot read property 'hasGlyphForCodePoint' of null


stevensacks:
**Describe the bug**
I'm using PDFDownloadLink and rendering a perfectly valid PDF document. It went from sometimes throwing this error to always throwing this error. It used to just appear in the console, but with the latest release of react-scripts (3.3.0), now when it occurs, my entire app crashes. I tried wrapping it in an ErrorBoundary but it won't catch it.

I render the document on its own prior to passing it to the PDFDownloadLink. The rendering of the document has no problems. But, if I render PDFDownloadLink with the document, my app crashes.

I don't know what it causing this, but you need to try catch this error internally so it won't just crash the entire app. Or throw it in a way that an ErrorBoundary will catch it.

**To Reproduce**
I've figured out how to reproduce it. If you have two PDFDownloadLink's rendering the same document, or you quickly re-render the same PDFDownloadLink, it happens every time.

**Expected behavior**
For this error not to happen anymore, or at least not crash the app and be catchable by ErrorBoundaries.

**Screenshots**
<img width="1029" alt="Part 1" src="https://user-images.githubusercontent.com/162252/70216433-d8e07f00-1782-11ea-9894-93d672f3b3d6.png">
<img width="1027" alt="Part 2" src="https://user-images.githubusercontent.com/162252/70216436-db42d900-1782-11ea-9dfa-ed8a6e09c57b.png">

**Desktop (please complete the following information):**
 - MacOS Catalina 10.15.1
 - Chrome, Firefox
 - React-pdf version 1.6.8

SamJarmanPP:
+ 1, this is happening intermittently for me too. I register custom fonts as early as I can in the application lifecycle.

Lionprincipe:
Hello,@diegomura  
 i had this issue yesterday when i was refactoring a  previously working code. My issue was definitely like the one reported here https://github.com/diegomura/react-pdf/issues/636 . On chrome debugging tool the variable sent to the function  buildSubsetForFont  was null. I could not reproduce this issue on codesandbox.
. While searching in the react-pdf  code I have noticed that the Font class has a name issue on its register method. this.fontFamily  should be "this.family". I am not sure if this is the reason for  this error. I could not look deeper and now my code is working again and the Font is included on my PDF fill. I will look again later on this.

class Font {
  static create(family) {
    return new Font(family);
  }

  constructor(family) {
    this.family = family;
[...]

PR #1270, opened 2021-04-26
Don’t call hasGlyphForCodePoint if font is null

OliverLeighC:
Related to #777 

This doesn't fix the root cause of why fragments intermittently have no font property, but it does prevent the app from crashing if there is no font.


diegomura:
Thanks!

PR #1305, opened 2021-05-07
Change renderer queue require to import statement

mlmmn:
This change will allow Snowpack (or any rollup-related build system) work without the need to workaround mixed esm/commonjs dependencies as described https://github.com/snowpackjs/snowpack/discussions/1321


mlmmn:
@diegomura thanks for quick response! Answering your question: I have runned the test suites - all good. Once I built the rendered I copied it to my project and everything seems fine 👍   Both `es` and `cjs` builds have proper imports as per each type as well.

PR #1307, opened 2021-05-07
fix: FontDescriptor TS type

khmm12:
It should be `fontFamily` instead of `family`.

https://github.com/diegomura/react-pdf/blob/721113e72e73b308caa7dd82b0776418e76c491b/packages/font/src/index.js#L37
https://github.com/diegomura/react-pdf/blob/721113e72e73b308caa7dd82b0776418e76c491b/packages/font/src/index.js#L52


diegomura:
Thanks!

PR #1311, opened 2021-05-11
Update 'display' to accept 'table' - value

JuhoKon:
Previously display could also be "table", I wonder if this change has been done on purpose?

### DIFF
--- a/packages/types/style.d.ts
+++ b/packages/types/style.d.ts
@@ -17,7 +17,7 @@ export interface Style {
   // Layout
 
   bottom?: number | string;
-  display?: 'flex' | 'none';
+  display?: 'flex' | 'none' | 'table';
   left?: number | string;
   position?: 'absolute' | 'relative';
   right?: number | string;


jbovenschen:
Thanks for the PR

I don't know why it was an option and when it was an option but from what I know from react-pdf is the following:

React-pdf uses [yoga](https://yogalayout.com/) for rendering the UI.

Yoga layout is a subset of CSS it only understands flexbox based layouts. So it wouldn't understand how the `display: 'table'` should be rendered. 

Do you maybe have the release where it would've worked?

diegomura:
`display: table` was recently deleted from types becuause it wasn't valid. If you had it in your code it's probably not doing anything? If it is valid I'm happy to add iy back

JuhoKon:
I checked with the old version (1.6.12) which still had the `display: table`, and turned out if wasn't doing anything :smile:. I'm confused why it was added as an option in the first place :thinking:.

But thank you for the replies, this resolves the issue for me.

tgermain-da:
Sounds good, I will use a different alternative. I was just referencing an example using the table style and was trying to get it working locally. I am happy to keep using flex (preferred anyway :))

diegomura:
No sure I understood 😅  Setting `display: table` was the same as not defining anything. Where is the example you mentioned?

tgermain-da:
https://react-pdf.org/repl?code=140630f60760ce02e004015021808c03605338178eb027a6009807401a8096180ee001911540039a4be0171c32a980dc7350a446000b764802b8c30fc518004e4433c80ca310864e50c1a21b2152f901d4868ce0119f62e500942807311304f0b17000315c30084c0c6900b62e661eb4bcc0a090b088363878eac4e45474c0006698001e002214f218203014909cf260d4fc0148f2f61410e252606111e0d0f008597104899434b4c082ae9c004c00ac00a45eca6a1a5a3a7a7072d6c6a66e960b06ca003218a9ce2b9c9eeb4b08604cc16e9e00e4c04d91ad88089d0918a4081819307d95d5b59c92d2152a8d420272627046fc54a40602a0a000bcb0e64f0ddc2cd28bc06c1826029e0b800050012870003e381e380703800078b260100480218080c18914ca752000a487b06159ecf655390e85e5c0d9fcca60a6ca4b178a25ed696cb65828431239a5220480a54803d0205532a56eab20ac55cb8da2d37f395c4843e098181d5ea4da6c779a0de2c15bb2d02a747394452223a5516cb6bb9d8acf7865dbef9050400edd707dd56a354605baa948695084cf267d5eef7529d98a401400b41cac800c4836994fcab31186de6ad4e80331c000aa50653531375f4f37bd91c6f46554377398009c65f710ccb13b8196e01389ecfe7d3dae8e3da996d160bc3a748d0035045bbdd87b7128cda71d3c5e4ca755c9e5f2fbb4fa63399ac826a3802582a56ab120f20810c3c4a94c5b1791e06d5896fd802000

diegomura:
Not sure who wrote that but `display: table` was never supported

PR #1312, opened 2021-05-11
Update "More APIs" section

ttcode10:


diegomura:
Can you explain the motivation behind this change?

PR #1329, opened 2021-05-24
feat: error handling when source factories of Image return undefined

byeokim:
When resolving Image's source prop, if `src`/`source` is `undefined`, `fetchImage` will show relevant error message and return:

https://github.com/diegomura/react-pdf/blob/66d6259971e99846602d69de85fcbba98825d608/packages/layout/src/image/fetchImage.js#L18-L20

However, if `src`/`source` is not `undefined` but function returning `undefined`, `fetchImage` will pass through the above `if` block, invoke `resolveImage` and encounter `Cannot read property 'data' of undefined` error when evaluating `src.data`:

https://github.com/diegomura/react-pdf/blob/66d6259971e99846602d69de85fcbba98825d608/packages/image/src/resolve.js#L159-L160

This pull request adds code that will show relevant error message and return before invoking `resolveImage`, when `src`/`source` is function returning `undefined`.

### DIFF
--- a/packages/layout/src/image/fetchImage.js
+++ b/packages/layout/src/image/fetchImage.js
@@ -22,6 +22,11 @@ const fetchImage = async node => {
 
   try {
     const source = await resolveSource(src);
+
+    if (!source) {
+      throw new Error(`Image's "src" or "source" prop returned ${source}`);
+    }
+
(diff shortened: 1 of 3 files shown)


diegomura:
Thank you! This is great

PR #1349, opened 2021-06-02
rework transform

### LINKED ISSUE #1268: Transform:"scale()" not working
**Describe the bug**
 Transform on both axis its not working 
 (transform on just one axis works)
 0.9<= values make the element disappear

**To Reproduce**
 { transform:'scale(nr)' }

https://react-pdf.org/repl?code=3187b0760ce02e00408a057025803c450298c0bc300500943807cf805030c00f0022230080b6198531165d400a02180e659a004f003619b006f61622003a0046200099080beed2870e9da804946fc3369d31a78a95144639480c0d5c6744004ec1b002200f4b70c44f011d9000add030009964001cc0f9dd1c613c3538298ca93d780492d2e8199959d80801b8c8c800c5c0a1659c30f891a0319cf0243800cc791890448400b86001c801e42001dc784494fa0068385d817afa002ca0a12221bb3d3c5a2ae4f9a078a091816541193cfc4046c6953c00dc011801993c013401f4005401a400d400d840001c0075100dde400552503d18af000680138780041246c9962d3e991544512a8120b03304070300032a58c42485860309560355f61826871142a5eb3474911e128944818bbc40915e83c00acd30e57279310010ba0a020462f57ea2e3273b9bcbe0002440ce24000bc2a632172b28aa314707c02367180cce5a981be0d03b010d303bb1b3836bb56a75fad60baee000600d8b381b181419c3c4816d9c0afe8419d623c1840853631862351880c6157d04d8c30b0e4ea643944cf47b5b9fcd899e7807aa64a2690d627164001286078c02817068a52a8b0940d3c15110a8508258838a00

**Expected behavior**
Scale on both axis simultaneously

**Desktop**
 - OS: Linux
 - [Firefox , Chrome ]
 - React-pdf version latest(2.0.8)

jeetiss:
refactoring transform parsing

- it not regexp based, and it should be faster)
- easier add `skew` and refactor all operations to matrix
- tests
- fix #1268

I have questions about errors and validation and unit parsing


packages/render/src/operations/transform.js
```diff
@@ -2,113 +2,133 @@ import * as R from 'ramda';
...
+      break;
+    }
+
+    case 'matrix': {
+      ctx.transform(...args);
+      break;
+    }
+
+    default: {
+      console.error(`Transform operation ${name} doesn't supported`);
```
jeetiss, line 55: does `@react-pdf` have better way for logging errors?

packages/render/src/operations/transform.js
```diff
@@ -2,113 +2,133 @@ import * as R from 'ramda';
...
+
+    case 'scaleX': {
+      return ['scale', [Number.parseFloat(args), 1]];
+    }
+    case 'scaleY': {
+      return ['scale', [1, Number.parseFloat(args)]];
+    }
+
+    case 'rotate': {
+      const unitsRegexp = /(-?\d*\.?\d*)(\w*)?/i;
```
jeetiss, line 79: what about parsing and converting units? this should work as before, but it can be also `turn` and `grad`

does `@react-pdf` have util for parsing unit that i can use here?

packages/render/src/operations/transform.js
diegomura: Unfortunately not for the moment 😞   What logic can be embedded in a error logging util that a `console.error` does not cover? It might be a good idea to create a new package (`@react-pdf/log` maybe?) in which we can log any sort of events (info, error, warning) and either use that to show in the console or generate reports

packages/render/src/operations/transform.js
diegomura: No, since all unit transformations happen in the `@react/stylesheet` package. That makes me think if it shouldn't be better to move that logic to there. See comment below

diegomura:
Hey @jeetiss ! 

First of all, thank you so much for your awesome contributions. It's always very exciting to see when talented people jumps into this project to fix or add new features. If contributing to react-pdf is something you would like to keep doing, just let me know and I'll try to facilitate everything as much as I can for you 😄 

And thanks for tackling this! The only comment I have is the following: as react-pdf work right now it's divided into two main  parts: [layout](https://github.com/diegomura/react-pdf/tree/master/packages/layout) and [rendering](https://github.com/diegomura/react-pdf/tree/master/packages/render). It's very important to keep these independent from each other since it enables us to just focus on a smaller issue on each step. Layout does not anything about PDF, just cares about computing where things should be in the final doc. On the other side, rendering does not care about how the item placement was computed, it's responsibility is only how to draw that data in the final target document. This also has an advantage I didn't exploded yet, which is rendering the data in other format than PDF. 

The layout process is also layered into different substeps, being one of the first ones style transformation [here](https://github.com/diegomura/react-pdf/blob/master/packages/layout/src/index.js#L34), but most of the magic happening on [@react-pdf/stylesheet](https://github.com/diegomura/react-pdf/tree/master/packages/stylesheet). This package transforms each style into it's most normal form, so successive steps do not have to care if the style was entered as `margin` or `marginTop`, or in `pt` or `cm`. That's why I think it may make sense to apply this `transform` style parsing and normalization there. What do you think? Then `@react-pdf/render` might just expect the final format to apply those operations in the final doc.

Let me know if that makes sense. Happy to explain this further. Take care!

packages/render/src/operations/transform.js
jeetiss: No, I mean wrap logging with process.env.NODE_ENV like react does, because this seems useless in prod code, but i'm not sure about that

jeetiss:
@diegomura thanks for explanation, woow now it's much clearer

I agree that `@react-pdf/stylesheet` is better place for parsing `transform` and I'm gonna move that logic

jeetiss:
and one more thing

I feel that unit tests doesn't cover all cases and I'm not sure that everything work as expected in final form and I didn't find anything for check that. I experimented with e2e tests for pdfs [jeetiss/react-pdf-testing-library](https://github.com/jeetiss/react-pdf-testing-library)

it converts pdfs to png images and compare them with `jest-image-snapshot`, maybe we should setup some thing similar for testing `@react-pdf/renderer`. it should be super helpful for testing stuff like `transform` and it works fast. what do you thing?

packages/render/src/operations/transform.js
diegomura: Oh got it. No there's nothing like that for the moment 😅

diegomura:
> it converts pdfs to png images and compare them with jest-image-snapshot, maybe we should setup some thing similar for testing @react-pdf/renderer. it should be super helpful for testing stuff like transform and it works fast. what do you thing?

I didn't know that package! It's really cool, thanks for working on it. I worked once with visual regression tests and didn't have the best experience tbh. But it was a long time ago and they were very slow. I take your word these are fast. I see those as complementary tests from the unit tests we have, basically testing the whole solution and not just bits of it such as we do now. If you like, feel free to submit a Pr with _some_ tests at least so I can see how it would work. I'm open to the idea. Maybe the `renderer` package itself would be a good place to have these e2e tests

diegomura:
I tested it and it's working very good! Thanks again for working on this and keep the PR's coming 😬

PR #1358, opened 2021-06-03
Fix examples in README

JuliusSr:
Fixing examples links not working and images not showing in README files:
- Fixed broken paths (also, removed Images example as it doesn't appear to be around anymore).
- Added anchor element pointing to example folder for each image, instead of letting GitHub adding one automatically that points to the thumbnail file.
- Also, removed rogue spaces indentation in favour of tabs (which seems to be the preferred style) for a consistant indentation.

### DIFF
--- a/packages/examples/README.md
+++ b/packages/examples/README.md
@@ -10,17 +10,23 @@
 	<tbody>
 		<tr>
 			<td align="center" valign="top">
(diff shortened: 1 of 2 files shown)


JuliusSr:
It ain't much, but it's honest work :)

diegomura:
Thanks!!

Issue #809, opened 2020-01-07
Multiple text decoration values produce no text decorations


SamJarmanPP:
**Describe the bug**
In CSS, its valid to use `text-decoration: underline line-through` however, when using the equivalent in ReactPDF, neither are rendered.

**To Reproduce**
```
const Test = () => (
  <Document>
    <Page style={styles.body}>
      <Text style={{textDecoration: 'underline line-through'}}>
        Created with react-pdf
      </Text>
   
    </Page>
  </Document>
);
```
[...]

PR #1404, opened 2021-07-14
Add support for multiple text decorations

aaronjedwards:
## Summary

Adds functionality for specifying multiple values for the `textDecoration` style attribute (Related to #809). 

This brings the `textDecoration` attribute in this library closer in line to  the standard CSS `text-decoration` attribute so it seems like a logical addition (and something that I was looking for in this library).

## Details

In particular, this change allows for specifying either `'underline line-through'` or `'line-through underline'` while disallowing invalid combinations such as `'underline none'`.  

I've updated the Typescript types as well to include `'underline line-through'` or `'line-through underline'` as valid values for the `textDecoration` style attribute.

I am unsure on how to approach implementing a test for this functionality or if one is needed, so any input on that front is welcome.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -4113,6 +4113,11 @@ estraverse@~1.5.0:
   resolved "https://registry.yarnpkg.com/estraverse/-/estraverse-1.5.1.tgz#867a3e8e58a9f84618afb6c2ddbcd916b7cbaf71"
   integrity sha1-hno+jlip+EYYr7bC3bzZFrfLr3E=
 
+estree-walker@^0.2.1:
+  version "0.2.1"
+  resolved "https://registry.yarnpkg.com/estree-walker/-/estree-walker-0.2.1.tgz#bdafe8095383d8414d5dc2ecf4c9173b6db9412e"
+  integrity sha1-va/oCVOD2EFNXcLs9MkXO225QS4=
+
 estree-walker@^0.6.1:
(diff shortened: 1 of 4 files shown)


yarn.lock
```diff
@@ -6574,7 +6569,7 @@ min-indent@^1.0.0:
   resolved "https://registry.yarnpkg.com/min-indent/-/min-indent-1.0.1.tgz#a63f681673b30571fbe8bc25686ae746eefa9869"
   integrity sha512-I9jwMn07Sy/IwOj3zVkVik2JTvgpaykDZEigL6Rx6N9LbMywwUSMtxET+7lVoDLLd3O3IXwJwvuuns8UB/HeAg==
 
-minimatch@^3.0.2, minimatch@^3.0.4:
+minimatch@^3.0.4:
```
diegomura, line 6572: Im curious why yarn.lock changed here 😅 
Besides that looks great

aaronjedwards: Sorry! I backed out the changes to yarn.lock!

aaronjedwards:
@diegomura I've removed the lock file changes now so if everything else looks good this should be good to go!

yarn.lock
aaronjedwards: @diegomura thats weird. I backed out these changes and simply reinstalled the packages and this still comes through. How would you like to resolve this?

Issue #1090, opened 2021-01-08
How can I use PDFViewer without rendering toolbar?


azat-io:
Hi there!

I need property in PDFViewer component to render pdf without toolbar. You can make it just by adding `#toolbar=0` to blob PDF URL

![image](https://user-images.githubusercontent.com/5698350/104010392-78114a80-51bd-11eb-9872-7de1cbbce612.png)

diegomura:
No. The PDFViewer its just an iframe and uses the native browser PDF viewer (as you would see if you open any PDF using your default browser). There's nothing this lib can to to change how it looks. However, you can try building your own PDF viewer, similar to how the REPL in the site works. Here's the code for reference: https://github.com/react-pdf/site/blob/master/src/components/Repl.js

PR #1419, opened 2021-07-22
Add prop 'showToolbar' to control whether the toolbar shows on supported Browsers

DanMHammer:
### Summary

Adds a prop, `showToolbar` to the PDFViewer to apply the `toolbar` parameter (`#toolbar=0` when the parameter is false and `#toolbar=1` when true) to the iframe in order to request to show/hide the toolbar in supported browsers. 

Additionally adds an example to the readme with details about browser compatibility. The parameter is implemented in Chrome, Chromium Edge, and Safari. Firefox requires local changes to userContent.css, which cannot be forced, ignored, or overwritten by javascript.

### Testing

Pass in showToolbar={false} to the PDFViewer.

```jsx
import React from 'react';
import ReactDOM from 'react-dom';
import { PDFViewer } from '@react-pdf/renderer';

const App = () => (
  <div>
        Without Toolbar:
        <PDFViewer height={1000} showToolbar={false} width="100%">
          <MyDoc />
        </PDFViewer>
      </div>
);

ReactDOM.render(<App />, document.getElementById('root'));
```

### Screenshot
image

See: https://github.com/diegomura/react-pdf/issues/1090


diegomura:
Thanks for working on this! I removed the README docs and moved them to the site. Hope you don't mind. The README should keep as a first-glance sort of documentation, and this is quite specific.

DanMHammer:
@diegomura Great! Makes sense about the readme. My team really likes ReactPDF, and this was one feature we really wanted to have because we have our own workflow for printing and downloading. Thanks for merging in!

Issue #1388, opened 2021-06-28
Ability to print PDFs?


codebu5ter:
I'm looking for a way to print PDF files on the fly, without rendering the components.
For download, the **PDFDownloadLink** component works well, but for printing, I'm not sure how to make it work.

Here's how my component is coded for now -

```
const MyDoc = () => (
  <Document>
    <Page>
      // My document data
    </Page>
  </Document>
);

const App = () => (
  <div>
    <PDFDownloadLink document={<MyDoc />} fileName="somename.pdf">
      {({ blob, url, loading, error }) =>
        loading ? 'Loading document...' : 'Download now!'
      }
    </PDFDownloadLink>
  </div>
);
```

Is there a way to print PDF files on the fly too?

contactsimonwilson:
import print from 'print-js';
...          
const blob = await pdf(<MyDoc />).toBlob();
print(URL.createObjectURL(blob));
...

uwemneku:
> ```
> import print from 'print-js';
> ...          
> const blob = await pdf(<MyDoc />).toBlob();
> print(URL.createObjectURL(blob));
> ...
> ```

This works. Thanks @contactsimonwilson

diegomura:
Thanks for answering! Closing it as resolved

PR #1420, opened 2021-07-22
Return a print function from usePDF

DanMHammer:
Adds a print function to usePDF to add a hidden iframe to document, assign the pdf url to the iframe, print it, and then remove it. 

Updates Readme with a usage example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,26 @@ const App = () => (
 ReactDOM.render(<App />, document.getElementById('root'));
 ```
 
+### `Web.` usePDF
+```jsx
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { usePdf, PDFViewer} from '@react-pdf/renderer';
+
+import MyDocument from ...;
+
+const App = () => (
+  const [instance, update, print] = usePDF({document: <MyDocument />})
+
+  return (
+      <button type="button" onClick={() => print()}>Print</button>
+  )
+);
+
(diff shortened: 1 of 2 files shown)


diegomura:
Thanks for this contribution! I have doubts if this should be added to this hook or can be provided as a separate fn. I'll give this some second thoughts but let me know what you think or if there's another way :)

diegomura:
I just found out this: https://github.com/diegomura/react-pdf/issues/1388#issuecomment-872732654

IMO it's a great solution. I'm not a huge fan of bundling all the features inside react-pdf while they can be achieved by third-party libs. This way we can keep this module as small as possible (it's not small already) and at the same time we reduce the amount of code to maintain in this end which it's already a lot. Let me know what you think

DanMHammer:
print-js is not compatible with NextJS (and possibly any other static generation, but I haven't confirmed), and the maintainer has not responded to an [open PR from Mar 1](https://github.com/crabbly/Print.js/pull/543) to merge in a fix. 

However, it's pretty trivial for users of ReactPDF to do exactly what I did in this function and just pass in the instance.url if you don't feel like it fits into the library. It might be worth putting this example and the print-js version into documentation somewhere though.

daphnesmit:
Anybody looking here for a fix. I ported library (everything except the Modal functionality, because thats UI IMHO):
(I am planning to rewrite to Typescript)
https://www.npmjs.com/package/prntr

PR #1465, opened 2021-08-18
fix: PDFViewer src while loading

### LINKED ISSUE #1463: PDFViewer is mounted with null?toolbar=1 url
Since the following change, the PDFViewer iframe is mounted with `url=null?toolbar=1` instead of no url at all
https://github.com/diegomura/react-pdf/commit/21c66cdaaa708aad3b3bdaf9aa852f341a56cfc0#diff-27423d243324698da9532db7315618c00e0023ee1257694d9367c02ff1ff4aa5

Easily fixable IMO:
```javascript
export const PDFViewer = ({
  title,
  style,
  className,
  children,
  innerRef,
  showToolbar = true,
  ...props
}) => {
  const [instance, updateInstance] = usePDF({ document: children });

  useEffect(updateInstance, [children]);

  return (
    <iframe
      title={title}
      ref={innerRef}
      style={style}
      src={instance.url ? `${instance.url}#toolbar=${showToolbar ? 1 : 0}` : null}
      className={className}
      {...props}
    />
  );
};
```

diegomura:
Fixes #1463

### DIFF
--- a/packages/renderer/src/dom.js
+++ b/packages/renderer/src/dom.js
@@ -97,12 +97,16 @@ export const PDFViewer = ({
 
   useEffect(updateInstance, [children]);
 
+  const src = instance.url
+    ? `${instance.url}#toolbar=${showToolbar ? 1 : 0}`
+    : null;
+
   return (
     <iframe
+      src={src}
       title={title}
       ref={innerRef}
       style={style}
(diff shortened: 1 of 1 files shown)


loun4:
Thanks !

PR #1468, opened 2021-08-18
Fix missing style type for zIndex support

Xiyitifu:
Hi, just a small fix to add zIndex property support to defined types.

Thanks for your work !


diegomura:
Thanks!

Issue #838, opened 2020-02-27
React pdf does not understand 'transparent' background color


brandonryan:
**Describe the bug**
Transparent background color displays as black

**To Reproduce**
Render the following
```jsx
<Text style={{backgroundColor: 'transparent'}}>
lorem ipsum
</Text>
```
(repl wont let me share link because its too long)

**Expected behavior**
Text element should be rendered with an rgba value of rbga(0,0,0,0)

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser chrome
 - React-pdf (renderer) 1.6.8

geerthoekzema:
I was able to reproduce your issue, strangly it dissapeared when the "lorem ipsum" was on the same line as the opening from <Text> or when i added a tabs before the Lorem Ipsum.

[...]

PR #1478, opened 2021-09-01
fix issue with background color opacity never being selected

mikey0000:
This addresses #838

Let me know if this is the correct logic and I'll add some tests against it.
[Edit] updating tests and my fix.

### DIFF
--- a/packages/render/src/primitives/renderBackground.js
+++ b/packages/render/src/primitives/renderBackground.js
@@ -9,7 +9,7 @@ const drawBackground = (ctx, node) => {
   if (node.box && node.style.backgroundColor) {
     const { top, left, width, height } = node.box;
     const color = parseColor(node.style.backgroundColor);
-    const opacity = R.defaultTo(color.opacity, node.style?.opacity);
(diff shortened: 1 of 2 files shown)


packages/render/src/primitives/renderBackground.js
```diff
@@ -9,7 +9,8 @@ const drawBackground = (ctx, node) => {
   if (node.box && node.style.backgroundColor) {
     const { top, left, width, height } = node.box;
     const color = parseColor(node.style.backgroundColor);
-    const opacity = R.defaultTo(color.opacity, node.style?.opacity);
+    const nodeOpacity = !R.isNil(node.style?.opacity) ? node.style.opacity : 1
```
diegomura, line 12: Nit. change :)

```suggestion
    const nodeOpacity = R.isNil(node.style?.opacity) ? 1 : node.style.opacity;
```

diegomura:
Thanks so much @mikey0000 ! This looks great

mikey0000:
Ah yeah I noticed the code style issue after pushing haha, thanks for changing it.

PR #1485, opened 2021-09-04
feat: add bidi support

diegomura:
Add RTL support


kassah-mahmoud:
Hi @diegomura, thanks for this great library
Could you please estimate when the RTL support will be done?

PR #1516, opened 2021-09-21
apply patch for links from v2 to v1 too

jeetiss:
Hi @diegomura 
Can you please release this patch for links in `react-pdf` v1?

now links doesn't works when they have other text elements inside, like this

```html
<Text>text: <Link src="https://link.com/"><Text style={bold}>link</Text></Link></Text>
<Text>text: <Link src="https://link.com/"><Text style={cursive}>link</Text></Link></Text>
```

this happen because `getFragments` function lose `src` prop when process `Link` children, so I fix that

### DIFF
--- a/src/utils/attributedString.js
+++ b/src/utils/attributedString.js
@@ -29,7 +29,7 @@ const transformText = (text, transformation) => {
   }
 };
 
-export const getFragments = instance => {
+export const getFragments = (instance, parentLink) => {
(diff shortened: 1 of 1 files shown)


omgovich:
🙏

diegomura:
Thanks @jeetiss . Merging. Will try to release new versions later today for both v1 and v2

PR #1535, opened 2021-10-06
feat(stylesheet): rework box model style parsing

jeetiss:
Here is the first step for a stricter style parser. The current one ignores all edge cases and `@react-pdf/render` fails with crazy unpredicted errors. I use `postcss-value-parser` to achieve that, this is an internal package from `post-css` that parse all possible CSS values to `AST`. it introduces other opportunities, like `background` style support or maybe `calc` support

I rewrite box model parsing (margin/padding) to a new approach in this PR. it now 
- auto works properly in all cases ( `2 auto` or `1 auto 2 3` )
- auto works only with margin
- checks units for margin and padding
- ignores all invalid values

@diegomura @izmaelmag what do you think about these changes? what does `@react-pdf` should do with invalid values in the best scenario? (ignore, log errors, throw errors)

### todos
- [x] write **unified** error message for all cases
- [x] check that PR doesn't have breaking changes

### DIFF
--- a/.changeset/empty-zoos-live.md
+++ b/.changeset/empty-zoos-live.md
@@ -1,0 +1,5 @@
+---
+'@react-pdf/stylesheet': minor
+---
+
+fixed `margin: auto` parsing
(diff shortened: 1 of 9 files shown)


diegomura:
Thanks @jeetiss ! I would like to test this in practice. Will try to do it asap, but looks great 😄

Issue #1496, opened 2021-09-10
Internals or other private api for tools, libraries, etc


jeetiss:
**react-pdf dev tools**

I made dev tools for `@react-pdf/renderer` for fun, it looks like [👇🏻 this 👇🏻](https://react-pdf-experiments.vercel.app/)

<details>
  <summary>screenshot</summary>
  <img width="1440" alt="Снимок экрана 2021-09-10 в 10 35 04" src="https://user-images.githubusercontent.com/6726016/132817588-1aae1f63-e851-4b3b-a47a-a16960cb8ece.png">
</details>

it shows parsed and computed styles, pages and `react-prd`s rendering primitives after all pagination magic. it more helpful than `debug` prop, because it interactive ✨. I wanna make it more library like, and I have problem — `@react-pdf/renderer` doesn't provide `layout` info in any way, so maybe you have any ideas how it should work.

**Describe the solution you'd like**
best solution here is dev-tool extension for chrome, but it requires alot of work and integration with [react-pdf](https://github.com/wojtekmaj/react-pdf) that usually used for rendering. I think that now will be enough some thing like reacts internal `__DO_NOT_USE_OR_YOU_WILL_BE_FIRED` or maybe something else.

**How it works now**
I patch `@react-pdf/renderer` with `patch-package` library, yeaas this is hack, but it works

<details>
  <summary>and here the change</summary>

```diff
diff --git a/node_modules/@react-pdf/renderer/lib/react-pdf.browser.es.js b/node_modules/@react-pdf/renderer/lib/react-pdf.browser.es.js
index 52d4b40..bc8a9fe 100644
--- a/node_modules/@react-pdf/renderer/lib/react-pdf.browser.es.js
+++ b/node_modules/@react-pdf/renderer/lib/react-pdf.browser.es.js
@@ -219,7 +219,9 @@ var pdf = function pdf(initialValue) {
 
             case 6:
               layout = _context.sent;
-              return _context.abrupt("return", renderPDF(ctx, layout));
+              var result = renderPDF(ctx, layout);
+              result.__INTERNAL_LAYOUT = layout
+              return _context.abrupt("return", result);
 
             case 8:
             case "end":
@@ -264,6 +266,7 @@ var pdf = function pdf(initialValue) {
                     callOnRender({
                       blob: blob
                     });
[...]

PR #1539, opened 2021-10-08
Exposing layout only

sagar-tts:
I have checked issue https://github.com/diegomura/react-pdf/issues/1496. I have also similar requirement to preview quickly and may be allow to edit ( in future ). so, I though if we only expose layout and provide preview and not spend time on generating actual  pdf but just display the layout on screen, then it will be quick turn around time for preview.

@diegomura please tell me if you have any other thoughts for this. I am happy to add feature for rendering preview or may be update REPL based on layout. 

Looking forward for your feedback.

### DIFF
--- a/packages/renderer/src/index.js
+++ b/packages/renderer/src/index.js
@@ -49,6 +49,11 @@ const pdf = initialValue => {
     return renderPDF(ctx, layout);
   };
 
+  const toLayout = async () => {
+    const layout = await layoutDocument(container.document, fontStore);
+    return layout;
+  };
+
   const callOnRender = (params = {}) => {
     if (container.document.props.onRender) {
       container.document.props.onRender(params);
@@ -113,6 +118,7 @@ const pdf = initialValue => {
   return {
     on,
     container,
+    toLayout,
     toBlob,
     toBuffer,
     toString,


jeetiss:
I think that this is not the best change right now, and that's why
- this method locks all breaking changes in the layout structure (it can change in near future, like [here](https://github.com/diegomura/react-pdf/pull/1535), or [here](https://github.com/diegomura/react-pdf/pull/1349))
- this method calculates layout again, and doesn't provide a way to reuse the layout between different renderers (pdf or svg)

Is will better to make a plugin system that is @diegomura talk about [here](https://github.com/diegomura/react-pdf/issues/1496#issuecomment-917315464)

If you wanna experiment only with layout you can recreate it from internal packages, like this

```js
import layoutDocument from '@react-pdf/layout';
import { Font as fontStore, createRenderer } from '@react-pdf/renderer';

let renderer = createRenderer({});

const generateLayout = (doc) => {
  const container = { type: 'ROOT', document: null };
  const mountNode = renderer.createContainer(container);

  renderer.updateContainer(doc, mountNode, null);
 
  return layoutDocument(container.document, fontStore);
}
```

and i hope that you understand that layout structure can changfe in near future

diegomura:
I'm 100% aligned with @jeetiss comments here. That's an internal structure and exposing it will force us not to be able to evolve it or change it in the future without breaking things. Dev bandwidth is already limited on my side so we should keep public api as neat as possible

PR #1567, opened 2021-11-01
Fix bestFit breakpoint issue

hayoung0Lee:
Hi, I have been experiencing a weird issue related to how `applyBestFit` handles position. 

## Issue
Pasted Graphic 6

- [link to bestFit.js](https://github.com/diegomura/react-pdf/blob/3bcb031858062ca12865de7513c5fb5eb6ab3e4f/packages/textkit/src/engines/linebreaker/bestFit.js)

As you see, the current `applyBestfit` function suddenly stops getting `nextBreakPoint`.
I think this issue occurred because of the two lines. 

1. `getNextBreakpoint` returning null, when it shouldn't
```
if (sum.width - sum.shrink > lineLength) break; // getNextBreakpoint. 
```
In the function, when `sum.width - sum.shrink` is bigger than `lineLength`, it breaks. However, because the default value of position is `null`, in some cases, it just returns `null`. 

It is problematic because it means that after any node that has a width bigger than lineLength, it stops there and doesn't calculate the next breakpoint. 
```javascript
    const breakpoint = getNextBreakpoint(subnodes, widths, lineNumber); // this just returns null 

    if (breakpoint) { 
      count += breakpoint;
      breakpoints.push({ position: count });
      subnodes = subnodes.slice(breakpoint + 1, subnodes.length);
      count++;
      lineNumber++;
    } else {
      // it ends up here even when it has more subnodes to calculate. 
      subnodes = [];
    }
```

So my solution is to give a position value based on `i` value. 

```
    if (sum.width - sum.shrink > lineLength) {
      if (i > 0) {
        position = i - 1;
      } else {
        position = i;
      }
      break;
    }
```

When it is the first block of the subnodes, I think the position value should be '0'. Otherwise, the position has to be `i-1`, because I think it makes more sense not to include the current subnode to the previous one to make it to the next line. 

2. if(breakpoint) condition
```javascript
if (breakpoint !== null) { // breakpoint -> breakpoint !== null
      count += breakpoint;
      breakpoints.push({ position: count });
      subnodes = subnodes.slice(breakpoint + 1, subnodes.length);
      count++;
      lineNumber++;
    } else {
      subnodes = [];
    }
```
I think this should be `breakpoint !== null` because this while loop shouldn't stop when the breakpoint is zero. There could be more subnodes to calculate. 

## Result
Pasted Graphic 7

Any feedback is welcome!

### DIFF
--- a/packages/textkit/src/engines/linebreaker/bestFit.js
+++ b/packages/textkit/src/engines/linebreaker/bestFit.js
@@ -72,8 +72,6 @@ const applyBestFit = (nodes, widths) => {
   let subnodes = nodes;
   const breakpoints = [{ position: 0 }];
 
-  // console.log('nodes', nodes);
-
   while (subnodes.length > 0) {
     const breakpoint = getNextBreakpoint(subnodes, widths, lineNumber);
 
--- a/packages/textkit/src/engines/linebreaker/bestFit.js
+++ b/packages/textkit/src/engines/linebreaker/bestFit.js
@@ -36,7 +36,7 @@ const getNextBreakpoint = (subnodes, widths, lineNumber) => {
     }
 
     if (sum.width - sum.shrink > lineLength) {
-      if (i > 0) {
+      if (i >= 1) {
         position = i - 1;
       } else {
         position = i;
--- a/packages/textkit/src/engines/linebreaker/bestFit.js
+++ b/packages/textkit/src/engines/linebreaker/bestFit.js
@@ -35,7 +35,14 @@ const getNextBreakpoint = (subnodes, widths, lineNumber) => {
       sum.shrink += node.shrink;
     }
 
-    if (sum.width - sum.shrink > lineLength) break;
+    if (sum.width - sum.shrink > lineLength) {
+      if (i > 0) {
+        position = i - 1;
+      } else {
+        position = i;
+      }
+      break;
+    }
 
     if (node.type === 'penalty' || node.type === 'glue') {
       const ratio = calculateRatio(node);
@@ -61,7 +68,7 @@ const applyBestFit = (nodes, widths) => {
   while (subnodes.length > 0) {
     const breakpoint = getNextBreakpoint(subnodes, widths, lineNumber);
 
-    if (breakpoint) {
+    if (breakpoint !== null) {
       count += breakpoint;
       breakpoints.push({ position: count });
       subnodes = subnodes.slice(breakpoint + 1, subnodes.length);
--- a/packages/textkit/src/engines/linebreaker/bestFit.js
+++ b/packages/textkit/src/engines/linebreaker/bestFit.js
@@ -36,22 +36,18 @@ const getNextBreakpoint = (subnodes, widths, lineNumber) => {
     }
 
     if (sum.width - sum.shrink > lineLength) {
-      let j;
-      if (i === 0) {
-        // the first block is always "box" type
-        j = i + 1;
-      } else {
-        j = i;
-      }
+      if (position === null) {
+        let j = i === 0 ? i + 1 : i;
 
-      while (
-        j < subnodes.length &&
-        (subnodes[j].type === 'glue' || subnodes[j].type === 'penalty')
-      ) {
-        j++;
-      }
+        while (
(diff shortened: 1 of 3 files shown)


jeetiss:
@hayoung0Lee thanks! your PR looks promising!

can you please add test for this change?

hayoung0Lee:
@jeetiss 
Thank you for the feedback! I'm going to work on adding test soon!

hayoung0Lee:
@jeetiss 
Hi! I added some test code but since I don't have much experience in writing test cases, I'd love to get some feedback if my test cases are not good enough. 

I made test cases with five sentences. 
* "1234567891011"
* "123 1234567891011"
* "123 1234567891011 123"
* "123 1234567891011 123 1234567891011"
* "123 1234567891011 1234567891011 1234567891011"

And the widths value is `const widths = [78, 78];`

I used numbers to make sentences with the purpose of making the width of each node bigger than the widths.

I am attaching the results!
* "1234567891011"
image

* "123 1234567891011"
image

* "123 1234567891011 123"
image


* "123 1234567891011 123 1234567891011"
image


* "123 1234567891011 1234567891011 1234567891011"
image

hayoung0Lee:
Well I found a bug in my code. I am going to fix it soon!

hayoung0Lee:
```javascript
    if (sum.width - sum.shrink > lineLength) {
      let j;
      if (i === 0) {
        // the first block is always "box" type
        j = i + 1;
      } else {
        j = i;
      }

      while (
        j < subnodes.length &&
        (subnodes[j].type === 'glue' || subnodes[j].type === 'penalty')
      ) {
        j++;
      }

      position = j - 1;
      break;
    }
```

I added while loop to take all the glue and penalty block before breaking line.

diegomura:
@jeetiss mind reviewing and merging? 🙏

packages/textkit/src/engines/linebreaker/bestFit.js
```diff
@@ -35,7 +35,25 @@ const getNextBreakpoint = (subnodes, widths, lineNumber) => {
...
+      }
+
+      while (
+        j < subnodes.length &&
+        (subnodes[j].type === 'glue' || subnodes[j].type === 'penalty')
+      ) {
+        j++;
+      }
+
+      position = j - 1;
```
jeetiss, line 54: please run all this code only when position is `null`, like

```js

if (position == null) {
  // code
}

```

packages/textkit/tests/engines/linebreaker/bestFit.test.js
```diff
@@ -0,0 +1,330 @@
+import applyBestFit from '../../../src/engines/linebreaker/bestFit';
+
+// Sentence: "1234567891011"
+const node1 = [
+  {
+    type: 'box',
+    width: 84.81599999999999,
+    value: {
```
jeetiss, line 8: please, remove all useless properties like `value`, `hyphenated`, `flagged` from test data

jeetiss (changes requested):
@hayoung0Lee so sorry this got postponed.

i'm here with small improvements. thanks again for contribution

packages/textkit/src/engines/linebreaker/bestFit.js
```diff
@@ -35,7 +35,25 @@ const getNextBreakpoint = (subnodes, widths, lineNumber) => {
...
 
-    if (sum.width - sum.shrink > lineLength) break;
+    if (sum.width - sum.shrink > lineLength) {
+      let j;
+      if (i === 0) {
+        // the first block is always "box" type
+        j = i + 1;
+      } else {
+        j = i;
+      }
```
jeetiss, line 45:
```suggestion
      // the first block is always "box" type
      let j = i === 0 ? i + 1 : i;
```

packages/textkit/src/engines/linebreaker/bestFit.js
hayoung0Lee: Thanks!

hayoung0Lee:
@jeetiss Thank you for your review! I applied it!

jeetiss (approved):
🕺🏻

PR #1584, opened 2021-11-06
fix: webpack 5 support

jeetiss:
- [x] stream package (`Writable`, `Readable`, `BlobStream`)
- [x] zlib
- [x] buffer ( ??? )
- [ ] check webpack5
- [ ] check vite


jeetiss:
I tried to make `react-pdf` compatible with `webpack 5` out of the box and failed,

because all browser polyfills are not built by `webpack 5` out of the box either and I need to fork/modify them too.

I think should be better to add instruction on how to use `react-pdf` with `webpack 5` and `vite` in readme instead of fixing all these polyfill stuffs at the library level.

Issue #1553, opened 2021-10-21
Border corner got cut off when applied border radius


handao-everfit:
**Describe the bug**
When I applied borderRadius to the borders of a box, the top border gets overlapped with the right border
![image](https://user-images.githubusercontent.com/87088870/138191568-821c387f-7143-4a82-9461-8820bbb725b6.png)

**To Reproduce**
[Reproduced Code](https://react-pdf.org/repl?code=3187b0760ce02e00408a057025803c450298c0bc300500943807cf805030c00f0022230080b6198531165d400a02180e659a004f003619b006f61622003a0046200099080beed3a72a00d490600ee31a78a95144639109183e620308f004e4aef8283dac647ea3a6ca5400543050d81d9c600290a0c4a801e8824234fdfc13429c94606831dc904420e35293930382d3c2b272f20b4a8afc4b12c2322a3cabe26b7ceb0b1b33b25bf2db123b35eaca9b7b73fb0a86b4bbd27b2b27db92526bbb9a27ab0797a967cbc75aa78b63740d6ae3780492e2e8199959d80801b8c8c800c4dd651c30f891a0bcf0241c00198f118b921000b8600072003c841f43c11129a1001a0e0411cc028742001650280001c2010d8ac4816e391f1a03c2812180b2502316210588801148a52c4006e004600332c4009a007d00801a5b40036100003800ea206e7c800aa4a7e63045000d00270f00082fad901281d0b22a89e2f5024160c6080e0600065331891db88c365195f5a460011c450a8a180cd212784a25358f80110212a1fc802b3a2436188cd80042e8280811850a96263aa1f0e4600122047120005e6e245c7f394551268c91fb3a55cac0f180bc418e902c4281a120bec03a78071a011130c0688ea299c5e68ec66075ce2ce945e0cc13b350de63728abaf000643040a836f77307de38004a483e3e2cf33b2dafaf61a4020493000132361b140e39b2d320af0c07861cb81a15e4f85867cf02a110541d02c1626212d200)

**Expected behavior**
The top border should not have the white circle when it overlaps with the right border
<img width="853" alt="Screen Shot 2021-10-20 at 8 48 41 PM" src="https://user-images.githubusercontent.com/87088870/138191827-4846ee8f-1dc8-4851-991b-778281e0d07b.png">

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser: Chrome
 - React-pdf version: 2.0.19

jeetiss:
[...]

PR #1603, opened 2021-11-19
fix: border radius cut off bug fix

rjsduf0503:
Hello @diegomura ! 

We're undergraduate students taking a class contributing to opensource. 
We're interested in your repository and trying to find some issues that we can contribute.

So we found #1553 issue, and solved border-radius cut off bug.
We found that cut off bug occurs when border-width is greater than border-radius.

So, in the resolveRadius function of packages/layout/src/steps/resolvePercentRadius.js, when it is not a percent value, we changed this function to return the largest value among borderTopWidth, borderBottomWidth, borderLeftWidth, borderRightWidth, and value (the entered border-radius value).

Here are bug image and a capture image for solved version!
<br/>
- bug image
image
- fixed image
image

Thank you for reading this commit! Have a nice day :-) 

- Desktop 
   - OS : Windows 10
   - Browser : Chrome
   - React-pdf version : 2.0.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,6 @@
     "babel-plugin-add-module-exports": "^1.0.0",
     "canvas": "^2.8.0",
     "chalk": "^2.4.2",
-    "pdfjs-dist": "2.8.335",
     "eslint": "^6.8.0",
     "eslint-config-airbnb": "^18.1.0",
     "eslint-config-prettier": "^4.0.0",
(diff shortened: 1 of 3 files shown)


jeetiss:
@rjsduf0503 thanks for contribution, but this is not exactly the fix, because you change the border radius that user define and expect.

you need to fix clipping logic in https://github.com/diegomura/react-pdf/blob/master/packages/render/src/primitives/renderBorders.js#L8-L104

rjsduf0503:
Hello! @jeetiss .
Thanks for your advice.
We'll fix it according your advice! :)
Have a nice day!!!!

jeetiss:
@rjsduf0503 have a nice day too and thanks for contribution!
if you fix this you will be my hero because the clipping a little bit tricky :)

PR #1604, opened 2021-11-20
fix: skip empty text instance creation in jsx conditions

### LINKED ISSUE #1602: error Cannot read properties of undefined (reading 'height') at setNodeHeight
```js
const Component = ({ name }) => {
    return (
        <>
            {name &&
            <Text>
                {name}
            </Text>}
        </>
    );
}

<Component name='' />
<Component name={0} />

```

Example above cause fatal errors in `@react-pdf`, but it should render nothing.

[👉🏻 repl](https://react-pdf.org/repl?code=140630f60760ce02e0040613016c00e902985e05e3802806f382010d90ce017c04a38700f8e43838db80270c6015c3884b94a00c985c003c00543000f180d0990a55c407a69721b02ac177868f00228f009632c0c4a79f1d460558480226040f0ad8b7b090015480730c4f2f0924344c6c410a1c007268b85520af7150f4082c78250c1c4200062a78c489555f00a0b567577779601a006e5d00250c5210186f470031003a2e0800130c0e7c7163330b4a04bae02000)

### Discussed in https://github.com/diegomura/react-pdf/discussions/1599

<div type='discussions-op-text'>

<sup>Originally posted by **joseDaKing** November 17, 2021</sup>
I get this error in my code and I don't why

error:

TypeError: Cannot read properties of undefined (reading 'height')
    at setNodeHeight</div>

jeetiss:
Fix #1602


diegomura:
@jeetiss I resume this. I think throwing in these cases is too agressive. On this PR I just ignore these cases with a warning

packages/renderer/src/renderer.js
```diff
@@ -11,6 +11,23 @@ import propsEqual from './utils/propsEqual';
...
+  const isChildTextInstance = child.type === 'TEXT_INSTANCE';
+  const isOrphanTextInstance = isChildTextInstance && !isParentText;
+
+  // Ignore orphan text instances.
+  // Caused by cases such as <>{name && <Text>{name}</Text>}</>
+  if (isOrphanTextInstance) {
+    console.warn(
+      `Invalid '${child.value}' string child outside <Text> component`,
+    );
+    return;
```
likyh, line 25: Hi, I think there are some mistakes here. If we haven't push these node, that may got error when `insertBefore` called. 
```
    insertBefore(parentInstance, child, beforeChild) {
      const index = parentInstance.children?.indexOf(beforeChild);

      if (index === undefined) return;

      if (index !== -1 && child)
        parentInstance.children.splice(index, 0, child);
    },
```
child will be droped if beforeChild is not OrphanTextInstance

jeetiss: Could you provide a test that shows the error?

likyh:
```
import { Document, Font, Image, Link, Page, PDFViewer, StyleSheet, Text, View } from '@react-pdf/renderer'
import ReactMarkdown from 'react-markdown'
import { ReactMarkdownProps } from 'react-markdown/lib/ast-to-react'

const App = () => (
  <PDFViewer style={{ width: '100%', height: 800 }}>
    <Document>
      <Page style={styles.body}>
        {/*<Text>*/}
        <ReactMarkdown components={{
          h1: ({ children }: ReactMarkdownProps) => <Text>{children}</Text>,
          p: ({ children }: ReactMarkdownProps) => <Text>{children}</Text>,
        }}>{`# a
b`}</ReactMarkdown>
        {/*</Text>*/}
      </Page>
    </Document>
  </PDFViewer>
)
```

image

if add `<Text>` out of `ReactMarkdown`:

image

Actually, That can be found without ReactMarkdown. But I don't know when will insertBefore be called.

PR #1607, opened 2021-11-22
fix: import only `md5` hash from `crypto-js`

jeetiss:
`crypto-js` contains a lot of hash algorithms and functions, but `pdfkit` uses only `md5`
I change the imports to include only `md5` code to bundle, it should save `~42kb`

Снимок экрана 2021-11-22 в 10 29 52

### DIFF
--- a/.changeset/violet-pears-bake.md
+++ b/.changeset/violet-pears-bake.md
@@ -1,0 +1,5 @@
+---
+'@react-pdf/pdfkit': patch
+---
+
+replaced `crypto-js` with `crypto-js/md5` to reduce bundle size
--- a/packages/pdfkit/rollup.config.js
+++ b/packages/pdfkit/rollup.config.js
@@ -9,11 +9,11 @@ import pkg from './package.json';
 
 const cjs = {
   exports: 'named',
-  format: 'cjs',
+  format: 'cjs'
 };
 
 const esm = {
-  format: 'es',
+  format: 'es'
 };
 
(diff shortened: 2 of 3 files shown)


jeetiss:
should fix #1618

chris-morgan:
I’m seriously confused here, because as far as I can tell it’s just completely broken, so that AttachmentsMixin no longer works.

78bfe8cff9aba40b74c16fd6196b3185829d3129 made sense: replace `crypto-js` import with `crypto-js/md5` and `crypto-js/core`.

But then 849dca127f9413b616ad82720c1fc91b44edd4a5 ruined everything by retreating to putting everything in `crypto-js/md5`, but treating it as though it exported `CryptoJS`, whereas in actual fact it gives you `CryptoJS.MD5`. So the checksum line will always throw an exception because it has the wrong object. After puzzling about it for a bit, I’m guessing the purpose of the commit was to adjust the Rollup `external` piece, and took the wrong approach due to some fascination with mapping. Instead, I say, just add new externals, e.g. `[...Object.keys(pkg.dependencies), 'crypto-js/md5.js', 'crypto-js/core.js']`.

A secondary problem is that the import paths don’t work in ECMAScript Modules: you need the file extension, so 'crypto-js/md5' should be 'crypto-js/md5.js', and 'crypto-js/core' 'crypto-js/core.js'. This change would restore this piece to working in both CommonJS and ESM. (This is related to #2068.)

Issue #1625, opened 2021-11-29
upgrade husky and lint-staged


jeetiss:
**Describe the bug**

husky and lint-staged outdated and have config that forces unstaged changes on every commit.

**To Reproduce**
0. create branch
1. modify two files
2. run `git add first.file`
3. run `git commit -m 'modify first.file'`
4. all files are committed because of husky and this is weird

**Expected behavior**

Git hooks work as before but don't stage all files on commit.

PR #1628, opened 2021-12-02
fix : bug fix about husky & lint-staged

rjsduf0503:
Hello! @jeetiss :)

We're here again~~
We read your issue #1625 , and we think that we solved this issue!

In the package.json file, "husky" part might cause problem.
In "pre-commit" part, we think "git add ." is useless and it cause bug.
So we deleted "git add ." and when we add/commit 1 file of 2 file(or more),  it was confirmed that only the added file was uploaded.

If our fix is wrong, please comment us!
Thank you for reading this PR, and have a wonderful day~!

Desktop
- OS : Windows 10
- Browser : Chrome
- React-pdf version : 2.0.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "git add . && lint-staged"
+      "pre-commit": "lint-staged"
     }
   },
   "dependencies": {}


jeetiss (approved):

jeetiss:
Nice catch, it works fine now

rjsduf0503:
Thanks a lot!

PR #1644, opened 2021-12-14
chore: add simple benchmark

jeetiss:


diegomura:
@jeetiss closing this. Code won't be lost. Might be interesting to have benchmarking as a separate package or in CI

PR #1659, opened 2021-12-22
fix: use styles for forcing nodes height after split

### LINKED ISSUE #1634: View component does not fill the remaining space on the page after 2.0.21 update
**Describe the bug**
There seems to be trouble with the View component render after updating to 2.0.21.
If the View component starts on the 1 page and end on the 2 page - it does not fill the remaining space on the 1 page (screenshot 2) - but it should.
The problem appears if the View component has child components that have not enough space to be rendered on the previous page of the pdf document

**Expected behavior**
View component should fill the remaining space to the end of the page
Снимок экрана 2021-12-06 в 17 32 25

**Actual behavior**
View component breaks itself causing fake padding at the bottom of the page
Снимок экрана 2021-12-06 в 17 36 04

Has anybody faced this issue?

jeetiss:
fix #1634

### TODOs

- [x] remove logs
- [x] add test names
- [x] add changeset

### DIFF
--- a/.changeset/angry-bikes-allow.md
+++ b/.changeset/angry-bikes-allow.md
@@ -1,0 +1,5 @@
+---
+'@react-pdf/layout': patch
+---
+
+use styles for forcing nodes height after split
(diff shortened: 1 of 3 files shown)


diegomura (approved):
LGTM!

PR #1660, opened 2021-12-23
chore: upgrade to yarn 3, remove lerna and unused dependencies

jeetiss:
- upgrade to yarn 3
- replace lerna with yarn workspaces plugin
- remove unused dependencies


chathu-novade:
Did you removed husky?

chathu-novade:
Isn't it better to use pnpm and it got good support with changeset too.

diegomura:
Doing some cleanup. Will close this as it's been left here. But please resume it in the future if you please!

PR #1609, opened 2021-11-22
fix: bundle latest `dfa` to fontkit

jeetiss:
I updated `dfa` to latest version, added it to browser bundle and process with babel to avoid problems with modern code as in #1061

deduplicate `regenerator-runtime` with update to `@babel/runtime/regenerator`
Снимок экрана 2021-11-22 в 20 30 35

and skip `core-js@2` that should fix #1608

PR #1673, opened 2022-01-10
fix: proper ignore node `fs` package in browser bundle

jeetiss:
was broken in #1609


tomaszzmudzinski:
@jeetiss could you tell us when all fixes/improvements from december 2021 / january 2022 will be published as a new release?

PR #1681, opened 2022-01-19
fix: define renderToStream to return promise

dschu-lab:
The old return type was `string` which is wrong, because it returns a `promise` which resolves in a `string` instead.

This pull request fixes the warning `'await' has no effect on the type of this expression` when using it like this:
```ts
    const document = await ReactPDF.renderToString(...)
```


jeetiss:
@dschu-lab thanks for your contribution! 🙌🏻

PR #1715, opened 2022-02-12
chore: upgrade to latest ramda

redbmk:
Fixes a vulnerability with an outdated version of ramda: https://snyk.io/test/npm/@react-pdf/renderer/2.1.1

### DIFF
--- a/packages/textkit/src/attributedString/runAt.js
+++ b/packages/textkit/src/attributedString/runAt.js
@@ -10,7 +10,7 @@ import runIndexAt from './runIndexAt';
  * @return {Object} run
  */
 const runAt = (n, attributedString) => {
-  const runIndex = runIndexAt(n)(attributedString);
+  const runIndex = R.max(runIndexAt(n)(attributedString))(null);
   return R.path(['runs', runIndex])(attributedString);
 };
 
--- a/packages/examples/package.json
+++ b/packages/examples/package.json
@@ -13,7 +13,7 @@
   "dependencies": {
     "@react-pdf/renderer": "^2.0.6",
     "camelcase": "^6.2.0",
-    "ramda": "^0.27.1"
+    "ramda": "^0.28.0"
   },
   "peerDependencies": {
     "react": "^16.8.6 || ^17.0.0",
--- a/packages/layout/package.json
+++ b/packages/layout/package.json
@@ -28,7 +28,7 @@
     "cross-fetch": "^3.1.5",
     "emoji-regex": "^8.0.0",
     "queue": "^6.0.1",
-    "ramda": "^0.26.1"
+    "ramda": "^0.28.0"
   },
   "devDependencies": {
     "jest-fetch-mock": "^2.1.1"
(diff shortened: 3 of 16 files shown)


jeetiss:
@redbmk thanks for your contribution
upgrade to the latest `ramda` is not so easy, because `ramda@0.28.0` contains breaking changes

redbmk:
@jeetiss ah, bummer. I wondered about that - I didn't see anything in Ramda's documentation about breaking changes so I was just hoping 🤞 .

I might take another look at this and see if I can find an upgrade guide - hopefully the changes aren't too big. It also sounds like 0.27.2 fixes the vulnerability which might be a safer upgrade.

redbmk:
I did just find a couple upgrade guides - I had expected them to be in the release notes, but found them in Issues:

Upgrade to 0.27.0: https://github.com/ramda/ramda/issues/2978
Upgrade to 0.28.0: https://github.com/ramda/ramda/issues/3218

redbmk:
@jeetiss Sorry, I keep having issues with the lock file and one of the tests wasn't passing. I got the tests all working locally and I did `npx lerna clean && yarn bootstrap` and it's able to install now so I think we should be good.

I did notice that `yarn check` though actually throws an error that seems unrelated to what I was working on:

```
yarn check v1.22.11
error "rollup-plugin-terser#terser#acorn@^8.5.0" doesn't satisfy found match of "acorn@7.4.1"
error Found 1 errors.
info Visit https://yarnpkg.com/en/docs/cli/check for documentation about this command.
```

packages/textkit/src/attributedString/runAt.js
```diff
@@ -10,7 +10,7 @@ import runIndexAt from './runIndexAt';
  * @return {Object} run
  */
 const runAt = (n, attributedString) => {
-  const runIndex = runIndexAt(n)(attributedString);
+  const runIndex = R.max(runIndexAt(n)(attributedString))(null);
```
redbmk, line 13: This was needed because `runIndexAt` can return `-1`, and `R.path` now supports negative indices, so instead of returning `undefined` it would return the last element.

https://github.com/ramda/ramda/pull/2670

This change seems to be a bit controversial as [it was reported as a bug](https://github.com/ramda/ramda/issues/2974) and there's [an open PR to revert it](https://github.com/ramda/ramda/pull/3116), but that's been open for over a year now so it seems like we should roll with the change.

diegomura:
The goal is to get rid of ramda eventually. I still like it and was helpful initially but I think it ends up creating a lot of confusion

redbmk:
@diegomura it's been a while since I looked at this but iirc I got was able to upgrade and get tests passing. Is there harm in upgrading this to fix the vulnerabilities until you remove ramda, or is that happening sooner rather than later?

diegomura:
IMO the risk of globally upgrading it is very big. Tests aren't exhaustive, so although it's a nice check doesn't mean there aren't any issues in the thousands of invocations we do of ramda. 

I really appreciate the contribution, but currently I don't have any bandwidth ether to respond and debug this kind of potential bugs. If the shared security concern is specially important to fix in your case you can always configure the new versions and diffs on your own project under our own risks. 

Ill try to finish the ramda removal very soon (that's also a bit risky) so hopefully next version won't have ramda vulnerabilities either way

redbmk:
oh yeah that makes sense, thanks

PR #1725, opened 2022-02-21
Add support for react 17

vegai:
https://github.com/diegomura/react-pdf/issues/1062


vegai:
This doesn't properly fix all the issues related to #1062 though, react-conciler is still a problem with `--strict-peer-deps`.

vegai:
Also did not test this with React < 17 at all.

diegomura:
This isn't enough, and breaks React 16 support. I'll work on this asap

Issue #1732, opened 2022-02-25
Incompatible with webpack 5 in case insensitive systems due restructure dependency


Sasurtio:
**Describe the bug**
When webpack compiles the node_modules folder in a case insensitive operative system (like Windows or MacOS), in the `restructure` dependency, it generates a file called `buffer.js` instead of `Buffer.js` for the development server.

When include the shim buffer library (following the README instructions), webpack will do an automatic `require` whenever it founds the unasigned variable `Buffer` (see highlighted line).

![image](https://user-images.githubusercontent.com/81442463/155771010-80a1c993-78c2-4875-8420-e5b991cbb07c.png)

Because of that, when the app tries to require the shim using `Buffer` what it really does is a `var Buffer = require('buffer/').Buffer`

![image](https://user-images.githubusercontent.com/81442463/155770685-e97b1f49-6cbc-4a65-925b-94f400beffb4.png)

But, because of the compilation of webpack 5 with case insensitive OS, it will find the file `buffer.js` that is relative to the `DecodeStream.js` file instead of the buffer library and result in `var Buffer = undefined`

**Expected behavior**
`var Buffer = Object`

**Desktop (please complete the following information):**
 - OS: Windows 10, MacOS
 - Browser: Edge
 - React-pdf version: 2.1.1

diarcastro:
@Sasurtio perhaps this solution can help you as well!
https://github.com/diegomura/react-pdf/issues/1740#issuecomment-1065509181

Issue #1029, opened 2020-10-14
Incompatible with Webpack 5 due to Buffer dependency


csvan:
Webpack 5 removes Buffer (see https://webpack.js.org/migrate/5/), effectively breaking this library since it has explicit usages of it in the code.

gilberto1991:
I also have this problem

GregdTd:
Same here, any news ?

GregdTd:
I was finally able to make it works.. 
`npm install assert browserify-zlib buffer process stream-browserify util`

Then in my webpack config:
```
const webpack = require('webpack');

module.exports = {
 {...}
    resolve: {
        alias: {
            process: 'process/browser',
            stream: "stream-browserify",
            zlib: "browserify-zlib"
        }
    },
    plugins: [
        new webpack.ProvidePlugin({
            process: 'process/browser',
            Buffer: ['buffer', 'Buffer'],
        }),
    ]
}
```

matheusgrieger:
@GregdTd that didn't work for me though...

csvan:
@matheusgrieger I solved it in the same way. What error are you getting? What if you add buffer under `resolve.fallback` as well or instead of `resolve.alias`?

matheusgrieger:
@csvan tried that but didn't work. Actually my error was being caused by a state update in another component while the PDF was rendering. Though it didn't change anything at all about the PDF content, it was crashing the app.

Some context would be good, too. I'm running React PDF in Electron and don't even need these fallbacks at all...

mstrk:
@matheusgrieger I also made it work with `resolve.fallback` but did not had to install `buffer` for example. and I think @GregdTd is missing some alias like `assert` and `util` that where installed.

I know that it's probably late and you already fixed it or move on, but if you have some error log to show us, then we could help you debug that.

matheusgrieger:
@mstrk I will happily get some logs to help you debug. But won't be possible for now though, as I have to change the way I wired up the app.

osimuka:
```

```
```Uncaught ReferenceError: Buffer is not defined
    at eval (DecodeStream.js:31)
    at Object.eval (DecodeStream.js:98)
    at eval (DecodeStream.js:102)
    at Object../node_modules/restructure/src/DecodeStream.js (bundle.js:9842)
    at __webpack_require__ (bundle.js:10651)
    at fn (bundle.js:10882)
    at Object.eval (EncodeStream.js:9)
    at eval (EncodeStream.js:151)
[...]

PR #1733, opened 2022-02-25
doc: updated README to work for case insensitive systems using webpack 5

Sasurtio:
Solves #1732
https://github.com/diegomura/react-pdf/issues/1029#issuecomment-1050397049
https://github.com/diegomura/react-pdf/issues/1029#issuecomment-997211684

Instead of looking for the buffer library using require('buffer').Buffer (that will look first for relative files and then for global path), it will look for a relative path to the node_modules/buffer/
In that way, we will be sure to import the library even if we have a non-case sensitive OS

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@ Webpack 5 doesn't include node shims automatically anymore and we must opt-in to
 
 ```sh
 yarn add process browserify-zlib stream-browserify util buffer assert
+yarn add -D imports-loader
 ```
 
 after the modules are installed, we need to adjust our `webpack.config` file:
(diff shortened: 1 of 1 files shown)


Sasurtio:
@jeetiss 
Hey, could you kindly review this PR?

https://github.com/diegomura/react-pdf/issues/1029#issuecomment-1059644087

It should help with the case-unsensitive Operative Systems like Windows and MacOS

diegomura:
Not convinced to add this in the README. We can't have details for all build tools there. Will close it but it will be left as docs for anyone searching for this issue

Issue #1811, opened 2022-05-06
Document and Page Typescript typings are missing a children property


pganster:
**Describe the bug**
As of lately, neither `Page` nor `Document` component can be used with children. I've copied the code from the quick start guide on https://react-pdf.org/ into a freshly created React app and receive the following Typescript issues:

```
ERROR in src/index.tsx:23:4
TS2769: No overload matches this call.
  Overload 1 of 2, '(props: DocumentProps | Readonly<DocumentProps>): Document', gave the following error.
    Type '{ children: Element; }' has no properties in common with type 'IntrinsicAttributes & IntrinsicClassAttributes<Document> & Readonly<DocumentProps>'.
  Overload 2 of 2, '(props: DocumentProps, context: any): Document', gave the following error.
    Type '{ children: Element; }' has no properties in common with type 'IntrinsicAttributes & IntrinsicClassAttributes<Document> & Readonly<DocumentProps>'.
    21 | // Create Document Component
    22 | const MyDocument = () => (
  > 23 |   <Document>
       |    ^^^^^^^^
    24 |     <Page size="A4" style={styles.page}>
    25 |       <View style={styles.section}>
    26 |         <Text>Section #1</Text>

ERROR in src/index.tsx:24:6
TS2769: No overload matches this call.
  Overload 1 of 2, '(props: PageProps | Readonly<PageProps>): Page', gave the following error.
    Type '{ children: Element[]; size: "A4"; style: { flexDirection: "row"; backgroundColor: string; }; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<Page> & Readonly<PageProps>'.
      Property 'children' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<Page> & Readonly<PageProps>'.
  Overload 2 of 2, '(props: PageProps, context: any): Page', gave the following error.
    Type '{ children: Element[]; size: "A4"; style: { flexDirection: "row"; backgroundColor: string; }; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<Page> & Readonly<PageProps>'.
      Property 'children' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<Page> & Readonly<PageProps>'.
    22 | const MyDocument = () => (
    23 |   <Document>
  > 24 |     <Page size="A4" style={styles.page}>
       |      ^^^^
    25 |       <View style={styles.section}>
    26 |         <Text>Section #1</Text>
[...]

PR #1825, opened 2022-05-24
Add children props

allondeveen:
Proposed solution to issue #1811.


chathu-novade:
@diegomura Shouldn't this be in @react-pdf/types too?

PR #1827, opened 2022-05-25
refactor: remove ramda from layout package

diegomura:
Gradually start removing ramda. Although I like it, it's probably confusing for most people, it's slow and adds unneccesary bundle size. Some helper fns will still be needed. For the moment I moved them to utils dir, in the future it worths building our own set of light and reusable utils

PR #1829, opened 2022-05-25
refactor: remove ramda from stylesheet package

diegomura:
Related to https://github.com/diegomura/react-pdf/pull/1827

### DIFF
--- a/packages/fns/castArray.js
+++ b/packages/fns/castArray.js
@@ -1,0 +1,11 @@
+/**
+ * Casts value to array
+ *
+ * @param {any} value
+ * @returns {Array} casted value
+ */
+const castArray = value => {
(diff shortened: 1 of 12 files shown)


packages/fns/castArray.js
```diff
@@ -0,0 +1,11 @@
...
+ * Casts value to array
+ *
+ * @param {any} value
+ * @returns {Array} casted value
+ */
+const castArray = value => {
+  return Array.isArray(value) ? value : [value];
+};
+
+export default castArray;
```
diegomura, line 11: Right now this `fns` dir (is there a better name? I like to keep it short and simple) isn't technically a package, and each package importing from here will include it's own copy of this helpers. That's a small price to pay. I plan to later convert it into our own set of "lodash" package

Issue #933, opened 2020-06-26
Font Fallback/Multilanguage support


oowowaee:
I know when looking at github issues over the last few days, I saw a thread somewhere that linked to https://github.com/diegomura/react-pdf/blob/1c0bcba15e1cddda6ae9adbba4bedf680e98c9fd/src/layout/fontSubstitution/index.js#L11 in the context of showing font fallbacks.

I have a use case where some of the output could be in one of many languages, and obviously this is causing an issue, because the font cannot register those glyphs.  On the web, we don't have this issue, because the fonts fall back.  How are you supposed to be able to specify font fallbacks, and more specifically, how can we support text in a language that isn't known at style time?  (Ie, I can't just do styles.korean and set a Korean font-family, as the text could be in Arabic, Chinese etc).

chimanSeyedhashemi:
Hello, I have problem with persian Font in react pdf, it show the last line in the first line. how i can solve this problem?
[b39af48d-873d-4433-89a7-b5f5a51bae78.pdf](https://github.com/diegomura/react-pdf/files/5000321/b39af48d-873d-4433-89a7-b5f5a51bae78.pdf)

oowowaee:
@chimanSeyedhashemi React-PDF doesn't support RTL languages.  My use case luckily is only names in Arabic.

canturan:
I also find the multilanguage support pretty important. For example, if you have some chinese users, they are writing mostly in english with some chinese words in between. It is impossible to find a right font for both and to render a pdf document, which supports both languages.

SashaShostyr:
> I also find the multilanguage support pretty important. For example, if you have some chinese users, they are writing mostly in english with some chinese words in between. It is impossible to find a right font for both and to render a pdf document, which supports both languages.

Yeah, it will be cool to have multilanguage support. @diegomura Please do it :bow:, thanks!

oowowaee:
@canturan I know when I was researching this it seemed like there were some threads where people suggested working Chinese fonts (and I know we ended up having Chinese support in our PDF), and I thought the default is Helvetica(?)...so I think if it's only Chinese/English you want that might be doable, that didn't work for you?

I was able to get multilanguage support in the end coming up with a really hacky workaround where I overrode the font-fallback logic, but never got it working with loading the fonts on the fly.  It seems like needing to be able to fallback to nonEnglish language fonts, or to multiple fonts is something a bunch of people need!

SashaShostyr:
[...]

Issue #856, opened 2020-03-23
Non Latin Hindi Unicode Characters Not Rendering


DeepakKapiswe:
Non Latin Characters specifically  Hindi Characters are not getting rendered  properly, some characters even words change to blank while rendering.

My sample code:
```
import React from "react";
import notoSansHindi from '../../assets/fonts/NotoSansDevanagari-Regular.ttf' 
import {
    Page,
    Text,
    View,
    Document,
    StyleSheet,
    Font
} from "@react-pdf/renderer";

const BORDER_COLOR = '#bfbfbf'
const BORDER_STYLE = 'solid'
const COLN_WIDTH = 100 / 3
const FONT_FAMILY = 'Noto Sans Hindi'

Font.register({
    src: notoSansHindi,
    family: FONT_FAMILY
    }
  );

Font.registerHyphenationCallback(word => [word]);

const styles = StyleSheet.create({
    body: {
        padding: 8,
    },
    table: {
        display: "table",
        width: "auto",
        borderStyle: BORDER_STYLE,
        borderColor: BORDER_COLOR,
    },
    heading: {
        textAlign: 'center',
        fontFamily: FONT_FAMILY,
    },
});

export default function PdfSubscriberList() {
    return (
        <Document>
            <Page size="A4" style={styles.body}>
                <View style={styles.table}>
                    <Text style={styles.heading} >
                    श्री माधव त्रिपाठी शर्मा
                    </Text>
                </View>
            </Page>
        </Document>
    );
}
```
which results in :

श्री माधव त्रिपाठी 

**Expected behavior**
श्री माधव त्रिपाठी शर्मा

This is just one sample, I have I can see many bugs rendering with half words, character missing etc.

Please help ! Tried almost everything but not working. I hope someone else must have fixed this library to use with Hindi fonts.

Desktop
 - OS: MacOS
 - Browser [chrome, safari]
 - React-pdf version v1.6.8

chimanSeyedhashemi:
i have this problem with persian font. please fix these bug i need it

DeepakKapiswe:
I switched to pdf-make and that library is awesome @chimanSeyedhashemi

darrynten:
Same issue with Vietnamese characters

PR #1833, opened 2022-05-28
feat: Add Font Fallback

### LINKED ISSUE #1771: Font not rendering special characters
So in my app i have multiple language options (English/Chinese/Japanese/etc). Im using **"lato", sans-serif** font fallback for Chinese/Japanese and some other language special characters and it renders fine. 

```
import LatoRegular from '@/.../.../Lato-Regular.ttf'

Font.register({
  family: 'LatoRegular',
  format: 'truetype',
  src: LatoRegular
})
```

So lato doesn't work for all the characters. Since font fallback is not possible in this library (correct me if I'm wrong?), is there a font that supports all, latin special characters, Chinese and Japanese at the same time? I already read the posts here and 'Zboom' font (or whats it called?) doesn't fit for me, it looks ugly when rendering latin chars.

Any workaround for this would be appreciated. Thank you

ahmed-novalabs:
- Updated and Optimized #508 for version 2.0.
- Added some basic tests for the font package

Should fix #1771,  #499.
Is part of #933, #856.


diegomura:
Thanks @ahmed-novalabs for working on this! Can you provide a test plan for this PR?

If I understood correctly, user will need to provide a `unicodeRange` to each font definition in the form of a regex. I find this a bit odd, in the sense that if feels like it's exposing something to the user that should be possible to compute inside. Isn't it? Most people doesn't know what unicode range the font they are using covers, so by exposing this API I feel many will be confused. Thoughts?

ahmed-novalabs:
@diegomura I think the main benefit is optimization by only loading the used fonts based on the text and `unicodeRange`. 
The behavior should be to load all the used fonts that don't have a `unicodeRange` and only load fonts with `unicodeRange` if they are used and the regex matches the text.
[There is a similar CSS property](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/unicode-range).
I am not sure if there is a better way to achieve this.

For the testing plan, I don't have a lot of experience with testing, so I am not sure how to make or approach that. If you can provide an example or a guide of what you expect, I can work on it.

ahmed-novalabs:
What's in this PR
- Added `unicodeRange` to the font object to support loading only needed fonts 
- Added support for multiple fonts per `<Text/>` by modifying the font substitution stage to split the font by `,` and going through the list of fonts one by one till a font that supports the current glyph is found.

Things to test: 
- Pass multiple fonts in the form of `Font1, Font2` with characters that aren't supported by the first font and see if the character is rendered in Font2 correctly. (Also with multiple languages)
- Set a limited `unicodeRange` for a font and try rendering the document once with at least one character that matches the range and another time with no characters that match the range. The font should be loaded the first time, but not the 2nd time.
- Try rendering text with one font, and then adding more text that needs another font to be rendered and update using the update function in React and see if it works correctly (Causes a weird bug).
- Test in node.js

chathu-novade:
Any update on this?

chathu-novade:
@ahmed-novalabs Is there specific reason to go with new parameter `fontStack` than allowing fontFamily to be a string or array?

chasovskiy:
Any updates on this PR?

ahmedwalid05:
I lost access to the `ahmed-novalabs` account.

@chathu-novade I am not sure I understand the question, `fontStack` is an internal variable that holds the font stack, it's not for use by the user.
The user can just pass multiple fonts using the `fontFamily` attribute just like HTML. 
e.g.: 
```js
fontFamily: "MyFont1, MyFont2"
```

nikgraf:
@diegomura the unicode-range is a standard CSS feature https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/unicode-range. For example in one project we use it to make sure all numbers are rendered with the monospaced version of the font for better readability.

Anyways, but it already would be very helpful to allow to support multiple fonts to support various glyphs e.g. arabic, cyrillic. Most fonts offer dedicated fonts for arabic and so on, because apparently there is a limit of 64K glyphs for a TrueType file (https://github.com/notofonts/noto-fonts/issues/13#issuecomment-110099560). So you can get Noto Sans and Noto Sans Arabic. Combining multiple fonts is one solution, but at some point you hit the 64K limit again.

The easiest going forward afaik would be to allow defining multiple fonts e.g.
`fontFamily: "NotoSans, NotoSansArabic"` or `fontFamily: ["NotoSans", "NotoSansArabic"]`. If an arabic glyph is not found in `NotoSans` it would fall back to  `NotoSansArabic`.

If you think this is a good idea I would be happy to extract this part from the PR and create a new one.

diegomura:
@nikgraf i think it is!

nikgraf:
great, will work on it in the coming days

nikgraf:
PR is ready: https://github.com/diegomura/react-pdf/pull/2640

nikgraf:
@diegomura wanted to ask if you could review the PR 🔝

diegomura:
We already support this. Closing this

PR #1834, opened 2022-05-28
refactor: remove ramda from render

diegomura:
## Summary


packages/render/src/primitives/renderSvg.js
```diff
@@ -193,79 +150,80 @@ const draw = ctx => node => {
...
+};
+
+const renderNode = (ctx, node) => {
+  const renderFn = renderFns[node.type];
+
+  if (renderFns) {
+    renderFn(ctx, node);
+  } else {
+    console.warn(`SVG node of type ${node.type} is not currenty supported`);
+  }
```
kidroca, line 179: Hello,

I was going through the code and I noticed a small potential oversight. It seems that the condition `if (renderFns)` would always evaluate to `true` given the nature of JavaScript's truthy/falsy evaluation. I believe it might have been intended to be `if (renderFn)` to check for the existence of the specific render function:

```js
  const renderFn = renderFns[node.type];

  if (renderFn) {
    renderFn(ctx, node);
  } else {
    console.warn(`SVG node of type ${node.type} is not currently supported`);
  }
  ```
  
Thank you for all your hard work on this repository. I hope this helps!

PR #1836, opened 2022-05-30
refactor: remove unused ramda dependency from renderer

diegomura:


davbrito:
This is cool. Some time ago I tried to see the code and it was hard to understand because of the usage of Ramda.

PR #1869, opened 2022-06-06
feat: variable dpi

### LINKED ISSUE #1066: dpi / resolution prop (300 dpi)
Please add a 300 dpi version. Generating PDFs for printing with 72 dpi is a no-go.
300 dpi is a must have!

How about a document prop "dpi" (number like 300) or "resolution" (string / boolean like "low - high" / true - false)?

diegomura:
Adds optional `dpi` prop to `Page` primitive

**Why Page and not Document?**

So the same document can have different DPIs in different pages

**What this prop does under the hood**

- Resizes page dimensions to match new dpi 
- Setup PDF page `UserUnit` field to dpi/72 

Reading at the PDF spect it seems that the latter point is the way to go, but I haven't seen any change in how readers (specially Adobe) renders pages or shows page dimensions. Seems like it's being ignored as far as I saw. But hopefully this will work for anyone requiring to have variable dpi.

Please open a new issue if someone trying this finds something not working as expected 

Fixes #1066

### DIFF
--- a/.changeset/pretty-jars-begin.md
+++ b/.changeset/pretty-jars-begin.md
@@ -1,0 +1,9 @@
+---
+'@react-pdf/layout': minor
+'@react-pdf/pdfkit': minor
+'@react-pdf/render': minor
+'@react-pdf/stylesheet': minor
+'@react-pdf/types': minor
+---
+
+feat: variable dpi
--- a/packages/layout/src/page/getSize.js
+++ b/packages/layout/src/page/getSize.js
@@ -71,6 +71,18 @@ const toSizeObject = v => ({ width: v[0], height: v[1] });
 const flipSizeObject = v => ({ width: v.height, height: v.width });
 
 /**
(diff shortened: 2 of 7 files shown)


nikischin:
So I had a look through the sources and did some research, and PDFKit should do the job with the userUnit already. I suspect, the changes in `packages/layout/src/page/getSize.js` should be reverted and the dpi variables in `packages/layout/src/steps/resolveStyles.js` and `packages/stylesheet/src/transform/units.js` should be fixed to 72 instead of the user provided dpi, as for my understanding PDFKit internally uses the 72 dpi as measuring unit and transforms it with the userUnit to the expected dpi. 

Unfortunaly, I haven't managed to get the packages linked locally, so I wasn't able to test. I created a branch https://github.com/nikischin/react-pdf/tree/bugfix/customDpi, maybe you could try this @diegomura

PR #1870, opened 2022-06-06
chore(render): tweak rollup config

jeetiss:
- remove minified build because nobody uses it and it takes so much time to compile.

- exclude `@babel/runtime` from the final bundle, it has weird imports like `@babel/runtime/**` so default approach with `Object.keys(pkg.dependencies)` doesn't work.

### DIFF
--- a/packages/render/rollup.config.js
+++ b/packages/render/rollup.config.js
@@ -19,7 +19,7 @@ const configBase = {
   input: 'src/index.js',
   external: Object.keys(pkg.dependencies).concat(
     /@babel\/runtime/,
-    /@react-pdf\/textkit/,
+    /@react-pdf/,
   ),
   plugins: [
     json(),
--- a/packages/render/rollup.config.js
+++ b/packages/render/rollup.config.js
@@ -1,6 +1,5 @@
 import json from '@rollup/plugin-json';
 import babel from '@rollup/plugin-babel';
-import { terser } from 'rollup-plugin-terser';
 
 import pkg from './package.json';
 
@@ -18,7 +17,10 @@ const getESM = override => Object.assign({}, esm, override);
 
 const configBase = {
   input: 'src/index.js',
-  external: Object.keys(pkg.dependencies),
+  external: Object.keys(pkg.dependencies).concat(
+    /@babel\/runtime/,
+    /@react-pdf\/textkit/,
+  ),
   plugins: [
     json(),
     babel({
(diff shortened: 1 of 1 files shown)


jeetiss:
@diegomura I am so happy that you are back 🙌🏻

diegomura:
Thanks man! That's very kind. Glad to be back as well, as much as I can 😄 

> remove minified build because nobody uses it and it takes so much time to compile.

This just made my day 😂 I was questioning why we were doing it lately. Your confirmation that isn't needed was all I needed haha.

In the topic: 

- Do we still need to ship CJS and ESM? Cannot we just create `type: module` packages? It is of my understanding that major bundlers and latest Node support module importing out of the box
- Should we bump min Node version from 12 to 14? 14 is in maintenance mode, while 12 is not even in it. I suspect that bundles will be thinner and faster due to unnecessary transformations, but I'm not sure and I haven't tested it

packages/render/rollup.config.js
```diff
@@ -36,12 +38,4 @@ const config = Object.assign({}, configBase, {
   ],
 });
 
-const prodConfig = Object.assign({}, configBase, {
```
diegomura, line 39: @jeetiss maybe we should remove the min versions in all of the packages?

jeetiss: Yep, we can do it, but it has the problem as with ESM bundle. Truly this is a breaking change for a package, but I think that we can write the sentence about it in a changelog and everything should be okay because nobody uses it

diegomura: We can remove minified bundles in other packages as separate PRs. Let me know how you prefer proceeding 😄 Regarding the major bump, I'm more than ok on proceeding with major! I definitely see your point.

jeetiss:
> Do we still need to ship CJS and ESM? Cannot we just create type: module packages? It is of my understanding that major bundlers and latest Node support module importing out of the box

it is harder because some tools have problems with native modules for example `jest` has
and I'm sure that it will work! lets try

> Should we bump min Node version from 12 to 14? 14 is in maintenance mode, while 12 is not even in it. I suspect that bundles will be thinner and faster due to unnecessary transformations, but I'm not sure and I haven't tested it

Yep, it should be fine to do, and we can skip all babel stuff and ship es6 code, because all major browsers support that

diegomura:
Great! Maybe we should do this gradually, just in case both changes are too aggressive together and people start having issues. Maybe bumping Node would be the first I'd do, since it has good impact on bundle. 

> we can skip all babel stuff and ship es6 code

You mean not transpiling code with Babel? Would that work for Node 14 and latest browsers at the same time?

PR #1871, opened 2022-06-06
chore: bundle layout package to esm

jeetiss:
`@react-pdf/layout` was only being published as a CJS module, so it is better to have `esm` too


diegomura (approved):
lgtm!

PR #1902, opened 2022-06-29
feat: add onClick event to PDFDownloadLink

### LINKED ISSUE #975: download pdf on custom button click
hi,

Is there any way i can download my pdf on my custom button click without using in built PdfDownloadLink ?

adamduncan:
Based on request #975 and our own experience, having an `onClick` handler to invoke for `PDFDownloadLink` would be benficial. E.g. We'd like to send a custom event to our analytics on download. (We're finding other means to lazily load the heavier components, so wouldn't need to resort to the workaround for this feature [proposed there](https://github.com/diegomura/react-pdf/issues/975#issuecomment-674261430).)

This implementation refactors the existing `onClick` handler slightly, to also allow consumer's to provide a function via props. We conditionally call this function after existing `handleDownloadIE`. (We'd generally opt for `onClick?.()`, but as this package targets Node v12, we'll check `typeof` before invoking.)

I've included the `event` as the first function argument, should the consumer need to use native event properties. That seems like the expected behaviour. Secondly, we pass along the `instance`, which might be helpful in certain circumstances. If that's not a beneficial use-case, would gladly omit.

Added types. Not sure that this component has unit tests, but happy to add if need be.

Fixes #975

### DIFF
--- a/.changeset/silly-rice-chew.md
+++ b/.changeset/silly-rice-chew.md
@@ -1,0 +1,5 @@
+---
+'@react-pdf/renderer': minor
+---
+
+feat: add onClick event to PDFDownloadLink
--- a/packages/renderer/index.d.ts
+++ b/packages/renderer/index.d.ts
@@ -448,6 +448,7 @@ declare namespace ReactPDF {
     children?:
       | React.ReactNode
       | ((params: BlobProviderParams) => React.ReactNode);
+    onClick?: function;
   }
 
   /**
--- a/packages/renderer/src/dom/PDFDownloadLink.js
+++ b/packages/renderer/src/dom/PDFDownloadLink.js
@@ -10,6 +10,7 @@ export const PDFDownloadLink = ({
   className,
   document: doc,
   fileName = 'document.pdf',
+  onClick,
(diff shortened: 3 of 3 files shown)


diegomura:
Thanks so much!

aclec:
Hi, I have a problem with your Typescript Type: 
`
node_modules/@react-pdf/renderer/index.d.ts:451:15 - error TS2552: Cannot find name 'function'. Did you mean 'Function'?

451     onClick?: function;
                  ~~~~~~~~

  node_modules/typescript/lib/lib.es5.d.ts:324:13
    324 declare var Function: FunctionConstructor;
                    ~~~~~~~~
    'Function' is declared here.


Found 1 error in node_modules/@react-pdf/renderer/index.d.ts:451
`

miladehghani:
@adamduncan I still can see the issue @aclec mentioned on @react-pdf/renderer@2.3.0 have you fixed it on that version?

adamduncan:
@miladehghani I believe the PR has been merged into master but a fixed 2.3.1 patch version of that package hasn't yet been released

Issue #1024, opened 2020-10-10
Need calculated height to print tickets


zariweyo:
I need to print tickets, which are in 2 width: 57mm and 80mm, but the height would be calculated to print in only one page.

I'd like to can set size height as Infinity, so then, this height will be calculated to print exactly my content height.

Other alternative in to get method which give me the height of the document I renderer.

FabianSchempp:
If the size of a ticket is fixed, isn't it possible to calculate the needed height from the amount of tickets you want to print?

zariweyo:
In my case, I print a list of products, so it's not fixed. I solve now with a function to calculate the height of each product text, but it's bad solution, because the length of each word can be change the amount lines to print. 

The correct solution will be to set infinity height o a method which give me the real height of each component renderer, to use in that function I say.

Look this solution in library to generate pdf in Dart. [Link to page formats](https://github.com/DavBfr/dart_pdf/blob/master/pdf/lib/src/page_format.dart)

They use infinity value to the height in roll57 and roll80, that are the formats for tickets. I think this is the correct way to solve this.

fulvous:
Hi! I am also interested in printing tickets. I was thinking of having a new page format where only width was specified. Disable all the wraps and having a one long page for printing. I understand this is not the way the library works. 

Looking into zariweyo's solution I was thinking of fixing the height  of each item row, then just counting rows. But it is messy because I will need to cut long item descriptions with ellipsis or something like that.

@zariweyo, could you share how did you managed to calculate your elements height and to setup the page size?

Regards!

zariweyo:
My solution it's awful, because I estimate a max height for each row, and then, I save the end with an offset height, which give me a white space, but the ticket is legible, so is a valid for now.

I tried to calculate the real height for each row, but this is very difficult, because some words may no wrap correctly in the area. 

At now I can't spend time in more efficient solution, I need to finish this app and the tickets length are acceptable. I put this feature because I think this is a problem which affect other developers, like you @fulvous , and other libraries to render PDF solve.

stone89son:
same problem

PR #1914, opened 2022-07-04
Add optional dimension callback 

wscullen:
- allow dimensions of nodes to be determined by rendering PDF more than once.
- callback happens during debug phase to match node dimensions that would be output to the user after the PDF finishes rendering with debug enabled
- dev can render a PDF more than once to determine height and width of dynamic components and use that information to inform the final layout of the PDF

### DIFF
--- a/packages/render/src/primitives/renderDebug.js
+++ b/packages/render/src/primitives/renderDebug.js
@@ -176,6 +176,20 @@ const debugOrigin = (ctx, node) => {
 };
 
 const renderDebug = (ctx, node) => {
+  if (node.props?.dimensionCallback) {
+    const { width, height } = node.box;
+    const {
+      marginLeft = 0,
+      marginTop = 0,
+      marginRight = 0,
+      marginBottom = 0,
+    } = getMargin(node.box);
+
+    const roundedWidth = Math.round(width + marginLeft + marginRight);
+    const roundedHeight = Math.round(height + marginTop + marginBottom);
+    node.props.dimensionCallback(roundedHeight, roundedWidth);
+  }
+
   if (!node.props?.debug) return;
 
   ctx.save();


diegomura:
Thanks @wscullen ! I like the idea but this isn't the best way to do it.

I'm a bit confused about the use case thought. If you want to set up nodes dimensions why not just using CSS directly?

The render phase shouldn't have any side effects. This should happen in the layout of the document. Don't see the relation to debug so it justifies being there either. 

Mind expanding more in the idea to see how would be the best way to meet the use case?

wscullen:
Thanks for looking at this so promptly! This idea is half baked but I think it could be viable. It's related to this issue:
https://github.com/diegomura/react-pdf/issues/1024

Here's a couple of use cases:

- When you set a fixed height or a fixed width, and you don't know how tall or wide the resulting node will be based on dynamic content (in our case text). So we have a set of nodes, but don't know if those nodes will take up one page or two pages, and we always want to have two pages (for double sided printing). So if we can determine the height of the nodes we know if we need to add an extra page break or not. 
- Another use case is determining how much space is left on a page, which would allow us to alter the layout (use a more compact layout, such as with 2 columns instead of one) that is guaranteed to fit the remaining space on the page. With the remaining space known we can make use of flex-columns that create new columns dynamically depending on how much text content there is by setting a fixed vertical height that perfectly fills that remaining space.

So basically it's about having a certain amount of dynamic content that you want to work around, given limitations surrounding page numbers or how much space you have left for things to be rendered on a page. I do realize that this is defeating the purpose of the pdf layout engine and how it flows things and creates pages as needed. But imagine if you had to limit yourself to a certain number of pages and/or the text content you have is variable within those space constraints. 

I agree it shouldn't be in the debug section, I added it there because I knew the debug portion had access to final node dimensions (for display purposes) so I quickly threw it in there. If there is a better place to access node dimensions through a callback mechanism during the layout phase, please point me in that direction. 

In addition, I'm currently doing 2 separate renders (one with display: none), when the first one finishes (with the callbacks added to the relevant elements), I feed that into the second render to change how that second render is laid out. It would make sense for the 1st render to instead just be a pre-render or headless render that doesn't output the final pdf data but maybe stops after the layout phase (I'm not sure how your library works so I'm probably wrong on that). In this way the pre-render would not be as much of a performance hit as doing 2 full renders of essentially the same document.

Hope that makes sense. I'm happy to answer more questions, and I really appreciate you looking at this. Thanks!

wscullen:
Closing as I don't think this will be the accepted approach.

Issue #1915, opened 2022-07-05
Primitives is not defined with parcel 2 build command output


andrewvmail:
Using  @react-pdf/renderer 2.3.0" and latest parcel 2, production build, builds fine but opening the app you will see 
<img width="478" alt="image" src="https://user-images.githubusercontent.com/2431354/177281424-dfe521da-e279-4bf1-854b-14310d9b1b94.png">

The reference error leads to primitives in the source map.
 
Minimal reproduction 
https://github.com/andrewvmail/parcel-react-pdf-bug-demo

```
git clone https://github.com/andrewvmail/parcel-react-pdf-bug-demo
cd parcel-react-pdf-bug-demo
yarn
parcel build src/index.html
cd dist
npx http-server 
# open localhost:8080

# another terminal
parcel serve src/index.html
# no console error
```

Thanks in advance!

carlobeltrame:
Sounds like a duplicate of #1317. The issues with ESM compatibility are deeper than just the default exports. @andrewvmail please join us at #1891 where we attempt to fix the browser and ESM compatibility, and discuss how the library will further develop in order to become isomorphic. It would help out greatly if you could test #1891 using parcel!

andrewvmail:
> I can not reproduce the problem, it's working!
> 
> Which version of parcel do you use?
> 
> <img alt="Снимок экрана 2022-07-06 в 12 02 37" width="1512" src="https://user-images.githubusercontent.com/6726016/177513139-f2c2ffaa-d13a-4bf7-930c-cedc56bc72f9.png">

parcel 2 latest as of yesterday

> I can not reproduce the problem, it's working!
> 
> Which version of parcel do you use?
> 
> <img alt="Снимок экрана 2022-07-06 в 12 02 37" width="1512" src="https://user-images.githubusercontent.com/6726016/177513139-f2c2ffaa-d13a-4bf7-930c-cedc56bc72f9.png">

I'm using parcel 2.5 globally installed, did you do parcel build or parcel serve? parcel serve works fine.
Have to clean it again and retry if you used parcel serve once.
```
rm -rf .parcel-cache
rm -rf dist
parcel build src/index.html
cd dist; npx http-server 
```

andrewvmail:
Just tried again using parcel 2.6.2. Parcel build still produces error. Identical problem to this one too https://github.com/diegomura/react-pdf/issues/1786

tarabishy2020:
"parcel": "^2.6.2",
[...]

PR #1920, opened 2022-07-06
fix: add default export, parcel build no error

andrewvmail:
https://github.com/diegomura/react-pdf/issues/1915


andrewvmail:
parcel 2 optimization expects default export

tarabishy2020:
Getting the same error this pull request fixes! Would be great if merged.

bgottschalk-cb:
Why was this PR not considered? I'm still having the same issue on the latest Parcel and @react-pdf/renderer versions

PR #1891, opened 2022-06-17
feat: compatibility with modern web bundlers and browsers

carlobeltrame:
As discussed in #1317, react-pdf currently cannot be used out of the box with modern ES bundlers such as Vite or Rollup (which react-pdf even uses itself for building the various packages). The reason was that react-pdf contains code which references node.js builtin modules such as `Buffer`, `stream`, `zlib` etc., which are not natively available in browsers.

This PR creates new browser-specific builds for all the @react-pdf packages that need it. This way, I was able to leave the node.js-specific builds almost untouched. After this PR, react-pdf can be used in an app without having to add any bundler config (such as described [here](https://react-pdf.org/advanced#webpack-config) for Webpack 4 and 5) or bundler shims (such as [this one](https://www.npmjs.com/package/vite-plugin-shim-react-pdf) for Vite). You can simply install @react-pdf/renderer and start using it right away, the way it should be.

Fixes #1317, fixes #1645, fixes #1899, fixes #1847, fixes #1886, fixes #1744, fixes #1670, fixes #1669, fixes #1915, fixes #1786

Screenshot from 2022-06-17 09-46-22

To acheive this, I had to bundle some commonjs dependencies with the browser versions of react-pdf:
- `blob-stream`
- `restructure`
- `dfa` (was already [bundled](https://github.com/diegomura/react-pdf/commit/3217a892e92ff98e92b6c7ea6e3244d403f679b6) before)
- `vite-compatible-readable-stream`
- `browserify-zlib`

This was necessary because these packages are in commonjs format (instead of ES6 module format) and will be misinterpreted when using a modern ES6 module bundler.

(Fun fact, I actually did the whole thing twice, but the first time around I learned so much about rollup that I decided it was worth doing a second time because the resulting rollup configs would be much cleaner :laughing:)

While I was at it, I also updated the restructure package to the latest version, which has been requested in #1451 for a long time now.

TODO:
- [x] Test in Vite (done by @carlobeltrame)
- [ ] Someone else test in Vite, development (@thekevinbrown)
- [ ] Someone else test in Vite, production build (@thekevinbrown)
- [ ] Test on Node.js (@thekevinbrown)
- [ ] Test in Webpack 4 (@diegomura or @jeetiss ?)
- [ ] Test in Webpack 5 (@jeetiss ?)
- [ ] If it works without additional config in Webpack 4 / 5, remove the corresponding [documentation](https://github.com/diegomura/react-pdf-site/blob/master/docs/webpack.md)
[...]

PR #1934, opened 2022-07-11
Restore the ability to load fonts by relative URL in browsers

carlobeltrame:
This was broken in #1927, when trying to optimize for bundle size.
See https://github.com/diegomura/react-pdf/pull/1891#issuecomment-1180200312 for an explanation.

### DIFF
--- a/packages/font/src/font.js
+++ b/packages/font/src/font.js
@@ -61,7 +61,7 @@ class FontSource {
     if (isDataUrl(this.src)) {
       const raw = this.src.split(',')[1];
       this.data = fontkit.create(toUint8Array(raw), postscriptName);
-    } else if (isUrl(this.src)) {
+    } else if (BROWSER || isUrl(this.src)) {
       const { headers, body, method = 'GET' } = this.options;
       const data = await fetchFont(this.src, { method, body, headers });
       this.data = fontkit.create(data, postscriptName);


diegomura:
Thanks @carlobeltrame and @CodeWitchBella for finding this and your comments in #1891 

I didn't realize this will break that way. However, this means we cannot load fonts from remote urls in Node right? That's what I tried to support

diegomura (approved):

carlobeltrame:
@diegomura done :+1:

.changeset/violet-nails-accept.md
```diff
@@ -0,0 +1,5 @@
+---
+'@react-pdf/font': patch
+---
+
+Restore the ability to load fonts by relative URL in browsers
```
diegomura, line 5:
```suggestion
fix: ability to load fonts by relative URL in browsers
```

diegomura:
Thanks man. You're great. I appreciate all the work you put on this project.

carlobeltrame:
Ah, yes, sorry I amnot used to conventional commits. Will try to improve that in my contributions.

diegomura:
No problem 😄 We are not following strictly now, but I find helpful at least flagging `feat` `chore` and `fix` for future reference

Issue #1805, opened 2022-05-03
Image caching causes wrong images to be shown for Buffer source


wyozi:
**Describe the bug**
Cache key creation (https://github.com/diegomura/react-pdf/blob/6ea172340b0e89b01a985e91eee257c43b243c54/packages/image/src/resolve.js#L160) is invalid for Buffer sources; a node.js Buffer has neither `.data` nor `.uri`, so cacheKey is always undefined and thus the first cached image will be reused for subsequent pdf renders.

**To Reproduce**
1. Pass a Buffer to `<Image src={buf}/>`

**Expected behavior**
A proper cacheKey is generated for buffers so that images are cached properly (or alternatively caching is disabled for buffers)

**Desktop (please complete the following information):**
 - React-pdf version: main

KevLehman:
⚠️ This is not a recommended approach

You can extend `Buffer` to include a `uri` property with a random value. That will ensure cache won't hit as the keys would be different.

However, this package needs to provide a way to disable caching, or at least configure. Defaulting to 30 elements stored can be a bit problematic, specially in memory

KevLehman:
A workaround is to pass `cache={false}` to the Image component when rendering. This will cause that you cannot recycle images, but if you, like in my case, were using totally different images, then that should work.

Would be good also to have a good caching mechanism (apart from a proper caching key). 
Defaulting the elements in cache to 30 and not cleaning them up is kinda troublesome in node envs with low memory.

emmafielduk:
Hello, I am having caching issues with images, where the image at src has been updated.
A logo is repeated multiple times on every page.

<Image src={src} />
<Image src={src} cache={false} />

Both options return the old image, but if I add a random parameter to the url it works.
It's been a couple weeks, how long does this cache last? 

Thanks Diego :)

AgostinhoMayoral:
I had the same problem and managed to solve it definitively by configuring cache control, pragma and expirations in the headers.

<Image
          style={styles.image}
          src={{
            uri: data.logoImagePath,
            method: "GET",
            body: "",
            headers: {
              "Cache-Control": "no-cache, no-store, must-revalidate",
              Pragma: "no-cache",
              Expires: 0,
            },
          }}
          cache={false}
/>

PR #1939, opened 2022-07-15
Hotfix: issue #1805 by not using the cache for images from a buffer

13Bytes:
Hi,
this is a Hotfix for #1805.
It will disable caching for images from a buffer, until someone comes up with a better solution.

### DIFF
--- a/packages/image/src/resolve.js
+++ b/packages/image/src/resolve.js
@@ -157,17 +157,15 @@ const resolveImageFromUrl = async src => {
 };
 
 const resolveImage = (src, { cache = true } = {}) => {
+  let image;
   const cacheKey = src.data ? src.data.toString() : src.uri;
 
-  if (cache && IMAGE_CACHE.get(cacheKey)) {
-    return IMAGE_CACHE.get(cacheKey);
-  }
-
-  let image;
-  if (isCompatibleBase64(src)) {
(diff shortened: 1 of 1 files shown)


13Bytes:
Hi,
it would be great if this (or a different solution to this bug) could be released, as it currently blocks a project I'm working on : /

greenking19:
hi, I use the buffer image in the browser, and the displayed images are all the same. Will this patch fix it?

13Bytes:
@greenking19 Yes, it does

greenking19:
@13Bytes I pulled this pr, but I don't know how to use it. How can I use this? I don't know how to use it after using 'npm run build', unlike a normal package😯

13Bytes:
> @13Bytes I pulled this pr, but I don't know how to use it. How can I use this? I don't know how to use it after using 'npm run build', unlike a normal package😯

Since this is just the source-code, you need to set up the repo as described here: https://github.com/diegomura/react-pdf/blob/master/.github/CONTRIBUTING.md#set-up-your-development-environment
After that, you can build (this branch) and use the artifacts.

greenking19:
@13Bytes  It looks like a lot of trouble. I'm now replacing it with base64. Hopefully this pr will merge soon

Issue #1965, opened 2022-08-05
SVG component types missing children prop


ghost:
**Describe the bug**
When try to use with Typescript setup for SVG component gives following error.

```
No overload matches this call.
  Overload 1 of 2, '(props: SVGProps | Readonly<SVGProps>): Svg', gave the following error.
    Type '{ children: Element; viewBox: string; style: Style; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<Svg> & Readonly<SVGProps>'.
      Property 'children' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<Svg> & Readonly<SVGProps>'.
```

**To Reproduce**
Steps to reproduce the behavior including code snippet (if applies):
1. Create Typscript project.
2. Try to create a Document with SVG component.

**Expected behavior**
SVG component should be able to have props.

**Desktop (please complete the following information):**
 - OS: [MacOS]
 - React-pdf version: "2.3.0",

RingoDev:
The same also holds true for the G component.
A possible solution would be to add the prop "children?: React.ReactNode" somewhere in the type definition similar to [#1811](https://github.com/diegomura/react-pdf/issues/1811)

RingoDev:
opened a PR which should fix this problem: #1975

dimitur2204:
Same here

MBeggiato:
Still open as of "@react-pdf/renderer": "^3.0.2" right?

dimitur2204:
> Still open as of "@react-pdf/renderer": "^3.0.2" right?

@MBeggiato  At least the issue from this branch https://github.com/diegomura/react-pdf/pull/2087 is still not fixed

bogdan-calapod:
I suppose this is due to a React 18 requirement, right ? 

Is there any workaround available other than downgrading back to 17 ?

dimitur2204:
> I suppose this is due to a React 18 requirement, right ?
Probably!

> Is there any workaround available other than downgrading back to 17 ?
You can do `declare module '@react-pdf/renderer'` which will ignore all types and you won't have typings

diegomura:
Fixed

PR #1798, opened 2022-04-24
Fix types for @types/react@18

karlhorky:
Ref: https://github.com/diegomura/react-pdf/issues/1062#issuecomment-1092100093
Ref: https://github.com/eps1lon/types-react-codemod#implicit-children
Ref: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/46691

Found some time to take a look at the breakages with `@types/react@18`, and opened a PR to fix the types in the @react-pdf/renderer package: https://github.com/diegomura/react-pdf/pull/1798

However, this does not fix the other packages in the monorepo. And this may cause breakage for @types/react versions lower than 18.

cc @diegomura @jeetiss

Copying my comment here for visibility:

> Hm, but it fails with [`@types/react` v18](https://www.npmjs.com/package/@types/react), which was just published (because of [the breaking changes](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/56210)):
> 
> ```
> Run yarn tsc
> yarn run v1.22.18
> $ /home/runner/work/project/node_modules/.bin/tsc
> Error: project/generatePdf.tsx(212,6): error TS2769: No overload matches this call.
>   Overload 1 of 2, '(props: DocumentProps | Readonly<DocumentProps>): Document', gave the following error.
>     Type '{ children: Element[]; }' has no properties in common with type 'IntrinsicAttributes & IntrinsicClassAttributes<Document> & Readonly<DocumentProps>'.
>   Overload 2 of 2, '(props: DocumentProps, context: any): Document', gave the following error.
>     Type '{ children: Element[]; }' has no properties in common with type 'IntrinsicAttributes & IntrinsicClassAttributes<Document> & Readonly<DocumentProps>'.
> Error: project/generatePdf.tsx(213,8): error TS2769: No overload matches this call.
>   Overload 1 of 2, '(props: PageProps | Readonly<PageProps>): Page', gave the following error.
>     Type '{ children: Element; size: "A4"; style: { fontFamily: string; fontSize: number; lineHeight: number; paddingTop: number; paddingBottom: number; paddingHorizontal: number; }; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<Page> & Readonly<PageProps>'.
>       Property 'children' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<Page> & Readonly<PageProps>'.
>   Overload 2 of 2, '(props: PageProps, context: any): Page', gave the following error.
[...]

PR #1975, opened 2022-08-18
fix: add children prop to G and Svg components

RingoDev:
fixes the issue described in #1965 where the Svg component and the G component do not have a children prop declared


RingoDev:
This issue would probably also fixed with #1798 as far as I can tell

PR #2034, opened 2022-10-04
fix: page wrap=false not preventing page from breaking

imatwork:
This fixes `<Page wrap={false}>` not actually preventing a page from breaking. It also can save large amounts of time (I noticed a 50% performance increase) on pages with many elements.

### DIFF
--- a/packages/layout/src/steps/resolvePagination.js
+++ b/packages/layout/src/steps/resolvePagination.js
@@ -260,6 +260,11 @@ const resolvePagination = (doc, fontStore) => {
 
   for (let i = 0; i < doc.children.length; i += 1) {
     const page = doc.children[i];
+    if (!page.props?.wrap) {
+      pages.push(page);
+      continue;
+    }
+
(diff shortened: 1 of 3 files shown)


chathu-novade:
@diegomura Can we get this merged and release if this looks good?

krayste:
Yes pls, I have the same performance issue and this fixes it 💪

timbrunette:
please merge

packages/layout/src/steps/resolvePagination.js
```diff
@@ -260,6 +260,11 @@ const resolvePagination = (doc, fontStore) => {
 
   for (let i = 0; i < doc.children.length; i += 1) {
     const page = doc.children[i];
+    if (!page.props?.wrap) {
```
diegomura, line 263: We should explicitely check for false here. Otherwise missing `wrap` prop would prevent pages from wrapping which isn't what we want

PR #2056, opened 2022-10-15
fix: TypeError when returning fragments or arrays from render prop

### LINKED ISSUE #1872: Error when returning an array on render prop
**Describe the bug**

When I try something like:
```jsx
<View
  render={({ pageNumber }) => (
    items.map(item => <Text>{item.label}, page: {pageNumber}</Text>)
  )}
/>
```

React Pdf throws an exception

**To Reproduce**
1. Create a document with react PDF
2. Pass the render prop and return an array from there
3. It throwa

Codesanbox showing the errro: https://codesandbox.io/s/objective-greider-i5p544

**Expected behavior**
It should work normally as if the array was passed as children.

**Screenshots**
image

 - OS: Windows and Linux
 - Browser Chrome
 - React-pdf version v2.1.2

jasnross:
This change fixes #1872 by supporting fragments and arrays as valid nodes to be returned from `render`. It also adds support for nested dynamic components.

### DIFF
--- a/packages/layout/src/node/createInstances.js
+++ b/packages/layout/src/node/createInstances.js
@@ -9,32 +9,30 @@ const isFragment = value =>
   value && value.type === Symbol.for('react.fragment');
 
 /**
- * Transforms a react element instance to internal element format
+ * Transforms a react element instance to internal element format.
+ *
+ * Can return multiple instances in the case of arrays or fragments.
  *
  * @param {Object} React element
- * @returns {Object} parsed react element
+ * @returns {Array} parsed react elements
  */
-const createInstance = element => {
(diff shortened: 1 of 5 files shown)


packages/layout/src/node/createInstances.js
```diff
@@ -0,0 +1,59 @@
+import { castArray } from '@react-pdf/fns';
```
jasnross, line 1: It looks like a new file in the diff here, but this is actually just an edit/rename of `createInstance.js`. Wasn't sure about renaming it but ultimately I thought it felt better with a new name since it now always returns an array of created element instances.

PR #2059, opened 2022-10-18
feat: Added DocProvider to access PDFDocument directly during render

imatwork:
Adds a `<DocProvider fn={...}/>` primitive that gives you access to underlying `pdfkit` `PDFDocument`. This means you can use both `@react-pdf/renderer` (doing things in React/JSX) while also being able to use `pdfkit` functionality when needed.

This PR includes:
* The new functionality
* Some tests for `packages/primitives` and `packages/render`
* A new example called `docProvider` to demonstrate usage

### Example usage:

```jsx
<Document
  title="Table"
  author="Jane Doe"
  subject="Rendering with react-pdf and modifying document during render"
  keywords={['react', 'pdf', 'table']}
>
  <Page size="A4">
    <Text>This is the first page</Text>
  </Page>
  <DocProvider fn={doc => {
    doc.addPage({ size: "A4" });
    doc.text('This is the middle page, added using pdfkit!');
  }} />
  <Page size="A4">
    <Text>This is the last page</Text>
  </Page>
</Document>
```

### Thoughts:

I would also be nice to go the other way, allow `pdfkit` to pass a `PDFDocument` instance to `react-pdf/render` to render out nodes created using React/JSX. The problem is React/JSX nodes need to be rendered by the React Fiber Reconciler and also need to go through `react-pdf/layout` (which currently is only setup to be passed a document, not pages, or nodes individually), so the task was larger than I was willing to take on currently.

### DIFF
--- a/packages/examples/src/docProvider/index.js
+++ b/packages/examples/src/docProvider/index.js
@@ -1,0 +1,24 @@
+import React from 'react';
+import { Page, Document, Text, DocProvider } from '@react-pdf/renderer';
+import PDFDocument from '@react-pdf/pdfkit';
+import { renderNode } from '@react-pdf/render';
+
+export default () => (
+  <Document
+    title="Table"
+    author="Jane Doe"
+    subject="Rendering with react-pdf and modifying document during render"
+    keywords={['react', 'pdf', 'table']}
+  >
+    <Page size="A4">
+      <Text>This is the first page</Text>
+    </Page>
+    <DocProvider fn={doc => {
+      doc.addPage({ size: "A4" });
+      doc.text('This is the middle page, added using pdfkit!');
+    }} />
+    <Page size="A4">
+      <Text>This is the last page</Text>
+    </Page>
+  </Document>
+);
--- a/packages/examples/src/index.js
+++ b/packages/examples/src/index.js
@@ -11,6 +11,7 @@ import Knobs from './knobs';
 import Resume from './resume';
 import Fractals from './fractals';
 import PageWrap from './pageWrap';
+import DocProvider from './docProvider';
 
 const MOUNT_ELEMENT = document.getElementById('root');
 
@@ -22,6 +23,7 @@ const EXAMPLES = {
   resume: Resume,
   pageWrap: PageWrap,
   fractals: Fractals,
+  docProvider: DocProvider,
 };
 
 const Viewer = () => {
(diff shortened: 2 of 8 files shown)


diegomura:
Thanks @imatwork ! Couple of questions/concerns:

- How is this different than Canvas? I know canvas just exposes a subset of pdfkit api
- I don't think it's a good idea to provide access to pdfkit doc tbh. This is an internal instance. Exposing it would make it part of the public API, and any change there might break any react-pdf usage. It's hard enough to keep public apis consistent, to also do it on internal concepts
- Anything should be achievable through a declarative API. What is it specifically you want to achieve with this provider? Maybe we can fix it without exposing pdfkit instance

imatwork:
> How is this different than Canvas? I know canvas just exposes a subset of pdfkit api

Interesting, I did not delve deeply into Canvas, but it looks like it has the API surface area I would need from a skim. I just needed `text`, `font`, and line related stuff.

> I don't think it's a good idea to provide access to pdfkit doc tbh. This is an internal instance. Exposing it would make it part of the public API, and any change there might break any react-pdf usage. It's hard enough to keep public apis consistent, to also do it on internal concepts

That's fair, it would probably require more maintainer overhead to explicitly track bundled/compatible versions of pdfkit and publish those.

> Anything should be achievable through a declarative API. What is it specifically you want to achieve with this provider? Maybe we can fix it without exposing pdfkit instance

Well, originally, I needed to render a table with many elements but found it very slow in `react-pdf` but after running a test against `pdfkit` alone rendering the same thing, `pdfkit` alone was much faster. I started looking at optimizations in `react-pdf` (how #2034 happened, and another one that got cut short) but found it a more "path of least resistance" to just use pdfkit directly (decision after discussion with my client/employer due to time investment).

I figured it might be nice to have the flexibility to jump in and out of imperatively coding a page when you needed to, hence the example which makes a document that has a mixture of declarative and imperative. It was more a side effect that I could then use a `pdfkit`/imperatively coded page in my current project for performance reasons.

mohadib:
Why would you not want to expose the pdfkit instance? This library is wrapper that limits the functionality but provides a nice api, it makes sense that you would allow the user an escape hatch to get at pdfkit for cases when the react api fails. An example: measuring text.

diegomura:
This needs rebase

@mohadib I explained some reasons above. While I agree this is a wrapper, I don't see this lib in the same way as you explained. The fact we use pdkit internally is an implementation detail. Originally this library produced the PDF document manually, in the future we might change to another solution. Exposing this increases the public api of react-pdf by a lot. 

As said above, new features should be added as part of the react-pdf api, not adding a backdoor to access internal instances. If we need an api to measure text let's add it to `<Text />`

Also, having a new primitive just for this does not look very good. It would be better as a prop of `doc`. I see how it's not the same. With a primitive you control when it runs based on elements order, but itself is very hacky. Will close this for the moment for bookeeping but please let's keep the discussion

PR #2071, opened 2022-11-04
feat: enable `PDFDownloadLink` to Receive Anchor Props

ITenthusiasm:
## Motivation
There are a few reasons why someone might want to leverage all of the attributes available on the related anchor (`a`) element for `PDFDownloadLink`. This PR makes this possible without ruining backwards compatibility (`download` is now legal; `fileName` is still preserved; and `fileName` will yield to `download` if the attribute is explicitly added). The TS props were also updated for clarity.

It might be useful to have this on the v2 and v3 versions? Since it seems like some people in the `react@18` world are having a hard time? But I leave that to you guys.

I considered adding `React.forwardRef` too, but I didn't want to dive through additional TS headaches. Might be worth considering later?

## Changes
- Pass ALL anchor tag props to the rendered anchor element -- EXCLUDING the reserved `href` attribute.
- Improve the `onClick` type of `PDFDownloadLink`.
- Minor cleanup in related files.


packages/renderer/src/dom/PDFDownloadLink.js
```diff
@@ -5,18 +5,17 @@ import React, { useEffect } from 'react';
 import usePDF from './usePDF';
 
 export const PDFDownloadLink = ({
-  style,
-  children,
-  className,
-  document: doc,
   fileName = 'document.pdf',
+  document,
```
diegomura, line 9: Let's keep alias so we avoid issues with global document value

```suggestion
  document: doc,
```

packages/renderer/src/dom/PDFDownloadLink.js
```diff
@@ -5,18 +5,17 @@ import React, { useEffect } from 'react';
...
+  children,
   onClick,
+  href: _filteredOutHref,
+  ...rest
 }) => {
-  const [instance, updateInstance] = usePDF({ document: doc });
-
+  const [instance, updateInstance] = usePDF({ document });
```
diegomura, line 15:
```suggestion
  const [instance, updateInstance] = usePDF({ document: doc });
```

packages/renderer/src/dom/PDFDownloadLink.js
```diff
@@ -5,18 +5,17 @@ import React, { useEffect } from 'react';
...
   useEffect(updateInstance, [children]);
 
-  if (!doc) {
+  if (!document) {
```
diegomura, line 18:
```suggestion
  if (!doc) {
```

packages/renderer/src/dom/PDFDownloadLink.js
```diff
@@ -5,15 +5,14 @@ import React, { useEffect } from 'react';
...
-  className,
-  document: doc,
   fileName = 'document.pdf',
+  document: doc,
+  children,
   onClick,
+  href: _filteredOutHref,
+  ...rest
 }) => {
   const [instance, updateInstance] = usePDF({ document: doc });
```
diegomura, line 15:
```suggestion
  const [instance, updateInstance] = usePDF({ document: doc });

```

diegomura (approved):
utACK

PR #2079, opened 2022-11-10
chore: bump react@^17.0.2, react-dom@^17.0.1, react-reconciler@0.26.2

modosc:
i started looking into the `react-reconciler` issues with `react@18`. 

as a first step i found `react-reconciler@0.26.2` was the last version that works with the current codebase. this has a peer dep of `react@^17.0.2` so i also upgraded `react` and `react-dom`. 

because this drops support for `react@16` it probably warrants a major bump?


jeetiss:
closed in favor #2140

PR #2087, opened 2022-11-19
Update `G` and `Svg` Props to include children

dimitur2204:
Addresses #1965


kachar:
It's gonna be great to see this PR included in the next versions. :pray: 
We're trying to upgrade to Next 13 and React 18 and this seems to be a temporary blocker.

MarkLinovy:
LinearGradient has no children either. might want to address that in this PR aswell.
It needs children for <Stop/>.

Issue #1522, opened 2021-09-27
Add `gap` property to StyleSheet API


JuanM04:
**Is your feature request related to a problem? Please describe.**
When generating dynamic content (like tables), [`gap`](https://developer.mozilla.org/en-US/docs/Web/CSS/gap) provides a simple way of creating spaces between elements inside a flexbox without necessity of doing weird things with margins and widths.

**Describe the solution you'd like**
Maybe something like `gap`, `gapVertical` and `gapHorizontal`.

jeetiss:
`@react-pdf` uses yoga layout under hood and should be implemented in yoga first

Upvote this issue if you agree. https://github.com/facebook/yoga/issues/812

process0:
There seems to finally be activity on the PR that implements gap for yoga https://github.com/facebook/yoga/pull/1116

axel7083:
Since the PR is still pending, here is a solution I found which work fine: [flex-gap-in-unsupported-browsers](https://nosleepjavascript.com/flex-gap-in-unsupported-browsers/)

ajayvikas:
Seems it has been implemented in yoga.

Sheharyar566:
Doesn't seem to be working for me

RNm-dove:
In my case, `gap` not works, but `columnGap` works.

Issue #510, opened 2019-02-26
Can't use with Content Security Policy


ekilah:
**Describe the bug**

```
EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive:
```

Our site has a Content Security Policy preventing `eval()` from being used. Several dependencies of this project use `eval()`, here are PRs where people have tried to fix this there unsuccessfully:
`fontkit`: https://github.com/foliojs/fontkit/pull/180
`restructure`: https://github.com/foliojs/restructure/pull/29
`yoga-layout`: (see below)

One guy has forked those two ( https://github.com/Hopding/fontkit and https://github.com/Hopding/restructure , @Hopding ) to fix these issues because it seems like the maintainer of the originals is somewhat unresponsive to this issue. 

I was able to work around this by replacing this repo's dependencies with @Hopding's forks, for instance:

```json
"resolutions": {
    "restructure": "https://registry.yarnpkg.com/@pdf-lib/restructure/-/restructure-0.0.1.tgz#e38c81b6ddaa8c982e7780620c89bddfb6c53653"
  }
```

the above is the result of a manual process to link to the fork's product because of a bug in `yarn` (see https://github.com/yarnpkg/yarn/issues/5235 ), something like `"restructure": "https://github.com/Hopding/restructure#8af8c49f3a63681d90554be1600d1b381ff9114d"` should work in theory. 

However, it looks like after doing that, there is another dependency here that uses `eval`, and this time it's `yoga-layout-prebuilt` (a wrapper around `yoga-layout`). I can't seem to find any forks or issues on either of those repos that attempt to remove the `eval` there. Seems like a large repo, so I've hit a bit of a wall towards solving this myself.

error message for reference: 

<img width="1037" alt="screen shot 2019-02-25 at 5 55 12 pm" src="https://user-images.githubusercontent.com/3213339/53381748-9aca4b00-3926-11e9-917b-9f4a037099cb.png">

if I add `unsafe-eval` in my CSP, everything works.

Note: I know these are issues in dependencies, and not in this repo itself, but:

* Ideally this repo wouldn't have dependencies that cause issues like this for a relatively common CSP configuration
* This repo could potentially use forks of the dependencies that fix these issues
* I'm not sure where else to ask about / point out these issues, since they are somewhat widespread

**To Reproduce**
Have a CSP that is restrictive of `script-src`, i.e.:

```
[...]

PR #2105, opened 2022-12-10
upgrade yoga-layout to latest version 

jeetiss:
### yoga-layout built by `emscripten` with `flex-gap` and without `eval`

I create PRs for all maintained yoga repositories but it is stuck for a while
- https://github.com/facebook/yoga/pull/1177
- https://github.com/shuding/yoga-wasm-web/pull/6

So I created my own fork, hope it can be replaced in future

- https://github.com/jeetiss/yet-another-yoga-layout-fork

it has a synchronous `asm` build and we can replace `yoga-layout-prebuilt` with no pain, but it is better to replace this with `wasm` asynchronous version that is two times faster and smaller in the future.

related to #1522 and #510

### TODOs
- [x] yoga-layout has updated default position type `Yoga.POSITION_TYPE_STATIC` need to check that it is not breaking the CSS position property.
- [ ] remove hacks with `setFlexBasisAuto`


diegomura:
@jeetiss I'll try to test it asap. Thanks for focusing on this. When we used the main package, there were several issues on the build step of yoga. Using a pre-built version is nice as it doesn't involve building yoga on each user machine. Do you think it's possible to keep that? Also (and not important) if this is a fork to use in react-pdf wonder if it isn't better to have it `@react-pdf/yoga` directly

diegomura:
@jeetiss this looks good! Can't test atm, but happy if you take the lead here 😄  This does not use your `yet-another-yoga-layout-fork` fork anymore right?

jeetiss:
> This does not use your yet-another-yoga-layout-fork fork anymore right?

Yep! but it still has a build step to provide a `commonjs` version

jhgeluk:
@jeetiss Does this PR fix the current memory leak issue?

PR #2150, opened 2023-01-24
Update the peerDependencies of react

umairgull5000:


jeetiss:
@umairgull5000 thanks for the PR, but this is not a proper solution because react-reconciler has a react as peer dependency too.

PR #2158, opened 2023-01-28
Add type for renderToBuffer method

jeetiss:


Smirow:
That does not solve the import issue of `renderToBuffer` still not correctly exported.

Issue #1590, opened 2021-11-10
Any good tricks for superscripting text characters?


oktalk:
Commonly, text characters will need to be superscript. In the case of disclaimers, or symbols like the `®` registration mark. You can accomplish this with the HTML tag `<sub>`, or via CSS like so `verticalAlign: 'super'`. Neither of these are supported in React-PDF. Further reading on how this is accomplished in HTML: https://www.w3schools.com/tags/tag_sup.asp

I would like more support for CSS rules to affect superscripting text. 
Here is a simple Fiddle with examples: https://jsfiddle.net/b6rq5sLc/

Superscripting, or approximating it, doesn't seem to be supported in any way.

jeetiss:
hi, @oktalk 

[transform should](https://react-pdf.org/repl?code=3187b0760ce02e00408a057025803c450298c0bc300500943807cf805030c00f0022230080b6198531165d400a02180e659a004f003619b006f61622003a0046200099080beed3a72a005430a360194100070c009c23033498ec6d002c9041d7b634f152a288c72209f3ea00755400f4bafac4a1e16c1c94a1bc021ad421740cccacec0400dc6464a0906e5e32383006451806f6181850b256183c9878121c8a2a005c302d9ac63c4a4a48607cda20c69d00cc00ac0034b130bdfd837c0042e850208c9d006c33730b0343001220d6005ee0503c2213bb94aab3947ea6669ddd9c50663c900066279b3000720f97c202246860009a78002d24d8c280200238aa322a9b2b9001283580502e0d00062b2330b094e63c15110a87416042c4745000) work with text, looks like a bug

oktalk:
This is not ideal, but I was able to fake superscript in the below code. I'm guessing what it comes down to is that the `Text` component can't have any positioning or block element styles.

`        <View
          style={{
            display: 'flex',
            flexDirection: 'row',
            flexWrap: 'wrap',
            position: 'relative',
          }}
        >
          <View>
            <Text style={{ color: '#4d4e53', fontSize: '24pt' }}>This thing is a Registered Trademark</Text>
          </View>
          <View>
            <Text style={{ fontSize: '14pt', color: '#4d4e53' }}>® </Text>
          </View>
          <View>
            <Text style={{ color: '#4d4e53', fontSize: '24pt' }}>some more text.</Text>
          </View>
        </View>`

oktalk:
If anybody is interested in how I eventually got around this issue, here's my solution:

[...]

PR #2214, opened 2023-02-24
feat: support for verticalAlign "super" and "sub"

KODIKAS-NL:
As noted in this issue : https://github.com/diegomura/react-pdf/issues/1590 there is currently no support for super- and subscript which I too needed, so I implemented it.
It basically just scales the fontSize to 75%, scales the letterspacing and adjusts the yOffset, I haven't tested this with much fonts, but this should be a good start to finetune the implementation if needed

Here is an example with the default Helvetica font
image


jeetiss:
@KODIKAS-NL thanks for your contribution, nice idea on how to implement `vertical-align`

I think `vertical-align` should take a number and doesn't scale font, it should be fine for the first iteration.

#### superscript font will be like:

```js
<Text>Hello <Text style={{ verticalAlign: 10, fontSize: 8 }}>super</Text></Text>
```

#### subscript font will be like:

```js
<Text>Hello <Text style={{ verticalAlign: -10, fontSize: 8 }}>sub</Text></Text>
```

What do you think?

packages/textkit/src/layout/verticalAlign.js
```diff
@@ -0,0 +1,28 @@
...
+ */
+const verticalAlignment = () => attributedString => {
+  attributedString.runs.forEach(run => {
+    const { attributes } = run;
+    const { verticalAlign, fontSize } = attributes;
+
+    if (verticalAlign === 'sub') {
+      attributes.fontSize = fontSize * 0.75;
+      attributes.yOffset = -0.2;
+      attributes.characterSpacing = (fontSize - attributes.fontSize) / -2.45;
```
jeetiss, line 18: what is `-2.45`? should this parameter be gotten from a font?

KODIKAS-NL:
> @KODIKAS-NL thanks for your contribution, nice idea on how to implement `vertical-align`
> 
> I think `vertical-align` should take a number and doesn't scale font, it should be fine for the first iteration.
> 
> #### superscript font will be like:
> ```js
> <Text>Hello <Text style={{ verticalAlign: 10, fontSize: 8 }}>super</Text></Text>
> ```
> 
> #### subscript font will be like:
> ```js
> <Text>Hello <Text style={{ verticalAlign: -10, fontSize: 8 }}>sub</Text></Text>
> ```
> 
> What do you think?

Hi, thanks for you feedback!

My intention was to replicate the CSS implementation of super- and subscript which (IMHO) sort of 'misuses' the verticalAlign property (since it does more than a simple vertical alignment and also scales the font) that's why I also included the font scaling.

But I think you're right, it would be nicer to keep the 'verticalAlign' property clean and implement it as a numeric value so it can be used for general vertical alignment.

Although you have enough to replicate super- and subscript with a simple vertical alignment implementation, I would suggest though, for convenience, to still add a new property to set super/subscript which sets the vertical alignment to a pre-defined value and also autoscales the font to around 75%.

For my personal implementation such an option is very useful. In my application I am converting an existing document tree to pdf using ReactPDF, and while parsing the original tree I do not always have the current fontSize in scope since for some lines this is just automatically inherited from the parent node. So if I would want to apply superscript I would have to lookup the current fontSize further up the tree or keep passing it downwards. Having an extra property I can just set to "superscript" would be a lot easier 😊

I'll change to implementation and make verticalAlign numeric as you suggested and see if I can implement an option to easily apply super/subscipt and see what you think about it.

packages/textkit/src/layout/verticalAlign.js
KODIKAS-NL: If I just applied the new fontSize here the character spacing would still use the spacing of the previous fontSize so it has to be adjusted. I came to a factor of -2.45 by trying out different parameters and comparing results.

But this will not be needed if I change the implementation to just setting vertical alignment as a numeric value.

jeetiss:
I tested the numeric `verticalAlign` and it caused bugs with line height like that:

text-test-js-text-vertical-align-should-work-1-snap

So it seems like it is better to keep hardcoded `yOffset` for `sub`/`super` instead of a number, sorry for the inconsistency here 🙌🏻

packages/textkit/src/layout/verticalAlign.js
jeetiss: Could you please remove other font modifications and keep only hardcoded `sub`/`super`?

I agree that it is not so useful, but it will follow the spec 
https://w3c.github.io/csswg-drafts/css2/#valdef-vertical-align-sub

KODIKAS-NL: > Could you please remove other font modifications and keep only hardcoded `sub`/`super`?
> 
> I agree that it is not so useful, but it will follow the spec https://w3c.github.io/csswg-drafts/css2/#valdef-vertical-align-sub

Yes will do, I just read your other comment about making it numeric 😊

I'll build the fontSize logic in my side of the application, setting it that way also has the advantage of not having to correct the character spacing.

KODIKAS-NL:
> My intention was to replicate the CSS implementation of super- and subscript which (IMHO) sort of 'misuses' the verticalAlign property (since it does more than a simple vertical alignment and also scales the font) that's why I also included the font scaling.
> 
I was actually mistaken about this.. the css implementation does not scale the font, I confused it the MS Word implementation, which by the way, uses different scaling and position values for the online and MacOS version 😣

jeetiss:
Thanks for your contribution

KODIKAS-NL:
> Thanks for your contribution
You're welcome!

I saw you just added a snapshot test, I didn't notice they where there, Sorry, If I have any more additions I'll be sure to add it too

Issue #1369, opened 2021-06-11
Render prop method on View does not work for Images


Abhir16:
The following throws an error: Error: `TypeError: Cannot read property 'width' of undefined`

```
<View fixed render={({ pageNumber }) => {
        // I added this 
    if (pageNumber === 1) {
      return <View style={{ height: 150, backgroundColor: 'blue' }}><Text>{pageNumber}</Text></View>
    }

    return <View style={{height: 50, width: 50}}><Image
		style={styles.image}
        src="/images/quijote1.jpg"
                   /></View>
```

[...]

Issue #2064, opened 2022-10-28
Using an Image within the render function of a View causes the render process crash


pganster:
**Describe the bug**
Using `<Image />` within the `render()` function of a `View` causes the render process to crash and the following error message is shown: `Error: TypeError: Cannot read properties of undefined (reading 'width')`.

**To Reproduce**
1. Use an image within the render function: `<View render={() => <Image src="/images/quijote1.jpg" />} />`
2. The rendering crashes

[In this REPL](https://react-pdf.org/repl?code=3187b0760ce02e00408a057025803c450298c0bc300500943807cf805030c00f0022230080b6198531165d400a02180e619da74e5401a920c01dc6002716004c30cec01bd0896a01251bf2c106706c008803d121d02209808ec8015ba0c011801d1d800e7c8cc13c402f8f908889af009055099d03332b3b010037190012860f3014170d00188b9c98228c9e15222a23a04259100000) you can see the crash in action.

**Expected behavior**
The rendering should not crash and the image should be rendered. I'm using the render function, so I can render an image on every page accept on the first page.

**Desktop (please complete the following information):**
 - OS: Linux
 - Browser: No Browser, I'm generating PDFs in a NodeJS context.
 - React-pdf version: v3.0.0

ZabeehUllah:
I am facing the same issue. I am unable to render dynamic image. Have you found the solution?

pganster:
No I haven't and there are many issues regarding this and I don't hink this will be fixed. 

https://github.com/diegomura/react-pdf/issues/1936
https://github.com/diegomura/react-pdf/issues/1630
https://github.com/diegomura/react-pdf/issues/1587
https://github.com/diegomura/react-pdf/issues/1369

ZabeehUllah:
@pganster any alternative library that might work better for such scenario?

pganster:
@ZabeehUllah You can try working with [pdfjs](https://github.com/mozilla/pdfjs-dist), which react-pdf uses underneath. But I have no idea whether this use case is possible or not with that library. I ended up displaying the image on every page.

ZabeehUllah:
Okay, I will give it a try. Thank you anayway.

diegomura:
Duplicated https://github.com/diegomura/react-pdf/issues/1630

Issue #1936, opened 2022-07-13
Image under View render prop Not working


Belkacem:
**Describe the bug**
Image under View render prop we get this error message : Error: TypeError: Cannot read properties of undefined (reading 'width')

**To Reproduce**
Repl [link](https://react-pdf.org/repl?code=3187b0760ce02e00408a057025803c450298c0bc300500943807cf805030c00f0022230080b6198531165d400a02180e659a004f003619b006f61622003a0046200099080beed3a72a00d490600ee300138b25188e4c2246074d0120ed500928dfa0a8a3c548f336525702aa301046c0d8004400f4fe6e1091008ec80056e818008cb249000e7ce130911a9a30410066a8184af9859ab6d491bc02855491740cccacec0400dc64640062e050b2267c48d0e678121c253c8c482242005c3000e400f210fa3c224a2b00341c21c0cb2b0016505059108b91912583727cd03c5048c0b2a08c9171209bdb4a91001b9a400cc910026801f4002a00696d000d840000e003a88101f2002a928418c4840034009c3c0020a936417128acc8aa6eaf540905834830101c0c000ca3e0c0734e180c10d802617861261c450a996534d164784a25120c07c68480b2cb10401580e32b942a95002174140408c65822b5b5597cb157c00048808c480017a0db6ea8b65154da980c40452daab88ca330369ccaf601ba60690f67103c1fb63a5dac48da4000ca9ef57b69f4b200094303c60140b8343eb0d4c132a22150a92abd2800) 

**Expected behavior**
Should display an Image

**Desktop (please complete the following information):**
 - OS:  MacOS
 - Browser chrome,
 - React-pdf version 2.3.0

MiranKorkmaz:
I did const Quixote = () => (
  <Document>
    <Page style={styles.body}>
      <View> 
      		<Image style={styles.image} src="/images/quijote1.jpg" />
      </View>
    </Page>
  </Document>
);

and it rendered just fine. not sure why you do       <View render={() => 
      		<Image style={styles.image} src="/images/quijote1.jpg" />
      } fixed />

MiranKorkmaz:
I did:

const Quixote = () => (
  <Document>
    <Page style={styles.body}>
      <View> 
      		<Image style={styles.image} src=YOUR CHOSEN IMAGE />
      </View>
    </Page>
  </Document>
);

and it worked fine. 
Not really sure why you do render={() => etc...

MiranKorkmaz:
I did `
[...]

Issue #1630, opened 2021-12-02
<Image src={logo}/> returned from render function crashes react-pdf


vitaliiznak:
**Describe the bug**
When returning an `<Image source={logo}/>` component from a render function react-pdf crashes with an error 
```
TypeError: Cannot read properties of undefined (reading 'width')
    at renderImage.js:36
    at _pipe.js:3
    at _pipe.js:3
    at _arity.js:10
    at renderImage (renderImage.js:58)
    at _curryN.js:34
    at _arity.js:10
    at cond.js:42
    at _arity.js:10
    at _pipe.js:3
    at _pipe.js:3
    at _pipe.js:3
    at _pipe.js:3
    at _pipe.js:3
    at _pipe.js:3
    at _arity.js:10
    at renderNode.js:80
.........
```

**To Reproduce**
```javascript
import logo_code from './assets/logo.png';
....
return (
<Document>
  <Page size="A4">
    <View fixed render={({ pageNumber }) =>
      <Image src{logo}/>
    } />
  </Page>
</Document>
)
```

**Expected behavior**
The document renders the same way regardless if the children are returned from the render method or supplied as children to the component.

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser Chrome Version 96.0.4664.45 (Official Build) (64-bit)
 - React-pdf version "^2.0.21"

vitaliiznak:
The exception is raised because the Image node was not resolved ( binary data for the image is not fetched/resolved).

[**Here**](https://github.com/diegomura/react-pdf/blob/master/packages/layout/src/index.js#L30),
**resolveAssets** function gets a DOM tree and for each Image, node resolves binary **src** attribute to binary data representation objects. .
However, because DOM tree does not contain dynamic nodes yet, ( the nodes from render attribute  `<View render={() => <Image> /* will create dynamic node*/ }> )`  it is not resolved.

Dynamic nodes are added on the  [resolvePagination](https://github.com/diegomura/react-pdf/blob/master/packages/layout/src/steps/resolvePagination.js#L36) stage binary assets are resolved to respective Image representation objects.

Will the addition of resolveAssets step [here](https://github.com/diegomura/react-pdf/blob/master/packages/layout/src/steps/resolvePagination.js#L50) fix the issue?

vitaliiznak:
[...]

PR #2240, opened 2023-03-10
fix: fetch assets in dynamic content (#1369, #1587, #1630, #1936, #2064)

mskec:
This PR fixes crashes with images inside dynamic content.

From looking at it, the problem was that assets weren't loaded for images inside dynamic content because `resolveAssets` was executed before `resolvePagination`. 
To resolve this, I've added `resolveAssets` as a step in relayout which forced me to add a bunch of async/await all over `resolvePagination`.

fixes: #1369, #1587, #1630, #1936, #2064


mskec:
Hi @jeetiss, what do you think about this PR? Do you want me to improve something?

tigerza117:
Approve !

mskec:
Hi @jeetiss, @diegomura, it's been a couple of months since opening the PR without receiving any feedback.

Can you please take a look at it? Please let me know if you have some concerns on how it's implemented and maybe we can find another solution.

The PR is fixing a couple of issues that others have reported and they and me would love to have it merged.

tigerza117:
@mskec I have some issue font load by fs not working in render function

PR #2284, opened 2023-04-30
feat: to react-pdf svg components from dom svg

HelgiBergmann:
Ability to generate react-pdf component from existing SVG node and pass to PDF document.

Have some issues:
nested svg - does not work
 strokeDasharray - does not work


edemen:
Was this implemented? Can't seem to find an option to render an SVG that already exists on the page to PDFViewer

PR #2285, opened 2023-05-02
feat: inject into react-pdf Document node from DOM

HelgiBergmann:
Ability to chose DOM element and inject into react-pdf template.

Have some issues:
nested svg - does not work
strokeDasharray - does not work


HelgiBergmann:
@dfdumaresq hello, please take a look

diegomura:
@HelgiBergmann I don't think this should be part of the lib. We need to scope it to react primitives only (same reason React native does not support dom elements). Any transformation between node types should be done by the consumer if needed. Going to close this atm but please do not stop to comment

PR #2293, opened 2023-05-09
fix: getRegisteredFonts type

BryanAbate:
The type of getRegisteredFonts is incorrect, the function returns an object and not an array.
The corresponding js line is here: https://github.com/diegomura/react-pdf/blob/master/packages/font/src/index.js#L76


jeetiss:
Thanks for your contribution ❤️‍🔥

PR #2296, opened 2023-05-13
Fix: code points for glyphs and caching

vaguue:
Hello! I wanted to commend you on the fantastic work you've done with this library, which I've been using recently. I've, however, encountered a bug, an example of which can be found [here](https://github.com/vaguue/react-pdf/tree/bug-example), I've set up an example in the 'prj' module inside the 'packages' directory, which is a Next.js application demonstrating the bug. To observe it, simply run `yarn watch` and open `localhost:3000`. The font used in this instance is Inter, downloaded from Google Fonts. The issue involves characters copied from the generated PDF differing from the characters displayed on the screen (e.g., copying 'C++' results in 'CJJ').

Upon inspecting the generated PDFs, I discovered that the data generated for embedded fonts was incorrect, particularly here:

```
  1 beginbfrange
  <0000> <${toHex(entries.length - 1)}> [${entries.join(' ')}]
  endbfrange
```

This prompted me to initiate some debugging. Initially, I suspected the problem to be in the font caching process. The code implementation struck me as odd: when an embedded font is used, this call

```js
ctx.font(typeof font.name === 'string' ? font.name : font, fontSize); // call to font(src, family, size)
```

eventually passes the Fontkit font object without the family argument, making it uncacheable in the current implementation. Despite adding the family argument, the issue persisted.

I later discovered this segment:

```js
        if (this.unicode[gid] == null) {
          this.unicode[gid] = this.font._cmapProcessor.codePointsForGlyph(
            glyph.id
          );
        }
```

Occasionally, this produces empty arrays, as can be observed in my example. I question the appropriateness of using undocumented private functions. However, replacing it with `glyph.codePoints` seemed to rectify the issue for me.

Best regards,
Vaguue.


vaguue:
I've noticed that my recent PR didn't pass the tests, but it appears to be due to an issue outside of the changes I made. It seems that the CDN image used for testing the module rendering has changed, while the repository still contains an older screenshot. This discrepancy might be causing the test failures.

I thought it would be helpful to bring this to your attention. Please let me know how we should proceed.

diegomura:
Thanks for the kind comments and contributing @vaguue ! This was already fixed in https://github.com/diegomura/react-pdf/pull/2488 :)

PR #2309, opened 2023-05-31
Make `x` and `y` optional in `RectProps`

wojtekmaj:
In SVG, when `<rect>` is not provided with `x` or `y`, it defaults them to 0. In fact, `renderRect` in `@react/pdf-renderer` does the same. So, there's no need to require users to explicitly provide these props.


diegomura (approved):
utACK

PR #2310, opened 2023-05-31
fix: fix all strings allowed in `fontWeight` in `Font.register`

wojtekmaj:
When working on the PDF, I wrote the following code:

```ts
Font.register({
  family: 'Helvetica Neue',
  fonts: [
    { src: helveticaNeueRegular },
    { src: helveticaNeueMedium, fontWeight: '500' },
    { src: helveticaNeueBold, fontWeight: '700' },
  ],
});

  foo: {
    fontFamily: 'Helvetica Neue',
    fontWeight: 500,
  }
});
```

To my surprise, this did not work properly. After investigation, I figured out that my mistake was in defining `fontWeight` as string in `Font.register`, which is something TypeScript was okay with me to do.

Narrowing `fontWeight` type fixes that.

### DIFF
--- a/.changeset/cyan-keys-sing.md
+++ b/.changeset/cyan-keys-sing.md
@@ -1,0 +1,5 @@
+---
+'@react-pdf/types': patch
+---
+
+Fixed all strings allowed in `fontWeight` in `Font.register`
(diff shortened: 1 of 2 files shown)


diegomura (approved):
utACK

PR #2311, opened 2023-05-31
fix: fix crash on `transform="translate(1)`

wojtekmaj:
Providing one value to `translate()` produces valid SVG and is a common thing to see especially in minified SVGs.

When working with such SVG, I got a crash with a rather cryptic error message: "Error: unsupported number: undefined". Upon investigation, the lack of default `y` value turned out to be the cause.

### DIFF
--- a/.changeset/silent-flowers-dream.md
+++ b/.changeset/silent-flowers-dream.md
@@ -1,0 +1,5 @@
+---
+'@react-pdf/render': patch
+---
+
+Fixed crash on `transform="translate(1)"`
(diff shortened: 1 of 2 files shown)


diegomura (approved):

diegomura:
Thanks @wojtekmaj . Tests are failing though 😞

wojtekmaj:
I don't think it's mu fault, I got the same error on all 3 PRs 🤔 Correct me if I'm wrong.

wojtekmaj:
@diegomura I don't think it's my fault, I got the same error on all 3 PRs 🤔 Correct me if I'm wrong.

jeetiss:
@wojtekmaj I fixed [broken test](https://github.com/diegomura/react-pdf/pull/2312), could you please add the changeset to your PR? (just run `yarn changeset`)

@diegomura could you please add `changeset bot` to this repo? https://github.com/apps/changeset-bot will add messages about missed changes, and make it easier to write them for contributors

PR #2355, opened 2023-07-04
Update CHANGELOG.md

tyki6:
fix typo for hook usePdf to usePDF && add options value


diegomura (commented):
Not a typo. document is optional

PR #2363, opened 2023-07-17
fix: SVGPresentationAttributes types

### LINKED ISSUE #1873: Presentation attributes in Svg tag show TS error
**Describe the bug**

In a TS project, when defining any [presentation attribute](https://react-pdf.org/svg#presentation-attributes) in a `<Svg>` element, I get the following TS error:

```
(JSX attribute) strokeWidth: string
No overload matches this call.
  Overload 1 of 2, '(props: SVGProps | Readonly<SVGProps>): Svg', gave the following error.
    Type '{ children: Element[]; width: number; height: number; viewBox: string; strokeWidth: string; strokeLinecap: string; strokeLinejoin: string; style: { marginRight: number; }; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<Svg> & Readonly<SVGProps> & Readonly<{ children?: ReactNode; }>'.
      Property 'strokeWidth' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<Svg> & Readonly<SVGProps> & Readonly<{ children?: ReactNode; }>'.
  Overload 2 of 2, '(props: SVGProps, context: any): Svg', gave the following error.
    Type '{ children: Element[]; width: number; height: number; viewBox: string; strokeWidth: string; strokeLinecap: string; strokeLinejoin: string; style: { marginRight: number; }; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<Svg> & Readonly<SVGProps> & Readonly<{ children?: ReactNode; }>'.
      Property 'strokeWidth' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<Svg> & Readonly<SVGProps> & Readonly<{ children?: ReactNode; }>'.ts(2769)
```

The svg is rendered correctly.

**To Reproduce**

1. Use a presentation attribute in a `<Svg>` tag, ie. `<Svg strokeWidth={2}>...</Svg>`.
2. See the TS error.

See [the reproduction in a TS codesandbox project](https://codesandbox.io/s/exciting-cori-8pcx28?file=/src/components/MyDoc.tsx:196-743). Note that the small svg icon is rendered correctly.

**Expected behavior**

It shouldn't show a TS error in these (as per the docs) allowed attributes.

 - I see the error in my VSCode (1.67.2) editor (in MacOS 11.6.6), as well as in codesandbox online.
 - React-pdf 2.1.1

azmy60:
So I was playing around with `<Svg>` component in `@react-pdf/renderer`, and I got some ts errors:

- Using [Presentation Attributes](https://react-pdf.org/svg#presentation-attributes) as component props like this `<Svg stroke="#00000">` will throw an error.
- There is a missing `strokeLinejoin` prop which I think it should be supported based on [this?](https://github.com/diegomura/react-pdf/blob/a37d927f54d7f367416788979e4b5a8248532952/packages/render/src/primitives/renderSvg.js#L43C1-L43C1)
- There is a typo for the attribute `strokeLinecap`.

The value options of `strokeLinejoin` was provided by looking at the options from [pdf-lib](https://pdf-lib.js.org/docs/api/enums/linejoinstyle).

Fixes #1873


packages/renderer/index.d.ts
```diff
@@ -236,7 +236,7 @@ declare namespace ReactPDF {
 
   class Canvas extends React.Component<CanvasProps> {}
 
-  interface SVGProps extends NodeProps {
+  interface SVGProps extends NodeProps, SVGPresentationAttributes {
```
diegomura, line 239: I'm not sure SVG supports all the props in `SVGPresentationAttributes`. All other fixes look correct

azmy60: Correct me if I'm wrong, but I just checked the [w3 docs](https://www.w3.org/TR/SVG/styling.html#PresentationAttributes), and it seems like SVG supports all Presentation Attributes, except some Geometry Properties like `cx`, `cy`, `r`, `rx`, and `d`. Also, if you try using the native `<svg>` element, the TS will autocomplete all the props in `SVGPresentationAttributes`.

TheMikeyRoss:
image
strokeLinejoin works fine but it shows a TS warning

packages/renderer/index.d.ts
diegomura: I'm not 100% sure our `<Svg />` component supports passing all of these, but I'll include it. I don't actively maintain types tbh, so I'll let base it on what the community finds :)

Issue #1676, opened 2022-01-12
Unbreakable elements taller than the page don't correctly start on next page and wrap, as expected


fnimick:
**Describe the bug**
Based on https://github.com/diegomura/react-pdf/issues/874, I would expect a view set to `wrap={false}` to start at the top of the next page, and wrap from there, if it is taller than the page height.

**To Reproduce**
Steps to reproduce the behavior including code snippet (if applies):
1. Go to [here](https://react-pdf.org/repl?example=disable-wrapping)
2. Observe that `unbreakable` is currently rendering on page 2.
3. Change `unbreakable` height from 400 to 800
4. Observe that rather than being rendered on pages 2 and 3, it instead renders on page 1 and overflows off the bottom of the screen.

**Expected behavior**
Unbreakable views that are too tall to fit on one page should start on the next page and wrap as described.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: [MacOS]
 - Browser [Chrome]
 - React-pdf version [e.g. 2.0.21]

diegomura:
Should be closed by https://github.com/diegomura/react-pdf/pull/3502. Create a new issue if persist

Issue #1037, opened 2020-10-23
Page wrap ( NOT WORKING 


alew140:
**Describe the bug**
```
    <Page wrap size="A4" style={styles.body}>
           <Text break>other page</Text>
     </Page> 
```

 i have only a page ( Not break and generate other page is one page )

**To Reproduce**
```
import { Page, Font, Image, Text, View, Document, StyleSheet } from '@react-pdf/renderer';
  body:{
    paddingTop: 100,
    paddingBottomm: 100,
    paddingLeft: 30,
    paddingRight: 30,
    fontSize: 10,
    fontFamily: 'Open Sans'
  },
    <Page wrap size="A4" style={styles.body}>
           <Text break>other page</Text>
     </Page> 

```
**Screenshots**
![image](https://user-images.githubusercontent.com/27329658/96994840-42808d80-14fb-11eb-93fa-51e287cc7a48.png)

**Desktop (please complete the following information):**
 - OS: [Windows]
 - Browser [chrome]
    "@react-pdf/renderer": "^1.6.11",

volkar1:
I had the same problem, seems like padding style on <Page> tag breaks page wrapping. Try nesting <View> in <Page> and put <Text> in <View> then remove all the styles from page and put in on <View>... Think of <View> tag like a div tag.

Your code would look something like this:
`    <Page wrap size="A4">
<View style={styles.body}>
           <Text break>other page</Text>
</View>
     </Page> `

alew140:
![image](https://user-images.githubusercontent.com/27329658/97093488-6b4b7480-161a-11eb-9b9e-0df0aa61b975.png)
```
  <Document>
  <Page wrap size="A4"> 
  <View>
  <Text>page 1</Text> 
  <Text break>other page</Text> 
  </View> 
  </Page>
  </Document>
```

alew140:
![image](https://user-images.githubusercontent.com/27329658/97093530-9d5cd680-161a-11eb-860f-221c9041b35f.png)
![image](https://user-images.githubusercontent.com/27329658/97093539-b6658780-161a-11eb-861c-bb8502cd51fa.png)

nothing

alew140:
*Downgrade to 1.6.10 and try...*

alew140:
IN VERION 1.6.10 NOT WORK show all my document 

```
  <Document>
    <Page wrap size="A4" style={styles.body}>
      <Image fixed
        style={styles.pageBackground}
        src={watermark}
      />
[...]

PR #2403, opened 2023-09-28
Initial nested page break handling. And change to unwrappable content fallback behavior..

Scong:
Adds loop to split nodes by nested page break using splitNode function.

If a node is found that has a breakable child, after performing standard pagination resolutions it will split children into two parts, putting the content after the page break in the second part, then rely on the relayout behavior to recompute as needed. 

I have also changed the behavior for nodes marked as non-wrappable to fallback to wrapping if they do not fit on a page and could be wrapped. 

I'm not sure if there will be other considerations here. 

Looking for feedback. 

Relating issues/enhancements might include:
https://github.com/diegomura/react-pdf/issues/1037

https://github.com/diegomura/react-pdf/issues/1676


Scong:
Verifying the behavior here a bit more, I found it possible to get in an infinite loop on complicated reports.

PR #2409, opened 2023-10-09
feat: add support for native ESM imports, make repository ESM-first 

### LINKED ISSUE #2068: node esm imports not working correctly
**Describe the bug**
recent versions of node (i think >= 16) support es modules natively:
```
import { StyleSheet, View } from '@react-pdf/renderer'
```
however, the above code actually pulls in the `.cjs` build instead.

**To Reproduce**
1. install node 18
2. create `foo.mjs` with the above line
3. `NODE_DEBUG=module node foo.mjs`
4. note the output:
```
MODULE 17732: looking for "@react-pdf/primitives" in ["/Users/redacted/git/project/node_modules/@react-pdf/renderer/lib/node_modules","/Users/redacted/git/project/node_modules/@react-pdf/renderer/node_modules","/Users/redacted/git/project/node_modules/@react-pdf/node_modules","/Users/redacted/git/project/node_modules","/Users/redacted/git/node_modules","/Users/redacted/node_modules","/Users/node_modules","/node_modules","/Users/redacted/.node_modules","/Users/redacted/.node_libraries","/Users/redacted/.nvm/versions/node/v18.12.0/lib/node"]
MODULE 17732: looking for "/Users/redacted/git/project/node_modules/@react-pdf/renderer/lib/react-pdf.cjs.js" in ["/Users/redacted/.node_modules","/Users/redacted/.node_libraries","/Users/redacted/.nvm/versions/node/v18.12.0/lib/node"]
MODULE 17732: load "/Users/redacted/git/project/node_modules/@react-pdf/renderer/lib/react-pdf.cjs.js" for module "/Users/redacted/git/project/node_modules/@react-pdf/renderer/lib/react-pdf.cjs.js"
```

**Expected behavior**
the module build should be loaded instead. 

i have a pr to resolve this (#2069)

wojtekmaj:
## Overview

Fixes #2068

Converts the repository to be ESM first. Ensures that every package published has `"type": "module" and set "exports"`. As a natural consequence, tools like ESLint, Jest are running in ESM mode too.

With these changes, I'm able to import and use react-pdf in ESM Node.js projects, and build it with esbuild too.

### Setting `"type": "module"` and `"exports"`

Before this PR, `package.json`s contained `main` and `module` entires. The former is universal, the latter - kinda-supported-by-most-bundlers-maybe way to point them to ESM modules. But this does not work with Node.js. To make Node.js aware of the ESM version, `"type": "module"` and `"exports"` must be set.

Note: We may want to push this further by introducing `.cjs` extensions, or use `/dist/cjs` and `/dist/esm` directories, each with their own package.json with `"type": "module"`/`"commonjs"`.

### Adding file extensions in select places

There were 4 files that appear to be running in Node without bundling or any other help that would handle extensionless module resolution. For the whole repository to work with Node.js in ESM mode, I had to add `.js` to 6 imports.

Note: We may want to push this further by adding all required extensions everywhere. This PR includes the **bare minimum** to make react-pdf work with ESM imports. In modern Node.js, you are actually _required_ to include extensions in your imports. However, this would mean I would need to change hundreds of files at once, which would make it unlikely for this PR to ever get merged. It's quite large anyway...

### Configuration

Due to Jest changes (see below), `jest/globals` were removed from ESLint `env` config. Because of `@jest/globals` import, an exemption from `import/no-extraneous-dependencies` was also added for unit tests.

#### Babel

By setting `"type": "module"`, `module.exports` no longer worked in `babel.config.js`. This had to be changed to `export default`.

#### ESLint

Similarly to Babel, `module.exports` was no longer an option. However, ESLint in your repository didn't want to play nicely with ESM either, so I opted for pure JSON instead.

#### Jest

By far most significant change in this PR. 

* Jest doesn't play nicely with ESM modules; support for it is experimental. It also required me to change the command from just `jest` to `yarn node --experimental-vm-modules \"$(yarn bin jest)\"`.
* Some unit tests were using `require()` syntax, which didn't work anymore.
* In ESM mode, `jest` global is not available, to `jest` had to be explicitly imported from `@jest/globals` instead.
* In ESM mode, `__dirname` is not available, so `path.dirname(url.fileURLToPath(import.meta.url))` "polyfill" was used instead.

## Additional notes

Note: I was forced to add updated node-gyp to our devDependencies for the project to even install correctly, in 4c0fbcd0edd3fba25b0f85552e536ba8b0a1c26a. I may split it into a separate PR if you want.

Note: While I'm quite confident this setup works very well in ESM Node.js, CJS was not verified in any way. It would be a good idea to test it out in legacy Node.js environment before merging.

Note: The PR is currently failing due to an issue in external dependency, browserify-zlib. I have raised a PR to fix it: https://github.com/browserify/browserify-zlib/pull/45


packages/renderer/package.json
```diff
@@ -6,8 +6,16 @@
   "types": "index.d.ts",
   "author": "Diego Muracciole <diegomuracciole@gmail.com>",
   "homepage": "https://github.com/diegomura/react-pdf#readme",
+  "type": "module",
   "main": "./lib/react-pdf.cjs.js",
   "module": "./lib/react-pdf.es.js",
+  "exports": {
+    ".": {
+      "import": "./lib/react-pdf.es.js",
```
wojtekmaj, line 14: Note: we also need to add `types` there, as types are not siblings of the outputted JS files.

diegomura:
@wojtekmaj would be great to add this. Can you rebase? 🙏🏻

wojtekmaj:
@diegomura There you go!

wojtekmaj:
We're still affected by https://github.com/browserify/browserify-zlib/pull/45. Wondering what we can do about it.

wojtekmaj:
@diegomura Resolved Jest issue, but I'm not entirely happy with it (_of course_, since Jest can't handle ESM properly). However, I'll have Vitest migration ready to go for you as soon as this one gets merged :D

packages/renderer/package.json
diegomura: How would this affect backwards compatibility? I know some module resolvers do not work with exports yet right?

diegomura:
Thanks @wojtekmaj ! Can't check in full detail now but I'll soon. In the meantime, do you mind explaining the benefits of moving in this direction? I get the benefits of ESM but my worry is this will cause issues for lots of consumers using old bundlers or node versions.

package.json
```diff
@@ -61,12 +62,14 @@
     "jest-image-snapshot": "^6.1.0",
     "lerna": "^8.0.2",
     "lint-staged": "^10.5.4",
+    "node-gyp": "^10.0.0",
```
diegomura, line 65: Is this still needed?

packages/layout/tests/node/setAlignSelf.test.js
```diff
@@ -1,7 +1,12 @@
-import Yoga from 'yoga-layout/sync';
+import { jest } from '@jest/globals';
+import yogaModule from 'yoga-layout/sync';
 
 import setAlignSelf from '../../src/node/setAlignSelf';
 
+// yoga-layout sets default export using non-standard __esModule property, so we need to
+// make an additional check in case it's used in a bundler that does not support it.
+const Yoga = 'default' in yogaModule ? yogaModule.default : yogaModule;
```
diegomura, line 8: Can we extract this in a `layout/yoga.js` file? We can capture it and I can do it separately. It might not be necessary either as I'd like to experiment with async yoga which I think it's more performant and uses wasm build

packages/pdfkit/src/font/data/compressData.js
```diff
@@ -1,6 +1,11 @@
 import fs from 'fs';
+import url from 'url';
 import { basename, extname } from 'path';
-import { parse } from '../afm';
+// This file is ran directly with Node - needs to have .js extension
```
diegomura, line 4: Don't we always need .js extension with ESM?

packages/renderer/package.json
```diff
@@ -3,26 +3,33 @@
   "version": "3.1.17",
   "license": "MIT",
   "description": "Create PDF files on the browser and server",
-  "types": "index.d.ts",
```
diegomura, line 6: No longer needed?

diegomura (approved):
utACK

All look good! Left some small questions to answer before I merge but I'll merge immediately after. Thanks for the work @wojtekmaj

diegomura:
> Older bundlers/Node.js versions will simply ignore the new entry

This is good to hear. I was thinking if would be convenient to otherwise do a major bump. Recent yoga-layout bump included library using `exports`, and I already got lots of complaints about envs breaking. But does not seem to be necessary then

package.json
wojtekmaj: Hmmm, works without it, but please mind, it's gonna stay in our deps anyway. I can see it was added in #2478.

packages/layout/tests/node/setAlignSelf.test.js
wojtekmaj: Sure, defining that 35 times was idiotic :D Fixed that.

I don't think move to wasm will change anything here though.

packages/pdfkit/src/font/data/compressData.js
wojtekmaj: Yes and no! In most cases, react-pdf is pre-bundled. When bundled, module resolution is already done by the bundler and these rules no longer apply. 

I WOULD do this if I were you sooner or later, see "Adding file extensions in select places" in original post for rationale.

packages/renderer/package.json
wojtekmaj: Not only no longer needed, but actively harmful for type resolution in TypeScript. Instead, each bundled .js file gets its own .d.ts sibling file. This makes the bundle larger, unfortunately, but this is the reason we'll be seeing a vast improvement on arethetypeswrong after this PR.

wojtekmaj:
> This is good to hear. I was thinking if would be convenient to otherwise do a major bump. Recent yoga-layout bump included library using exports, and I already got lots of complaints about envs breaking. But does not seem to be necessary then

I genuinely believe we can do this without the major bump. I successfully migrated more than 30 OSS repositories of my own to ESM-first in a minor bump and received 0 complaints. Granted, they were much less complex than react-pdf, but I now know a thing or two, and worry not - if anything goes south, I'll have your back while bugfixing :)

diegomura:
Thanks for addressing all my questions and the work done here!

PR #2411, opened 2023-10-11
Break on breaks for View components

Scong:
**Changes**
1. Add logic for node splitting to continue to split nodes while there is a nested break inside of attached to a View. Some adjustments to logic to avoid a possible infinite loop. 

2. Change to unwrappable fallback behavior for View / Text components to wrap when they don't fit within a page, to avoid content being lost. A warning is still logged, this behavior should likely be handled explicitly.

**Notes:** 

Splitting on nested text breaks has some complexities, I believe, in that Lines are computed at the top level Text component (in the case of text components inside of text components). So a recomputation of lines would need to occur after doing a regular node split or such.

I believe nested text breaking is not really necessary, and should be able to be accomplished by adding break tags to view components. All though digging to top level text components with break tags could be reasonable to add. I added a test to demonstrate the relating behavior for nested text / view breaks.

This relies on the split logic and supporting logic behaving reasonably on deeply nested items. It appeared to, all though there my be other considerations here?

I published the fork to [npm scong-react-fork-pdf-renderer](https://www.npmjs.com/package/scong-react-fork-pdf-renderer) for our use. (This is done through the distribution-2 branch in my fork repo)

**Some related issues / enhancements** 

https://github.com/diegomura/react-pdf/issues/1037

https://github.com/diegomura/react-pdf/issues/1676


Scong:
I believe there are additional considerations that need to be handled by the split function in order to handle this appropriately. All though the tests behave as expected, in more complicated reports the splitting still does not behave as desired.

Scong:
@diegomura I know in the past you mentioned that nested page breaks is a tricky topic. Do you know where else I should be looking, I believe there is some consideration around height re-computation perhaps? 

We really like the library, unfortunately we need a bit more nested pagination support for our report generation consistency. Appreciate any insight here!

PR #2416, opened 2023-10-18
feat: updated svg.d.ts to include strokeLinejoin

TheMikeyRoss:


TheMikeyRoss:
Hi @diegomura ,

I'm kinda of a noob and I'm not entirely familiar with what a changeset file is, I'll have to read more about it.

From the quick reading I did it seems to be a command line tool only. And I've made the changes to this code directly in github web. Is there a way I can add the changeset file using github website?

PR #2437, opened 2023-11-04
added currentPageCallBack for the process of the current page

KMJ-007:
in large pdf, want to show the progress of the currently processed pages, so user can also know where the process has reached, and don't close the window

let me know what i can do better, i just had this in my mind, so implemented it


KMJ-007:
@dfdumaresq can you look into this ?

ASutrick:
Would love to see this implemented

packages/renderer/src/index.js
```diff
@@ -45,7 +45,7 @@ const pdf = initialValue => {
       pageMode,
     });
 
-    const layout = await layoutDocument(container.document, fontStore);
+    const layout = await layoutDocument(container.document, fontStore, currentPageCallBack);
```
diegomura, line 48: This seems very specific to this case. Seems likely we want to report more things to the outside. Maybe we can pass an event emitter to layout, and internally emit events?

KMJ-007: call back looks fairly simple and straightforward, but as you are saying to pass an event to layout, when i was doing the call back, i tried to add the calllback to all the 12 functions, but it was slowing the speed of pdf creation, so i just added them to few, worried about the performance overhead if we implement in all the steps,

emitting events also makes senes to me, both works!

diegomura:
Check https://github.com/diegomura/react-pdf/pull/3502

PR #2453, opened 2023-11-28
fix: add clipPath to typing for SVGPresentationAttributes

chuckspencer-aiq:
Using the clipPath attribute was causing a type error in Typescript because it was missing from the type def.


diegomura (approved):
utACK 

Thanks!

PR #2463, opened 2023-12-17
feat: support for webp

imyasha:
Support for webp

Need to update pdfkit to support webp as well.


imyasha:
@diegomura 
I think we need to update pdfkit to support webp.

diegomura:
Tested this but it does not work. Pdfkit does not seem to support webp images either. Is there something I missed?

PR #2465, opened 2023-12-19
fix for dynamic image render

Justinio99:
There are ongoing challenges with dynamically rendering the <Image /> component. The primary issue identified was the absence of image properties on the child node. To rectify this, I introduced a new validation step for the Image type, ensuring the creation of a new node instance with the appropriate attributes.
```javascript
  if (isImage(element)) {
    const node = { props, type, style, box: {} };
    await fetchImage(node);
    return [node];
  }
``` 
To facilitate this solution, I needed to modify several functions to execute asynchronously, particularly because fetchImage was already an asynchronous function.

As my expertise with react-pdf is limited, I'm hopeful that this solution proves effective. I'm more than open to feedback and help in this area.

### DIFF
--- a/packages/layout/src/node/createInstances.js
+++ b/packages/layout/src/node/createInstances.js
@@ -1,9 +1,14 @@
 import { castArray } from '@react-pdf/fns';
-import { TextInstance } from '@react-pdf/primitives';
+import { TextInstance, Image } from '@react-pdf/primitives';
+import flatten from 'lodash.flatten';
+
+import fetchImage from '../image/fetchImage';
 
 const isString = value => typeof value === 'string';
 
 const isNumber = value => typeof value === 'number';
+
+const isImage = value => value.type === Image;
 
 const isFragment = value =>
   value && value.type === Symbol.for('react.fragment');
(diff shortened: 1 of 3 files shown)


mskec:
Tests do not pass for me with the fix.

jrtell1:
Would appreciate this fix a lot! :)

packages/layout/src/node/createInstances.js
```diff
@@ -16,7 +21,7 @@ const isFragment = value =>
  * @param {Object} React element
  * @returns {Array} parsed react elements
  */
-const createInstances = element => {
+const createInstances = async element => {
```
diegomura, line 24: I don't think this is correct architectural-wise. We are mixing concepts here. `createInstances` should not fetch images, that's done separately on another layer. Can we preserve that?

Justinio99: I'm not very familiar with this project and was just looking for a quick solution. Perhaps this issue would be better addressed by someone more acquainted with the codebase.

santialbo: I suggested 2 possible approaches here https://github.com/diegomura/react-pdf/issues/1369#issuecomment-1087910517
Do they make sense @diegomura?

PR #2471, opened 2024-01-05
chore(deps): upgraded lerna to v5

mskec:
Upgraded lerna to v5 because v3 does not work with macos with arm.


diegomura:
Upgraded to 8. Thanks for the contribution!

PR #2473, opened 2024-01-05
chore(deps): Upgraded yoga to v2 and removed bundled package

mskec:
Yoga v2 can be used as a normal package so there is no need to bundle it and publish with react-pdf.


diegomura:
Amazing! Thanks so much

diegomura (approved):
utACK

PR #2494, opened 2024-01-15
Go all-in with Jest ESM migration

wojtekmaj:
Requires #2409.

This is an _optional_, intermediate step between full Vitest migration. It completely drops usage of Jest's magic globals, instead relying on `@jest/globals`, for importing `describe`, `test`, `expect`, `it`… 

This is probably for the better regardless of whether or not we would like to migrate to Vitest, as this makes Jest far easier to work with in TypeScript environment, especially when working along other testing tools like Playwright. 

If we _do_ want to migrate to Vitest though, this PR will make majority of the migration to be as simple as find-and-replace `@jest/globals` to `vitest`.

* If you don't want react-pdf to be migrated to Vitest, consider the above, feel free to close if you don't find this useful.
* If you want react-pdf to be migrated to Vitest, and:
  * don't want extra clarity (noise?) in your commits on how this migration is done, feel free to close. This will not block Vitest migration PR.
  * want extra clarity on Jest -> Vitest migration process in your commits, please merge.


wojtekmaj:
Screw this - Jest is fundamentally broken. Let's go straight for #2495.

PR #2495, opened 2024-01-15
chore: migrate from Jest to Vitest

wojtekmaj:
Requires #2409.

~~Requires #2494.~~

Migrates the entire monorepo from Jest to Vitest.


diegomura:
@wojtekmaj Just to understand this better (not familiarized with vitest), the benefit of this is better support for ESM and more performant? About the latter I'm not so concern as the entire testsuite runs in like 6sec. Also curious how tight this lib is ti vite, considering we don't use it here

diegomura:
Actually, about performance, I tested this branch and tests take 50% more to run (9s). It's not bad either s not concerned, but still curious :)

wojtekmaj:
Hmmm, interesting. You're right that the overall run time has increased a bit, and this is while the individual tests run faster:

Before:
 PASS  packages/renderer/tests/resume.test.js (6.884 s)

After:
 ✓ |renderer| tests/resume.test.jsx  (3 tests) 3767ms

Before:
 PASS  packages/renderer/tests/pageWrap.test.js (7.924 s)

After:
 ✓ |renderer| tests/pageWrap.test.jsx  (1 test) 3553ms

This makes me convinced that while there is probably something I could still tweak, the more tests and the more complex they are, the faster Vitest will be in relation to Jest.

wojtekmaj:
Now it's 35% faster than Jest on CI! ;)

PR #2502, opened 2024-01-17
feat: include text preceeding special case emojis, include ™ symbol in excluded emojis

DominikMeca:
This PR fixes an issue where text fragments containing a special case symbol (trademark/copyright symbol) would get replaced with just the symbol. It also adds a new excluded symbol (™).


diegomura (approved):
utACK

Thanks!

PR #2505, opened 2024-01-18
feat: wasm yoga

diegomura:
Uses `loadYoga` from `yoga-layout` main entrypoint to leverage wasm internal usage.

This is supposed to ship less code and also be more performant. In tests I couldn't see much of a difference to be honest, but it's cool we run flex engine in wasm code as it's supposed to be more performant. Maybe in larger documents will be more noticeable. It does not introduce breaking changes as layout process is already async

I'd appreciate some thoughts about this :) cc/ @wojtekmaj @jeetiss

### DIFF
--- a/.changeset/slimy-kiwis-cry.md
+++ b/.changeset/slimy-kiwis-cry.md
@@ -1,0 +1,6 @@
+---
+'@react-pdf/renderer': minor
+'@react-pdf/layout': minor
+---
+
+feat: wasm yoga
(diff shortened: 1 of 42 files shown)


wojtekmaj:
From my ESM stuff perspective, that looks alright :D

moishinetzer:
This seems to have caused the following issue: #2589

moishinetzer:
For anyone following this issue, as a temporary fix downgrade to the any of the following packages you may use:

@react-pdf/layout@3.9.0
@react-pdf/renderer@3.2.0

And make sure to pin it in your package.json

PR #2510, opened 2024-01-18
fix: fix require @react-pdf/primitives from CJS

wojtekmaj:
In #2409 I missed changing the extension of CJS code to .cjs, resulting in (inaccurate) ERR_REQUIRE_ESM error thrown in Node in CJS environment.

### DIFF
--- a/.changeset/brown-houses-count.md
+++ b/.changeset/brown-houses-count.md
@@ -1,0 +1,5 @@
+---
+'@react-pdf/primitives': patch
+---
+
+Fix require @react-pdf/primitives from CJS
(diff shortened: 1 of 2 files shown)


gabrielhamel:
arfff thx i didn't figure out why i have esm error in my project 👍

wojtekmaj:
No need to thank me, I was the one to mess it up in the first place 🤣

gabrielhamel:
Yeah i saw that ^^ you managed to broke an +13k⭐️ project for few hours. It's happen sometimes ^^
That recall me that i need to freeze my npm package versions and doesn't trust the carret prefix in package.json !

wojtekmaj:
Carret is fine, unless you fail to commit your lockfile, which you definitely should.

PR #2514, opened 2024-01-18
fix: fix 100s of (mostly) JSDoc issues found by TypeScript

wojtekmaj:
This is a DX improvement and a preparation PR if we ever decide to land TypeScript in this repository (@diegomura, let me know your thoughts!).

In this PR:

- I added TypeScript dependency and configured it to non-strict JS type checking (not included in the PR)
- I went through errors one by one and fixed them.
- I pushed it a bit more by changing generic Array type to something[] to improve type accuracy.

Before:
> Found 379 errors in 148 files.

After (incl. other fixes I've raised PRs for):
> Found 24 errors in 8 files.

Much more manageable, isn't it.

Typical issues found:

- parseFloat does not receive second argument. parseInt does.
  ```js
  parseFloat(num, 10);
  ```
- Incorrectly typed definitions: param name not provided (only description)
  ```js
  /**
   * @param {Object} parent document
   */
  const foo = (node) => // …
  ```
- Incorrectly typed definitions: params defined at top level when function returned a function
  ```js
  /**
   * @param {string} bar
   * @param {string} baz
   */
  const foo = (bar) => (baz) => …
  ```

This is by no means a finished job, but it's a clear improvement that pushes us in the right direction, TypeScript or not.

### DIFF
--- a/packages/textkit/src/engines/linebreaker/linkedList.js
+++ b/packages/textkit/src/engines/linebreaker/linkedList.js
@@ -19,6 +19,7 @@ class LinkedList {
     this.head = null;
     this.tail = null;
     this.listSize = 0;
+    this.listLength = 0;
   }
 
   isLinked(node) {
--- a/packages/fns/src/asyncCompose.js
+++ b/packages/fns/src/asyncCompose.js
@@ -3,9 +3,17 @@
 import reverse from './reverse';
 
 /**
+ * @typedef {Function} AsyncCompose
+ * @param {any} value
+ * @param {...any} args
+ * @returns {any} result
+ */
+
+/**
  * Performs right-to-left function composition with async functions support
  *
- * @param  {...any} functions
+ * @param {...Function} fns functions
+ * @returns {AsyncCompose} composed function
  */
 const asyncCompose = (...fns) => async (value, ...args) => {
   let result = value;
--- a/packages/fns/src/capitalize.js
+++ b/packages/fns/src/capitalize.js
@@ -1,12 +1,12 @@
 /**
  * Capitalize first letter of each word
  *
- * @param {String} string
- * @returns {String} capitalized string
+ * @param {string} value string
+ * @returns {string} capitalized string
  */
-const capitalize = value => {
+const capitalize = (value) => {
   if (!value) return value;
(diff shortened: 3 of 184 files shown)


packages/textkit/src/engines/linebreaker/linkedList.js
```diff
@@ -19,6 +19,7 @@ class LinkedList {
     this.head = null;
     this.tail = null;
     this.listSize = 0;
+    this.listLength = 0;
```
wojtekmaj, line 22: @diegomura This is an interesting one. TypeScript caught this.listLength being used (see [line 95](https://github.com/diegomura/react-pdf/pull/2514/files#diff-43ed2d44fbfd569877dfaafe36f3f9526ee1b38490aa3a11ce3326cca33338a2R95)) but chances are that it's a bug and it should be listSize instead. Thoughts?

diegomura (approved):
utACK

Thanks for taking the time to do this! I knew the JSDocs were incorrect or lazy, but never wrote them to perform type-checks on them, was mostly for internal documentation (mostly for myself as I was the only one opening contributing), so never stuck to actual schema. But obviously this is way better :)

About TS, I'm don't think I'll be considering adding it anytime soon (sorry 😅 ). Despite not being a huge fan myself of it, it would just get in the way in relation to the pros it would bring. Issues we have right now are not type related, and code and data models are build in a way it's not necessary. I'm open to hear thoughts about it ofc

wojtekmaj:
@diegomura what TypeScript requires from you and what, in return, it can give back can be fully configurable.

This PR was based on not much more than installing TypeScript, teaching it to allow and check JS files, and setting strict to false (obviously not gonna happen anytime soon).

At this stage, it behaves more like a smarter linter than a real type checker, but it did catch a ton of minor issues nonetheless.

I'll keep it on my working branch, and when it's ready (it's not - as you can see if not for this PR you'd see the code pretty much all in red) and when I feel like it wouldn't "get in the way" for you, I'll raise a proposal PR, letting you play around and agree or disagree with my opinion.

Whether or not to perform a full TypeScript migration is not the scope of this discussion and doesn't need to be, as the road to that needs to be paved first, so let's put this aside.

PS. Please check my review comment above - I'm 90% sure we caught a bug there.

PR #2518, opened 2024-01-18
fix: invalid prop (style) in Resume example

wojtekmaj:
This custom Item component actually accepts children only.


diegomura (approved):
utACK

PR #2519, opened 2024-01-18
fix: props unnecessarily required in LinearGradient and RadialGradient in TypeScript

wojtekmaj:
Examples show that this is totally fine:

https://react-pdf.org/repl?example=lineargradient
https://react-pdf.org/repl?example=radialgradient


diegomura (approved):
utACK

PR #2520, opened 2024-01-18
fix: remove unnecessary check for response.buffer

wojtekmaj:
If TypeScript is to be believed, cross-fetch's Response never returns buffer, so this check is unnecessary.

### DIFF
--- a/.changeset/slimy-suns-drive.md
+++ b/.changeset/slimy-suns-drive.md
@@ -1,0 +1,5 @@
+---
+'@react-pdf/image': patch
+---
+
+Remove unnecessary check for response.buffer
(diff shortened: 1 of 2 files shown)


diegomura (approved):
Looks good but not sure how can I be sure 😄

wojtekmaj:
Will provide more technical details when I find some free time.

In the meantime, let's keep it open, especially since I now realized that the check below my change is also redundant if TypeScript was indeed right :D

There are more exciting changes though :) I hope I didn't overwhelm you in the last couple of days!

diegomura:
Not at all @wojtekmaj ! I'm thrilled you are contributing so much. Thanks for all the work man. 

I hope I don't become a big pita with my TS scepticism 😄 I'm open to discussing but to be honest I don't feel very excited about switching to it. Mostly commenting it so you don't lose a lot of time migrating before discussing it :)

wojtekmaj:
I absolutely get you, especially since I hesitated for YEARS before I convinced myself to try TypeScript out.

I respect your decisions and trust your intuition when it comes to react-pdf maintenance. It's your project after all, and I have no right to make it feel any less yours.

PR #2521, opened 2024-01-18
fix: fix invalid display property in objectFit example

wojtekmaj:
There's no display: block in react-pdf, only flex or none.


diegomura (approved):
utACK

PR #2524, opened 2024-01-19
feat: parameterize Styles

ludicast:
The `Styles` interface has the `Style` interface as its values.  Sometimes it's useful to generate these types dynamically, so you can take advantage of type hinting.

This PR adds the option for the Styles interface to also allow as its values functions that return a `Style` object.

### DIFF
--- a/.changeset/brown-houses-count.md
+++ b/.changeset/brown-houses-count.md
@@ -1,0 +1,5 @@
+---
+'@react-pdf/primitives': patch
+---
+
+Fix require @react-pdf/primitives from CJS
--- a/packages/primitives/package.json
+++ b/packages/primitives/package.json
@@ -6,12 +6,12 @@
   "author": "Diego Muracciole <diegomuracciole@gmail.com>",
   "homepage": "https://github.com/diegomura/react-pdf#readme",
   "type": "module",
-  "main": "./lib/index.js",
+  "main": "./lib/index.cjs",
   "module": "./src/index.js",
   "exports": {
     ".": {
       "import": "./src/index.js",
-      "require": "./lib/index.js",
+      "require": "./lib/index.cjs",
       "default": "./src/index.js"
     }
   },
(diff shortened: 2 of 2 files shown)


packages/renderer/index.d.ts
```diff
@@ -487,7 +487,7 @@ declare namespace ReactPDF {
    * @see https://react-pdf.org/advanced#on-the-fly-rendering
    * @platform web
    */
-  class PDFDownloadLink extends React.Component<PDFDownloadLinkProps> {}
+  class PDFDownloadLink extends React.Component<PDFDownloadLinkProps> { }
```
diegomura: Can we revert these changes? Was good as it was before too and this is unrelated to PR

diegomura (commented):
Looks good! Can we add some tests to make sure this does not break in the future?

packages/renderer/index.d.ts
```diff
@@ -212,8 +212,7 @@ declare namespace ReactPDF {
      * @see https://react-pdf.org/advanced#debugging
      */
     debug?: boolean;
-    href?: string;
-    src?: string;
+    src: string;
```
diegomura, line 215: Regression here

packages/renderer/index.d.ts
```diff
@@ -362,9 +361,7 @@ declare namespace ReactPDF {
    */
   class Tspan extends React.Component<React.PropsWithChildren<TspanProps>> {}
 
-  interface GProps extends SVGPresentationAttributes {
-    style?: Style;
-  }
+  interface GProps extends SVGPresentationAttributes {}
```
diegomura, line 364: Regression here

packages/renderer/index.d.ts
```diff
@@ -419,11 +416,11 @@ declare namespace ReactPDF {
...
-    cx?: string | number;
-    cy?: string | number;
-    fr?: string | number;
-    fx?: string | number;
-    fy?: string | number;
+    cx: string | number;
+    cy: string | number;
+    fr: string | number;
+    fx: string | number;
+    fy: string | number;
```
diegomura, line 423: Ditto

ludicast: yeah, cleaning up & adding tests, going to force push over this

PR #2527, opened 2024-01-19
fix: add missing types for parseDataPath function

wojtekmaj:
Types references were copied from the original repository, but type definitions were not. This fixes that.


diegomura (approved):
utACK

PR #2544, opened 2024-01-22
fix: creationDate and modificationDate types

satelllte:
These are still missing in the renderer, so adding them.

In general, `renderer/DocumentProps` should extend from `types/DocumentProps` ideally, but that will be a larger refactor task


diegomura (approved):
utACK

PR #2549, opened 2024-01-23
fix: specifying ids on elements (named destinations)

carlobeltrame:
Fixes https://github.com/diegomura/react-pdf/commit/36c6ba30ae73a512f19fe5bc47ac8c304887c0da#r137623912

Unfortunately I can't provide a REPL, because the REPL seems to run react-pdf in an older version. The bug was introduced in @react-pdf/pdfkit@3.1.1 and led to our CI failure in https://github.com/ecamp/ecamp3/pull/4495

Instead of a REPL, I added a regression test which at least uses the `id` attribute (with a non-number value). Running this test with 3.1.1 would yield the error `unsupported number: NaN`. This is because the name_tree class was wrongly changed to use parseInt instead of `new String()` in 36c6ba30ae73a512f19fe5bc47ac8c304887c0da.


diegomura (approved):
utACK

Thanks!

PR #2565, opened 2024-01-26
chore: split GitHub Actions workflows into jobs; run tests on multiple Node.js versions

wojtekmaj:
This PR revamps GitHub Actions workflows.

What was done:
- All actions used were bumped to their latest versions to ensure no deprecation warnings are being shown.
- Split workflows into multiple jobs, so e.g. checking size can run only once, but unit tests - multiple times, which brings me to…
- Unit tests were configured to run on multiple versions of Node.js (18 and 20).
- Other jobs were bumped to run on Node.js 20 instead of 18.
- Added `--frozen-lockfile` to install command to ensure no changes are being sneaked in.
- Set up Yarn cache to speed up CI process

While the total CI run time is significantly increased (we're doing far more testing than we used to), the overall CI duration has _decreased_, because all jobs are running in parallel and additional optimizations were done to each of them.

Before:
image

After:
image

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -28,9 +28,6 @@ jobs:
           cache: yarn
           node-version: ${{ env.NODE }}
 
-      - name: Install required packages
-        run: sudo apt install libsdl-pango-dev libgif-dev
-
       - name: Install dependencies and build
         run: yarn --frozen-lockfile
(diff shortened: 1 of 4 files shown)


diegomura (approved):
utACK

Thanks so much!

PR #2567, opened 2024-01-26
chore: update Yarn to v4; migrate from Lerna to Yarn Workspaces

wojtekmaj:
This is an **experimental** PR to see the potential effect of upgrading Yarn to v4.

Why not Yarn v1:
- It's no longer maintained.
- It has many known bugs.
- It's far slower than virtually every modern competitor (including npm).

Why Yarn v4:
- Fast.
- Very easy, pretty much automatic upgrade.
- Bulit-in workspaces, allowing us to drop Lerna and has all its features built-in.
- Automatic dependency management ("workspace:" protocol) - changes to ^x.x.x where x.x.x is the latest dependency version at the moment of publishing
- Built-in portable shell, allowing us to use e.g. environment variables in scripts, and it'll work on Windows too.

What would make this PR ready for review
- Publishing process - I don't see why it wouldn't work, but I also cannot test it
- Analyzing dependency tree; would running builds in parallel be okay? If so, we can shave off additional 6 seconds from each build.


wojtekmaj:
@diegomura I'd love to know what you think about this!

diegomura:
I like this!

wojtekmaj:
I'm glad! I don't understand 100% of your workflow though, so I would really appreciate if you could play around with it and see if everything behaves 100% like you expect.

diegomura:
@wojtekmaj this needs rebase 😄 Sorry, lots of changes going on recently

wojtekmaj:
> lots of changes going on recently

I wonder who can be at fault 🤣

Just to be sure though, @diegomura did you find some time it works alright and doesn't disrupt your workflow?

I'm **more** than happy to address any feedback you might have, and this IS a big change!

wojtekmaj:
Regarding publishing process - my biggest worry was about how version descriptors are managed during publish. Now Yarn handles automatically what Changesets used to do - so we use `workspace:^` version descriptors that gets swapped to latest patch version on publish. This seems to be handled well since https://github.com/changesets/changesets/pull/585 so we should be okay!

PR #2570, opened 2024-01-26
chore: test against multiple versions of React

### LINKED ISSUE #2568: Test against multiple versions of React
As specified in peerDependencies, React 16, 17 and 18 should all work. We need to test that.

Issue on me, but can't self assign :)

wojtekmaj:
Closes #2568

Now the _real_ fun with GitHub Actions begins :D


wojtekmaj:
Node resolution fails with Yarn v1. This means #2567 would need to be merged first.

PR #2573, opened 2024-01-27
feat: add support for resolving blob images

### LINKED ISSUE #1884: Allow to use Blob as the Image source
On web contexts it is more handy to be able to use a Blob as the data source.
Even on modern versions of Node and other runtimes like Deno, have implemented this Web API.

A common use case would be taking an image from a canvas:
```jsx
<Image
   src={() => {
     const canvas = document.createElement('canvas')
     renderQR(canvas, "Hello world")
     // Now what?
   }}
/>
```

A naive approach would be creating a Data url with `canvas.toDataURL()`. But it is known to have some [drawbacks](https://stackoverflow.com/questions/59020799/which-function-should-i-use-todataurl-or-toblob).

So. Why don't create a Blob and use URL.createObjectUrl?
```jsx
<Image
  src={async () => {
    const canvas = document.createElement("canvas");
    renderQR(canvas, "Hello world");
    const blob = await new Promise((resolve, reject) =>
      canvas.toBlob((blob) => {
        if (blob) {
          return resolve(blob);
        } else {
          return reject(new Error("couldn't create Blob"));
        }
      })
    );
    return URL.createObjectUrl(blob);
  }}
/>
```

Now we have a worse problem. **There is a memory leak!**. Since we are not calling URL.revokeObjectUrl to release the resources. We could workaround it with a `useRef` and a `useEffect`, but the code becomes very verbose and hard to follow.

Allow to pass a Blob as the source for images, instead of only allowing Buffer, which is a Node-only API (we could polyfill it on other environments, but it just doesn't feels natural)

It would allow something like:
```jsx
<Image
  src={() => {
    const canvas = document.createElement("canvas");
    renderQR(canvas, "Hello world");
    return new Promise((resolve, reject) =>
      canvas.toBlob((blob) => {
        if (blob) {
          return resolve(blob);
        } else {
          return reject(new Error("couldn't create Blob"));
        }
      })
    );
  }}
/>
```

> As a sidenote I think this can be a starting point to reduce the dependencies on node specific APIs and replacing them by Web APIs, like Web Streams, the Fetch API, etc.

- Polyfilling Node buffer and convert the blob to that.
- Using URL.createObjectURL as pointed above.

**Additional context**

davbrito:
fixes #1884

### DIFF
--- a/packages/image/tests/resolve.test.js
+++ b/packages/image/tests/resolve.test.js
@@ -162,4 +162,33 @@ describe('image resolveImage', () => {
 
     expect(image1).not.toBe(image2);
   });
+
+  test('Should render a blob image', async () => {
+    const blob = new Blob([localJPGImage], { type: 'image/jpeg' });
+    const image = await resolveImage(blob);
+
+    expect(image.data).toBeTruthy();
+    expect(image.width).toBeGreaterThan(0);
+    expect(image.height).toBeGreaterThan(0);
+  });
+
+  test('Should render a blob whithout type', async () => {
+    const blob = new Blob([localJPGImage]);
+    const image = await resolveImage(blob);
+
+    expect(image.data).toBeTruthy();
+    expect(image.width).toBeGreaterThan(0);
+    expect(image.height).toBeGreaterThan(0);
+  });
+
+  test('Should render a blob image with type application/octet-stream', async () => {
+    const blob = new Blob([localJPGImage], {
+      type: 'application/octet-stream',
+    });
+    const image = await resolveImage(blob);
+
+    expect(image.data).toBeTruthy();
+    expect(image.width).toBeGreaterThan(0);
+    expect(image.height).toBeGreaterThan(0);
+  });
 });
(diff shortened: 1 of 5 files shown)


diegomura:
Thanks @davbrito ! Can you rebase? I'll take a look after but looking good

davbrito:
> Thanks @davbrito ! Can you rebase? I'll take a look after but looking good

Done!

packages/image/tests/resolve.test.js
```diff
@@ -162,4 +162,33 @@ describe('image resolveImage', () => {
...
+  test('Should render a blob image', async () => {
+    const blob = new Blob([localJPGImage], { type: 'image/jpeg' });
+    const image = await resolveImage(blob);
+
+    expect(image.data).toBeTruthy();
+    expect(image.width).toBeGreaterThan(0);
+    expect(image.height).toBeGreaterThan(0);
+  });
+
+  test('Should render a blob whithout type', async () => {
```
diegomura, line 175:
```suggestion
  test('Should render a blob without type', async () => {
```

diegomura (approved):
utACK

Thanks for this great contribution!

Issue #2460, opened 2023-12-10
I can't use renderToBuffer, renderToStream, etc ... in server side with NextJS 13+ with App router


mrkjofficial:
**Describe the bug**
In Next.js 13+ using App router , I'm unable to render PDF in Stream or Buffer in Next API routes 

**Error Message**
'ba.Component is not a constructor'

**To Reproduce**
Steps to reproduce the behavior including code snippet (if applies):
1. Create a new Next.js porject select App Router during setup.
2. Add @react-pdf/renderer in the project.
3. Create a dummy PDF.
4. Try to render the PDF in the server side.

**Expected behavior**
App should run, react-pdf should render pdfs.

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser Chrome
 - React-pdf version v3.1.14

jameschetwood:
Try using the pages router instead. This worked for me.

mrkjofficial:
> Try using the pages router instead. This worked for me.

No solutions for app router yet? My whole application is based on the App router.

Dru1DD:
Unfortunetly, you can you @joshuajaco/react-pdf-renderer-bundled, but if you using packages like @david.kucsai/react-pdf-table you would have a problem, that it doesn't work with App Router. On Client component yeah, but in server component when you tried to use renderToBuffer - unfortunetly - not(

bordeo:
I think you should resolve using   `transpilePackages` as I mention [here](https://github.com/diegomura/react-pdf/issues/2350#issuecomment-1871950404)

Dru1DD:
You can create pages/api - route even in App Router and it works

wojtekmaj:
Duplicate of #2350

airtonix:
> You can create pages/api - route even in App Router and it works

That's fine if you want to create a server side endpoint for use at runtime, but remember that just because someone is using the `apps/*` directory, doesn't mean that they are running nextjs server side at runtime.

https://nextjs.org/docs/pages/building-your-application/deploying/static-exports

diegomura:
Closing it as duplicated

Issue #2350, opened 2023-06-29
Unable to render in Next.js 13 app directory route handler


joshuajaco:
**Describe the bug**

Using `renderToBuffer` in a Next.js 13 app directory route handler throws: `TypeError: ba.Component is not a constructor`.
It works as expected when using a route handler inside of the pages directory instead.

<details>
<summary>full output</summary>

```
- warn ./node_modules/node-fetch/lib/index.js
Module not found: Can't resolve 'encoding' in '/Users/foo/react-pdf-nextjs/node_modules/node-fetch/lib'

Import trace for requested module:
./node_modules/node-fetch/lib/index.js
./node_modules/cross-fetch/dist/node-ponyfill.js
./node_modules/@react-pdf/font/lib/index.cjs.js
./node_modules/@react-pdf/renderer/lib/react-pdf.cjs.js
./src/app/api/app.pdf/route.tsx
- error TypeError: ba.Component is not a constructor
    at $$$reconciler (webpack-internal:///(sc_server)/./node_modules/@react-pdf/renderer/lib/react-pdf.cjs.js:631:50)
    at createRenderer (webpack-internal:///(sc_server)/./node_modules/@react-pdf/renderer/lib/react-pdf.cjs.js:3889:12)
    at pdf (webpack-internal:///(sc_server)/./node_modules/@react-pdf/renderer/lib/react-pdf.cjs.js:3997:28)
    at _callee$ (webpack-internal:///(sc_server)/./node_modules/@react-pdf/renderer/lib/react-pdf.cjs.js:4189:32)
    at tryCatch (webpack-internal:///(sc_server)/./node_modules/@babel/runtime/helpers/regeneratorRuntime.js:36:25)
    at Generator.eval (webpack-internal:///(sc_server)/./node_modules/@babel/runtime/helpers/regeneratorRuntime.js:120:30)
    at Generator.eval [as next] (webpack-internal:///(sc_server)/./node_modules/@babel/runtime/helpers/regeneratorRuntime.js:64:29)
    at asyncGeneratorStep (webpack-internal:///(sc_server)/./node_modules/@babel/runtime/helpers/asyncToGenerator.js:4:28)
    at _next (webpack-internal:///(sc_server)/./node_modules/@babel/runtime/helpers/asyncToGenerator.js:22:17)
    at eval (webpack-internal:///(sc_server)/./node_modules/@babel/runtime/helpers/asyncToGenerator.js:27:13)
    at new Promise (<anonymous>)
[...]

PR #2575, opened 2024-01-29
chore: test against multiple versions of React

wojtekmaj:
Gives us a lot more confidence, especially when e.g. upgrading react-reconciler. This, in turn, is probably going to be required if we ever gonna make a fix for #2350 and #2460 without going mentally insane.

Closes https://github.com/diegomura/react-pdf/issues/2568


diegomura (approved):
utACK

Nice! Thanks so much for this

PR #2578, opened 2024-01-29
chore: do not run Prettier on /lib/

wojtekmaj:
By ignoring /lib/, time it takes on my machine to run Prettier on the entire repo falls from 8.5 to 2.95s.


diegomura (approved):
utACK

Thanks!

PR #2579, opened 2024-01-29
chore: update react-reconciler to 0.26.0

wojtekmaj:
0.26.0 is the last version before introducing breaking changes we need to actually change something in the code for.

Now that #2568 is resolved, I'm more confident diving into this. Let's go step by step!


wojtekmaj:
Actually, not yet

PR #2582, opened 2024-01-31
feat: pass _INTERNAL__LAYOUT__DATA_ to onRenderCallback in toBuffer

SethThoburn:
This is already passed in toBlob, but the APIs aren't unified for other methods. I have the same use case as #2201, but I'm in Node not the browser.


diegomura (approved):
utACK

PR #2587, opened 2024-01-31
chore: add JSDoc types to @react-pdf/textkit

wojtekmaj:
This is a sample PR that improves type safety while keeping the code 100% in JavaScript. It's not fully functional (yet - unless #2588 gets merged), but gives an insight on how the codebase would look like with JS + JSDoc.

@diegomura, I would love to hear what you think. Should I continue?

### DIFF
--- a/packages/textkit/src/attributedString/advanceWidth.js
+++ b/packages/textkit/src/attributedString/advanceWidth.js
@@ -10,7 +10,9 @@ import runAdvanceWidth from '../run/advanceWidth';
  * @param {AttributedString} attributedString attributed string
  * @returns {number} advance width
  */
-export default function advanceWidth(attributedString) {
+const advanceWidth = (attributedString) => {
   const reducer = (acc, run) => acc + runAdvanceWidth(run);
   return attributedString.runs.reduce(reducer, 0);
-}
+};
+
+export default advanceWidth;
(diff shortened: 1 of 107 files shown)


diegomura:
@wojtekmaj passing by here, but couldn't read the code yet. I'll do on the weekend! Thanks again for all the effort and ideas you're bringing

diegomura:
@wojtekmaj needs rebase 😬

wojtekmaj:
Just 40 files, no problem :D

packages/textkit/src/attributedString/advanceWidth.js
```diff
@@ -1,14 +1,16 @@
...
+
 /**
  * Returns attributed string advancewidth
  *
- * @param {Object} attributedString attributed string
+ * @param {AttributedString} attributedString attributed string
  * @returns {number} advance width
  */
-const advanceWidth = (attributedString) => {
+export default function advanceWidth(attributedString) {
```
diegomura, line 13: Nit. Any reason to switch to functions? I kind of prefer arrow fns. Not important but curious :)

diegomura (approved):
utACK

Left a nit question, but looks good!

packages/textkit/src/attributedString/advanceWidth.js
wojtekmaj: Also personal preference (and affects bundle size a bit, but that's not really relevant for react-pdf I guess). Still, regardless of motives, that is out of scope :)

Issue #2574, opened 2024-01-28
Can't resolve 'fs' in jpeg-exif/lib


mishenThakshana:
I tried overidding `@react-pdf/image , @react-pdf/pdfkit ` it giving another error saying create is undefined.

mrzdtydlntm:
i have same issue can't resolve 'fs' with the error:

./node_modules/.pnpm/jpeg-exif@1.1.4/node_modules/jpeg-exif/lib/index.js
Module not found: Can't resolve 'fs'

It keep throwed to an error when I tried to build the image on docker

woramit:
I fixed it by this way
https://stackoverflow.com/questions/76005377/module-not-found-cant-resolve-fs-nextjs-nextauth

mrzdtydlntm:
> I fixed it by this way https://stackoverflow.com/questions/76005377/module-not-found-cant-resolve-fs-nextjs-nextauth

thanks!

mrzdtydlntm:
i fixed it using ticket #2535 because i build it using pnpm

Issue #2535, opened 2024-01-20
jpeg-exif: can't resolve 'fs'


arobert93:
**Describe the bug**

```
./node_modules/.pnpm/jpeg-exif@1.1.4/node_modules/jpeg-exif/lib/index.js:3:10
Module not found: Can't resolve 'fs'

https://nextjs.org/docs/messages/module-not-found

Import trace for requested module:
./node_modules/.pnpm/@react-pdf+pdfkit@3.1.1/node_modules/@react-pdf/pdfkit/lib/pdfkit.browser.js
./node_modules/.pnpm/@react-pdf+renderer@3.3.0_react@18.2.0/node_modules/@react-pdf/renderer/lib/react-pdf.browser.js
./src/components/core/pdf-viewer.tsx
```

**To Reproduce**

// page.tsx
```tsx
import * as React from 'react';

import { config } from '@/config';
import { PDFViewer } from '@/components/core/pdf-viewer';

export default function Page(): React.JSX.Element {
  return (
    <PDFViewer>
      {/* Document here */}
    </PDFViewer>
  );
}
```

/// pdf-viewer.tsx

```tsx
'use client';

import dynamic from 'next/dynamic';
import type ReactPDF from '@react-pdf/renderer';

export const PDFViewer = dynamic<ReactPDF.PDFViewerProps>(
  () => import('@react-pdf/renderer').then((m) => m.PDFViewer),
  {
    ssr: false,
    loading: () => null,
  }
);

```

**Expected behavior**
I expect it to not use any Node.js functions because if I remove the `use client` I get this error:

```
Error: PDFViewer is a web specific API. You're either using this component on Node, or your bundler is not loading react-pdf from the appropriate web build.
```

**Screenshots**

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser: -
 - React-pdf version: v3.3.0

genius-autos-admin:
I have the same issue. It only started today. I'm also using "@react-pdf/renderer": "3.1.9"

mcewball13:
I have the same issue that also started today or just super recently. When I try to build, I get that error.

mcewball13:
Yeah, I was trying to track down the dep that was just recently updated to see if I could find it, but I couldn't find it. It says the issue is with `jpeg-exif`, but that package hasn't been updated in a million years.

[...]

PR #2591, opened 2024-02-03
fix: replace jpeg-exif with jay-peg

diegomura:
Fixes: https://github.com/diegomura/react-pdf/issues/2535
Fixes: https://github.com/diegomura/react-pdf/issues/2574

Wrote a JPEG decoder [here](https://github.com/diegomura/jay-peg) that's very performant and leverages restructure which is already used in the project, so it should not add much bundle size (maybe it will even reduce it). This does more than getting EXIF data, so maybe in the future can be used to embed extra info of the image in the document. This should fix the node dep errors people are getting out of `jpeg-exif`

### DIFF
--- a/packages/image/package.json
+++ b/packages/image/package.json
@@ -33,7 +33,7 @@
     "@babel/runtime": "^7.20.13",
     "@react-pdf/png-js": "^2.3.0",
     "cross-fetch": "^3.1.5",
-    "jpeg-exif": "^1.1.4"
+    "jay-peg": "^1.0.0"
   },
   "files": [
     "lib"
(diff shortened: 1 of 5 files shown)


packages/examples/src/index.jsx
```diff
@@ -3,7 +3,7 @@ import { createRoot } from 'react-dom/client';
 
 import { PDFViewer } from '@react-pdf/renderer'
 
-import Document from './pageWrap'
+import Document from './jpgOrientation'
```
diegomura, line 6:
```suggestion
import Document from './pageWrap'
```

dmitryTurov:
https://github.com/diegomura/jay-peg/issues/4#issue-2147278184

PR #2604, opened 2024-02-06
fix: add key to pdf viewer iframe

haihuynhDF:
## What
Fix the browser history issue
https://stackoverflow.com/questions/45277276/react-router-and-iframe-browser-history
## Why
With React Router, NextJS..., when the PDF data changes and the user clicks to browser's `Back` button, it doesn't go to the previous page
## How
Add a `key` parameter to the PDFViewer's iframe

### DIFF
--- a/packages/renderer/src/dom/PDFViewer.js
+++ b/packages/renderer/src/dom/PDFViewer.js
@@ -29,6 +29,7 @@ export const PDFViewer = ({
       ref={innerRef}
       style={style}
       className={className}
+      key={crypto.randomUUID()}
       {...props}
     />
   );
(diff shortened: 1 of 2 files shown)


packages/renderer/src/dom/PDFViewer.js
```diff
@@ -29,6 +29,7 @@ export const PDFViewer = ({
       ref={innerRef}
       style={style}
       className={className}
+      key={crypto.randomUUID()}
```
diegomura, line 32: Why does it need to be here and not added by consumer? Not everyone uses react-router to see this issue. This could generate unwanted rerenders for other people right?

PR #2610, opened 2024-02-06
fix: CJS compatibility in @react-pdf/pdfkit

### LINKED ISSUE #2608: TypeError: PNG is not a constructor - Can't generate pdf with a PNG in the document
**Describe the bug**
After the latest version fix, my pdfs which were previously working fine now do not work with png images. The error is TypeError: PNG is not a constructor. 

**To Reproduce**
I am generating my pdfs on the server, which are sent via stream to the client. Images are included into the document with an Image tag:

` <Image
                        src={ "https:/cloudfrontlink.png" }
                    />`

**Expected behavior**
Ideally PDF documents would function the same as before the latest version fix

macbook pro m1
React-pdf/renderer version 3.3.7

wojtekmaj:
Missed that one in #2607!

Closes #2608

### DIFF
--- a/.changeset/silver-poems-look.md
+++ b/.changeset/silver-poems-look.md
@@ -1,0 +1,5 @@
+---
+"@react-pdf/pdfkit": patch
+---
+
+fix: fix CJS compatibility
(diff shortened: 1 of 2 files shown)


diegomura (approved):
utACK

Thanks!

PR #2618, opened 2024-02-07
chore: update husky to 9.0.0

wojtekmaj:
This update brings simplified config and increased compatibility with Windows.


diegomura (approved):
utACK

PR #2619, opened 2024-02-07
chore: use ESLint and Prettier before commit on JSX and TS files too

wojtekmaj:
Also sneaking in changes in .d.ts file, which previously was not auto-formatted on commit because of this very issue I'm fixing rn.


diegomura (approved):
utACK

PR #2605, opened 2024-02-06
fix: fix CJS compatibility

wojtekmaj:
Closes #2602
Closes #2603

Sorry about that!

We will also need some sanity checks to prevent this from happening in the future.

Issue #2602, opened 2024-02-06
v3.3.6 causes `TypeError: layoutEngine is not a function` in @react-pdf/renderer when using Next.js


EvHaus:
**Describe the bug**
After upgrading from v3.3.5 to v3.3.6, my Next.js app's `react-pdf` server-side generations are failing with:

```
TypeError: layoutEngine is not a function
    at Object.<anonymous> (/vercel/path0/node_modules/@react-pdf/layout/lib/index.cjs:308:16)
    at Module._compile (node:internal/modules/cjs/loader:1241:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1295:10)
    at Module.load (node:internal/modules/cjs/loader:1091:32)
    at Module._load (node:internal/modules/cjs/loader:938:12)
    at Module.require (node:internal/modules/cjs/loader:1115:19)
    at mod.require (/vercel/path0/node_modules/next/dist/server/require-hook.js:65:28)
    at require (node:internal/modules/helpers:130:18)
    at Object.<anonymous> (/vercel/path0/node_modules/@react-pdf/renderer/lib/react-pdf.cjs:14:22)
    at Module._compile (node:internal/modules/cjs/loader:1241:14)
```

Rolling back to v3.3.5 makes the issue go away.

**To Reproduce**

Minimal repro available here: https://stackblitz.com/edit/stackblitz-starters-rxmqpt?file=app%2Fapi%2Fbuild-pdf%2FMyDocument.tsx

**Expected behavior**
`react-pdf` should not crash.

**Screenshots**

![image](https://github.com/diegomura/react-pdf/assets/226640/7df0335d-e171-4df1-a2a5-7f2924307696)

**Desktop (please complete the following information):**

- OS: All
- Browser: Chrome
- React-pdf version: v3.3.6

iamMohammadAamir:
I am still getting error, even after update react-pdf/renderer 3.3.6

TypeError: Cannot read property 'Hidden' of undefined
web/node_modules/@react-pdf/layout/lib/index.js:2179:25

NCSmit:
Still getting this error after updating to 3.3.6. 

`"@react-pdf/renderer": "^3.3.6",`

TypeError: layoutEngine is not a function

glarivie:
Same error here, but I'm not using Next.js

diegomura:
Will take a look today

NCSmit:
Downgrading versions does not resolve the error on my end, anyone else ?

[...]

PR #2628, opened 2024-02-09
chore: add basic E2E tests

### LINKED ISSUE #2611: Implement tests testing bundled code
Issue on me, I can't assign myself.

See #2602, #2608 ➡️ These issues were absolutely preventable. To achieve this confidence level, we need to run basic tests on bundled code, on top of having unit tests for certain modules.

#2602 was preventable by a running simple command taking <1s to execute. See #2605.
For #2608, merely importing affected package is not enough, so something more E2E-like would be desirable.

Perhaps to do after #2567 as there's a lot of tests we would like to run fast, but in a topographical manner that Yarn offers?

wojtekmaj:
This proposal adds E2E tests. It runs these tests on Node.js 18, 20, and 21 (so just like unit tests), in CJS and ESM mode.

Each test creates a stupid simple and compares it to reference PDF file stored inside the repository. It's purposefully minimal: the aim of these E2E tests is to catch any bugs that would break module resolution on different Node.js environments, rather than test advanced react-pdf rendering features, which is beautifully done in unit tests already.

Would have prevented #2602, #2608. Here's what happens if I remove `interop: 'compat'` bit added in #2605 from @react-pdf/renderer's Rollup config:

```
% yarn --cwd e2e/node-esm start
yarn run v1.22.19
$ node ./index.js
✔ rendering a PDF (22.69725ms)
ℹ tests 1
ℹ suites 0
ℹ pass 1
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 27.5615
✨  Done in 0.28s.
```
⬆️  ESM is fine. But it was never broken in the first place.

```
% yarn --cwd e2e/node-cjs start
yarn run v1.22.19
$ node ./index.js
/react-pdf/packages/renderer/lib/react-pdf.cjs:4178
var fontStore = new FontStore();
                ^

TypeError: FontStore is not a constructor
    at Object.<anonymous> (/react-pdf/packages/renderer/lib/react-pdf.cjs:4178:17)
    at Module._compile (node:internal/modules/cjs/loader:1241:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1295:10)
    at Module.load (node:internal/modules/cjs/loader:1091:32)
    at Module._load (node:internal/modules/cjs/loader:938:12)
    at Module.require (node:internal/modules/cjs/loader:1115:19)
    at require (node:internal/modules/helpers:130:18)
    at Object.<anonymous> (/react-pdf/e2e/node-cjs/index.js:5:50)
    at Module._compile (node:internal/modules/cjs/loader:1241:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1295:10)

Node.js v20.9.0
error Command failed with exit code 1.
```
⬆️  Whooops there we go! Caught CJS setup error.

Will prevent any future issues with module resolution caused by e.g. dependency updates or Rollup config updates.

Closes #2611

PS. I'm considering another E2E test that would spin up Vite-based front-end application, opens it using Playwright and does a similar comparison, although that's a bit too much work for me at the moment.


diegomura (approved):
utACK

Nice! Thanks (again 😄 ) @wojtekmaj for all the hard work. I'll try to catch up during the weekend with all the threads you commented during the week 😄

PR #2631, opened 2024-02-10
fix: CI workflow error

wojtekmaj:
Fixes an issue where .toUpperCase() was used in an expression that's not JS, causing an error.

Please kindly run this workflow against this branch - no other way to test it, and I can't trigger it myself.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -108,7 +108,7 @@ jobs:
         run: REACT_VERSION=${{ matrix.react_version }} yarn test
 
   e2e-node:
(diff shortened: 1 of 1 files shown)


diegomura (approved):
utACK

Couldn't find a way to run it 😅 But change looks good. Let's get it in and see from there

wojtekmaj:
I think we're getting there but it might be one more issue: ternary being used.

Let's try with `cond && when_true || when_false` maybe?

https://docs.github.com/en/actions/learn-github-actions/expressions#example

PR #2635, opened 2024-02-12
fix: bump jay-peg dependency to fix CJS module resolution error

wojtekmaj:
jay-peg 1.0.0 suffered from an issue that caused module resolution error in CJS mode.


diegomura (approved):
utACK

diegomura:
Thanks! Forgot to bump it 🤦🏻

PR #2645, opened 2024-02-15
fix: rendering of \xc2\xa9 copyright symbol

mattreid1:
The current `specialCases` in `packages/layout/src/text/emoji.js` only ignores the copyright symbol `\xc2\xa9\xef\xb8\x8f` and not the standard UTF-8 copyright symbol `\xc2\xa9`. The former has the VS-16 variation selector appended to the standard symbol, but the latter still appears to be treated as an emoji and doesn't render.

This PR simply adds the standard copyright symbol `\xc2\xa9` to the `specialCases` array along with some comments explaining why there appears to be two copyright symbols.

Before the fix:
Before

After the fix:
After

### DIFF
--- a/packages/layout/src/text/emoji.js
+++ b/packages/layout/src/text/emoji.js
@@ -69,7 +69,12 @@ export const fetchEmojis = (string, source) => {
   return promises;
 };
 
-const specialCases = ['©️', '®', '™']; // Do not treat these as emojis if emoji not present
+/**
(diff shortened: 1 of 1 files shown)


wojtekmaj (approved):
LGTM!

Can you please run `yarn changeset` as explained in CONTRIBUTORS.md?

diegomura:
That logic changed. all symbols are being rendered now

PR #2672, opened 2024-03-12
fix: typo that breaks attempts of rendering non-visual elements like DEFS

somonek:
Rendering content that contains `<defs>` fails because `renderFns[node.type];` is undefined for nodeType DEFS, and the check is clearly wrong, since the renderFns is a declared object. Seems a typo.


wojtekmaj (approved):
LGTM. Can you please run `yarn changeset` as explained in CONTRIBUTORS.md?

diegomura:
Fixed in separate PR. Thanks for pushing this though!

PR #2689, opened 2024-03-25
fix: jpeg precision metadata

joelybahh:
Fix https://github.com/diegomura/react-pdf/issues/2625

Implementing the improved fix off of the back of oogas pull request, and information from this [comment thread.](https://github.com/diegomura/react-pdf/pull/2646)

- By dynamically adapting to the actual bit depth, we avoid hardcoding 8 bits (which would limit quality for higher bit-depth images).
- Avoids `this.bits` being undefined, which was problematic in more strict PDF viewers such as Adobe, Mac Preview and Safari.

I had the time and wanted to trial the open source flow anyway (Feel free to decline this if this is overstepping someone else's work I'm not trying to take credit or anything, I am simply looking to get this fix rolled out ASAP as we are using it in a company project).

Trying to be helpful in expediting is all.

### DIFF
--- a/.changeset/wicked-jeans-switch.md
+++ b/.changeset/wicked-jeans-switch.md
@@ -1,0 +1,5 @@
+---
+"@react-pdf/pdfkit": patch
+---
+
+fix: jpeg precision metadata
--- a/packages/pdfkit/src/image/jpeg.js
+++ b/packages/pdfkit/src/image/jpeg.js
@@ -26,6 +26,7 @@ class JPEG {
       }
 
       if (marker.name === 'SOF') {
+        this.bits ||= marker.precision;
         this.width ||= marker.width;
         this.height ||= marker.height;
         this.colorSpace ||= COLOR_SPACE_MAP[marker.numberOfComponents];


diegomura (approved):
utACK

diegomura:
Thanks!

joelybahh:
@diegomura I installed latest version of `@reactpdf/renderer` and noticed the lib for `pdfkit` isn't the same as the output I get locally for it. Is this simply because these changes aren't deployed yet?

If they are deployed, just a note that the build output from this change doesn't align to the expected one, the line change I added isn't present basically. 

Sorry if its just because deployment is bundled with other changes, just saw a release yesterday so thought it might have been this.

knuula:
I am wondering the same thing as @joelybahh above. I've been trying to deploy an update on this all day. It looks like npm is not updated with these changes yet. Anything you can do here @diegomura ?

joelybahh:
@knuula I noticed some E2E tests failed on the merge of this, maybe the tests are checking against PDF output that is now technically different due to the `jpeg` fixes, meaning a false negative. For me it had a syntax error on the github action `.yml` file. So I submitted a PR fixing that but that seems to have uncovered an entire e2e test can of worms.

joelybahh:
See this PR: https://github.com/diegomura/react-pdf/pull/2690

PR #2633, opened 2024-02-11
fix: CI workflow error & add separate reference files for different platforms

wojtekmaj:
Follow up after #2631 🫠

Fixes invalid syntax in workflow file.

Also, it turns out, generated PDFs differ slightly from platform to platform. As the previous reference file I generated comes from macOS, it fails when E2E is ran on linux. This fixes that by introducing platform-specific reference files.

PR #2690, opened 2024-03-27
Fixed `.yml` build error caused by ternary in script name for `e2e-node`

joelybahh:
Screenshot 2024-03-27 at 11 20 19 am
In response to the above error, which seems to have been triggered after my recent PR merge, I removed the ternary from the name. At a glance, this appears to be for vanity purposes anyway.

### DIFF
--- a/.changeset/great-badgers-pull.md
+++ b/.changeset/great-badgers-pull.md
@@ -1,6 +1,0 @@
----
-"@react-pdf/pdfkit": patch
-"@react-pdf/image": patch
----
-
-chore: bump jay-peg
--- a/.changeset/wicked-jeans-switch.md
+++ b/.changeset/wicked-jeans-switch.md
@@ -1,0 +1,5 @@
+---
+"@react-pdf/pdfkit": patch
+---
+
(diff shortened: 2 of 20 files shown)


joelybahh:
I thought this was a simple change because of small syntax error, looks like a can of worms I wasn't ready for 😅

joelybahh:
Could the E2E tests output be different now that this.bits is correctly being set?

ranma42:
> Could the E2E tests output be different now that this.bits is correctly being set?

That does not seem to be the cause, as there are no images involved in the test.
Additionally, in my environment the same diff is also detected running the tests on 6ffd1b96f5962f7abfb2e19d6249f1b5b519477e (right before the JPEG change). Finally, the very same issue seems to be present since the very beginning of the e2e tests 021c354ab834cfef30c3a95a618017f1d626ab87.

The difference seems to be in the encoding of the deflated stream (deflating the streams in reference.pdf and the test output using qpdf results in identical files, excepts for the "moving parts").

wojtekmaj:
Duplicate of #2633

Issue #2613, opened 2024-02-06
Use main pdfkit repo


diegomura:
I'd like to remove pdfkit fork and start using https://github.com/foliojs/pdfkit again. I branched from this a long time ago for multiple reasons, so I'm not sure the lift here. It will mean for sure to add several changes in that repo, which I think it could benefit the entire community, but not sure yet what those are. 

I'll keep this issue as a sort of umbrella issue where I'll capture all the differences and report progress as I tackle them. It will be a slow path.

wojtekmaj:
Hey, regarding png-js bit - consider pngjs too! See my comment: https://github.com/diegomura/react-pdf/issues/676#issuecomment-1930922597

diegomura:
Thanks! Just took a quick look, and pngjs has node deps too, meaning I'll need to fork it as for now, or have them onboard on shipping a browser build. And also convince pdfkit to adopt this. Because pdfkit already uses png-js the path towards unifying this is shorter, but I'll check what would be more convenient

kishaningithub:
@foliojs Can you kindly pitch in and add your thoughts here?

KhalidIrfani:
@diegomura Hi!
I am getting this error the @swc/helpers@0.4.11 with the react-pdf/render and i am using the next 13v
I do all the others way like cache clean , node module delete and reinstall and change the next version but didnt get any solution to resolve this issue
is there any way to resolved this issue

materio-mui-react-nextjs-admin-template@2.0.0 D:\alyzone\n-travel-scheduler\n_travel_web_dash
├─┬ @react-pdf/renderer@3.4.5
│ └─┬ @react-pdf/font@2.5.2
│ └─┬ fontkit@2.0.4
│ └── @swc/helpers@0.5.13
└─┬ next@13.0.0
└── @swc/helpers@0.4.11

error - ./node_modules/fontkit/dist/browser-module.mjs:2:0
Module not found: Can't resolve '@swc/helpers/_/_define_property'

Import trace for requested module:
./node_modules/@react-pdf/font/lib/index.browser.js
./node_modules/@react-pdf/renderer/lib/react-pdf.browser.es.js
./src/pages/to/accounting/invoices/list/index.js

https://nextjs.org/docs/messages/module-not-found

<img width="649" alt="save" src="https://github.com/user-attachments/assets/ad8871d7-82b5-45e7-8079-42368e21d668" />

Issue #672, opened 2019-07-29
Can I password protect a pdf generated using react-pdf?


lutherdexter:
I have a requirement to allow users to scrape contents of a page by exporting to pdf.  A must have feature is to give users the ability to password-protect the pdf at run-time. 

Is this feature available in react-pdf or do you know of third-party tools that I can use alongside your component?

diegomura:
It's not, and not sure if `pdfkit` supports it

mpopv:
It does support it. https://github.com/foliojs/pdfkit/blob/master/docs/getting_started.md#encryption-and-access-privileges

However I'm not sure if react-pdf is the right place to implement it, as the browser would prompt the user for a password on every rerender.

alexturpin:
I might be interested in taking this on at some point in the next year as I will need this for a project! For context, I am wanting to generate password protected PDFs client-side for download, not for in-browser display.

chetanbura:
+1 for the password-protected PDF feature.
This is a very useful feature.

Any plans to consider this feature?

suprMax:
Did you guys figure out a way to pass those options down to the pdfkit? The way I see it there is no way at the moment right?

Would be nice to allow for extra options/props to be just passed through to the pdfkit.

andreasss55:
Hi, I need this

PR #2696, opened 2024-03-29
Add encryption support and access privileges

klimeryk:
Hi, @diegomura! Love your library - I've been using it extensively at https://github.com/klimeryk/recalendar.js/ (https://recalendar.me/). I stumbled upon https://github.com/diegomura/react-pdf/issues/672 and figured it was a great chance to contribute back :)

I've tried my best to bring over the changes from the upstream pdfkit version. Encryption was originally added in https://github.com/foliojs/pdfkit/pull/820 there. It was mostly a straightforward update... except changes in `reference.js`. Looks like that class has been fundamentally changed - upstream it uses internally `Buffer`, while in react-pdf, the class extends `Writeable` instead.
I've tried first keeping the changes to `reference.js` minimal (see https://github.com/diegomura/react-pdf/pull/2696/commits/4376bcb805d0ce326d39ec393ab855ea8994ab11). Most of the changes were easy to port - basically copy and paste.

<del>The only part I did not know was the changes in `finalize` in `reference.js` [here](https://github.com/foliojs/pdfkit/blob/5635f8a02135acae1a6c1b904af55afe05d0ab7e/lib/reference.js#L52). As a result, this was a partial success - non-encrypted PDFs look as before (good), you can generate a PDF that requires a password, the password is properly validated... but then the opened PDF is missing content. See this file: [recalendar-password.pdf](https://github.com/diegomura/react-pdf/files/14809024/recalendar-password.pdf) (password: `secret`)</del>

<del>I've then tried just switching to upstream's `reference.js`, especially seeing your intention of doing that from https://github.com/diegomura/react-pdf/issues/2613. See https://github.com/diegomura/react-pdf/pull/2696/commits/5d9a8e9f720d55f36f3b65f2cac8484407b862fa. That looks more promising - the decrypted PDF has the text and almost looks perfect... except, for some reason, the font seems to be missing, so it looks off. See this file: [recalendar-password2.pdf](https://github.com/diegomura/react-pdf/files/14809017/recalendar-password2.pdf) (password: `secret`)
And this applies to both encrypted _and_ non-encrypted generated PDFs. So I'm guessing there's some modification/customization, maybe in some other file that I missed that was done in this version of pdfkit that is not compatible with the upstream version of `reference.js`? Unfortunately, I could not find it. I've tried also going back in history of `reference.js`, but unfortunately these changes seems to be from before it was moved to monorepo and I cannot find the previous repository 😞 
I figured I'd create this PR for you to have a look, maybe something will spark a quick solution or hint from you. If not, I can continue digging 🙇 </del>

Everything should be working smoothly now, see https://github.com/diegomura/react-pdf/issues/2924#issuecomment-2453621314 🎉 

### Testing
I'm guessing you have your preferred ways of testing different use cases :) Enabling password protection simply boils down to passing `userPassword` when creating a new instance of `PDFDocument` like so:
```
const ctx = new PDFDocument( {
    // ...
    userPassword: 'secret',
} );
```

To be on the safe side, I've also tried enabling all optional permissions, but that did not make a difference:
```
    permissions: {                                                                                                                                        
        printing: 'highResolution',
        modifying: true,
        copying: true,
        annotating: true,
        fillingForms: true,
        contentAccessibility: true,
        documentAssembly: true,
    },
```

Additionally, I've ported some unit tests from `pdfkit`, see https://github.com/diegomura/react-pdf/pull/2696/commits/798995ceabaf35f053a5f5e7645f417573273cad


packages/pdfkit/src/reference.js
```diff
@@ -1,91 +1,74 @@
...
-      this.chunks.length = 0; // free up memory
+      this.buffer = []; // free up memory
       this.document._write('\nendstream');
     }
 
     this.document._write('endobj');
-    return this.document._refEnd(this);
+    this.document._refEnd(this);
   }
-
```
wojtekmaj, line 88: Nit: we probably want to keep this space

packages/pdfkit/src/security.js
```diff
@@ -5,33 +5,557 @@
...
-
-  for (let i = 0; i < wordArray.sigBytes; i++) {
-    byteArray.push(
-      (wordArray.words[Math.floor(i / 4)] >> (8 * (3 - (i % 4)))) & 0xff
-    );
-  }
-
-  return Buffer.from(byteArray);
-};
+import CryptoJS from 'crypto-js';
```
wojtekmaj, line 8: I wonder how this change affects bundle size and what was the original reason for selective import done like that.

wojtekmaj (commented):
Woah, nice one! Looks good to me.

I wonder if we could have some tests for that. As long as #2633 repairs them 🥲 

Also, changeset is missing. Can you please run `yarn changeset` as explained in CONTRIBUTORS.md?

packages/pdfkit/src/security.js
klimeryk: Yeah, I tried digging in revision history, but it was introduced with the selective import in https://github.com/diegomura/react-pdf/pull/1894. Upstream never used the selective import and imported the whole library from the beginning: https://github.com/foliojs/pdfkit/pull/820. I'm guessing it's because the full/upstream `security.js` file uses extensively this library - `CryptoJS.MD5`, `CryptoJS.RC4`, `CryptoJS.lib.WordArray`, `CryptoJS.SHA256`, `CryptoJS.AES`. Whereas the simplified version (not supporting password protection) indeed just used `CryptoJS.MD5`, so the selective import might have made sense for that use case.

For completeness sake (so that the implications of this PR on the bundle size are known):

#### Current `master`
```
646,480 pdfkit.browser.cjs
644,160 pdfkit.browser.js
326,577 pdfkit.browser.min.cjs
325,782 pdfkit.browser.min.js
407,839 pdfkit.cjs
405,488 pdfkit.js
232,858 pdfkit.min.cjs
231,996 pdfkit.min.js
```

#### This branch/PR
```
683,846 pdfkit.browser.cjs
680,853 pdfkit.browser.js
343,476 pdfkit.browser.min.cjs
342,353 pdfkit.browser.min.js
445,176 pdfkit.cjs
442,171 pdfkit.js
249,962 pdfkit.min.cjs
248,604 pdfkit.min.js
```

So, 9% increase for `pdfkit.js` and 7.2% for `pdfkit.min.js`.

packages/pdfkit/src/reference.js
klimeryk: Makes sense, I missed this during merge - I've restored it 🙇

klimeryk:
Thanks for looking at the PR, @wojtekmaj! Sorry for the late follow-up - private life got in the way, but now I'm back and determined to see this PR through. I've addressed your comments and generated the changeset 🙇 

To be extra clear - **I don't think this PR is ready to merge as-is**. At the very least there's an issue with the fonts. There's links to generated PDF files in the description and my attempt at best describing the possible issue, but it might be best shown on screenshots:

#### Production file, correct fonts

Screenshot 2024-06-14 at 12 10 04

#### PR file, missing fonts
Screenshot 2024-06-14 at 12 10 12

And there are indeed warnings from the Firefox internal PDF previewer:
```
Warning: getFontFileType: Unable to detect correct font file Type/Subtype.
Warning: FormatError: Required "loca" table is not found
```

My guess is that these attachments/resources are not correctly encoded or decoded with the passwords/security logic introduced in this PR. I was hoping I'd get some pointers from @diegomura or anyone else familiar with PDF internals 🙏  Though, if not possible (I understand that we're all busy one way or another), I'll try my best to dig into - it's an interesting problem and I love understanding how things work. Just I might not be as effective without expert help 🙇

klimeryk:
PR updated - it should be good for a final/proper review. The font issues have been solved thanks to @florianbepunkt, see https://github.com/diegomura/react-pdf/issues/2924#issuecomment-2453621314. I've also ported some unit tests from `pdfkit`, see https://github.com/diegomura/react-pdf/commit/798995ceabaf35f053a5f5e7645f417573273cad. Only ported the ones that were passing and trivial to move over. More work needs to be done there, since this version of pdfkit is a subset of upstream. But it will be helpful in the efforts in https://github.com/diegomura/react-pdf/issues/2613

kev2480:
I'd love to see this in the next release! Any updates? 🙏

sengitu-florio:
Is there any work required to make this ready for release? I'd be happy to utilize this `userPassword`. I'd be happy to contribute if additional work is required.

andreasss55:
Hi, any timeline when this is going to be released?

klimeryk:
Looks like password support got just merged: https://github.com/diegomura/react-pdf/pull/3344, so I'll close this PR 🙇

diegomura:
Ouch. So sorry @klimeryk , i did not see this. I got claude to work on it and I didn't check open PRs. I'm so sorry again

klimeryk:
@diegomura, no worries, totally understand - ultimately, I'm just glad to have the password support in! I'll be sure to test it out when it's released 💖

Issue #2499, opened 2024-01-16
Yoga-layout v2 leads to failing builds


takethefake:
**Describe the bug**
I migrated today from `3.1.14` to `3.1.17` and received the following error while deploying.

```
+ yarn run deploy
/azp/_work/1/s/node_modules/yoga-layout/src/generated/YGEnums.ts:10
export enum Align {
^^^^^^

SyntaxError: Unexpected token 'export'
    at internalCompileFunction (node:internal/vm:73:18)
    at wrapSafe (node:internal/modules/cjs/loader:1274:20)
    at Module._compile (node:internal/modules/cjs/loader:1320:27)
    at Module._extensions..js (node:internal/modules/cjs/loader:1414:10)
    at Object.require.extensions.<computed> [as .ts] (/azp/_work/1/s/node_modules/ts-node/src/index.ts:1608:43)
    at Module.load (node:internal/modules/cjs/loader:1197:32)
    at Function.Module._load (node:internal/modules/cjs/loader:1013:12)
    at Module.require (node:internal/modules/cjs/loader:1225:19)
    at require (node:internal/modules/helpers:177:18)
    at Object.<anonymous> (/azp/_work/1/s/node_modules/yoga-layout/src/wrapAssembly.js:10:1)
```

this seems to be related to the migration to `yoga-layout@2.01`. I for now downgraded to `@react-pdf/layout:3.6.4` via resolutions.

jjose230391:
Hello, I'm experiencing an issue when installing "@react-pdf/renderer": "3.1.17" and "@react-pdf/layout": "3.6.4". It's showing the error "Module not found: Can't resolve 'yoga-layout/sync'." I'm using Node version 12.

kxc184:
I am also experiencing similar issue in Jest when running my unit test on node 14.12.0.

`Cannot find module 'yoga-layout/sync' from 'node_modules/@react-pdf/layout/lib/index.cjs.js'`

I agree with OP [ Upgraded yoga to v2 and removed bundled package](https://github.com/diegomura/react-pdf/pull/2473/files#diff-e15fac666d27e15671fdcb60f01cf492d6b095a52f672943639c436192078174) seems to be the issue.

Adding overrides to [@react-pdf/layout:3.6.4](https://www.npmjs.com/package/@react-pdf/layout/v/3.6.4?activeTab=versions) (version before problem commit) temporarily resolves issue.

diegomura:
[...]

PR #2711, opened 2024-04-10
feat!: update Yoga to 3.0

### LINKED ISSUE #2507: [solution] __dirname is not defined in ES module scope
**Describe the bug**

When building react-pdf in ESM mode, you may encounter an error:

```
__dirname is not defined in ES module scope
```

This is because yoga-layout uses `__dirname` in their code. This has already been reported to them: https://github.com/facebook/yoga/issues/1559

To work around this issue, for now, consider using `inject` option in esbuild:

`cjs-shim.ts`
```ts
import { createRequire } from 'node:module';
import path from 'node:path';
import url from 'node:url';

globalThis.require = createRequire(import.meta.url);
globalThis.__filename = url.fileURLToPath(import.meta.url);
globalThis.__dirname = path.dirname(__filename);
```

`esbuild.ts`
```ts
await esbuild.build({
  // …
  inject: ['cjs-shim.ts'],
  // …
});
```

If you're not using esbuild, you may manually import the `cjs-shim.ts` file BEFORE react-pdf is loaded.

### LINKED ISSUE #2692: Update Yoga to 3.0
Updating Yoga to 3.0 will bring several features into React-PDF:

- Support for position: static
- Support for align-content: space-evenly
- Improvements to layout correctness
- Yoga’s JavaScript bindings are now distributed as an ES Module
- Fixes several crashes in Java bindings
- Some existing Yoga APIs have been removed

Source: https://www.yogalayout.dev/blog/announcing-yoga-3.0

But also, it will:

- Resolve https://github.com/diegomura/react-pdf/issues/2507, bringing in true, zero configuration ESM compatibility 

Potential breaking changes:

- Yoga 3.0 is distributed as ESM-only. We might be forced to make React-PDF ESM-only as well.
- Given that there are "layout correctness" improvements makes me believe this should be considered a breaking change, as people may have used certain workarounds that would stop working after an update.

wojtekmaj:
Closes #2692
Closes #2507

This PR updates Yoga to 3.0.

See the list of new features, improvements, and potential breaking changes in #2692.


diegomura:
@wojtekmaj this looks good. What's missing?

wojtekmaj:
@diegomura See the original ticket: we will be unable to provide CJS builds as Yoga 3 is a pure ESM module.

diegomura:
Sorry! Missed that. Huh, good decision to make. Would be nice to push forward that and just embrace ESM but I'm afraid that would break a lot of setups. Will think about this. Any input would be welcomed!

wojtekmaj:
I think it's really up to you: how much further are you able (and willing to!) push the v2 forward? If you think that significant improvements can still be made, you may want to hold off with v3 for a while, to avoid maintaining two releases at the same time. If you think v2 is "pretty much ready" and you're basically only _maintaining_ it, then it would be best to bite the bullet and remove CJS builds.

Please note that removing CJS builds does not stop CJS users from:
- Using v2 🤷 They clearly like outdated solutions anyway
- Importing ESM modules into CJS asynchronously (although that's a bit of a burden)
- Using Node with `--experimental-require-module` flag which _may_ just work
- Using a bundler like Rollup before running code in Node to benefit from their built-in ESM/CJS interop to avoid all module resolution issues

karlhorky:
I wonder if Yoga v2 with its unusual CommonJS wrapper code is also causing issues like this property access error `Yoga.Overflow.Hidden` that we've been experiencing:

- https://github.com/diegomura/react-pdf/issues/2499#issuecomment-2166770560

karlhorky:
@wojtekmaj is there an easy way to try this PR out?

If this isn't published somewhere already, would it be possible to get a prerelease version published?

I'm currently patching `@react-pdf/layout` in `node_modules` and using pnpm Overrides to force `yoga-layout@3.0.4`, and it appears to be working, but it's a bit brittle...

wojtekmaj:
I think this is the best we can do until it's released :(

packages/layout/src/yoga/index.js
```diff
@@ -1,14 +1,13 @@
 /* eslint-disable import/prefer-default-export */
 
-import * as Yoga from 'yoga-layout';
+import Yoga from 'yoga-layout';
 
 export const loadYoga = async () => {
```
michaelgmcd, line 5: Looked a little bit and it has some trickle-down changes needed, but this doesn't need to be `async` anymore, correct?

wojtekmaj: Good eye. Thanks!

michaelgmcd:
It may actually be more beneficial to go back to keeping things async and using the `yoga-layout/load` export. The default one requires top-level await (may break some users):
https://github.com/facebook/yoga/tree/main/javascript#requirements

wojtekmaj:
Thanks for the heads up. PDF.js introduced top-level await recently and they rolled it back after a huge backslash, so I guess you'd be right.

seanroddy-fbg:
Hi, is there an expected release data for this feature?

diegomura:
I'm very tempted on removing CJS support. We should look forward. I feel it's kind of our responsibility to help things moving forward as lib authors. Newer solutions won't become adopted if people do not have an incentive to upgrade. Also, given bandwidth of lib authors feels like it simplifies things. @wojtekmaj curious what you think

wojtekmaj:
I can't agree more. And frankly, it's not even a question if we want to move react-pdf forward. More and more libs are abandoning CJS, like our core dependencies.

diegomura:
Amazing. I'll remove CJS support and we can land this after then!

diegomura:
Done in https://github.com/diegomura/react-pdf/pull/2871

diegomura (approved):
utACK

Issue #2700, opened 2024-04-03
Bookmarks all link to last page in document


danomatic:
**Describe the bug**
All bookmarks link to the very last page

**To Reproduce**
1. Open repl link:
https://react-pdf.org/repl?code=3187b0760ce02e00408a057025803c450298c0bc300500943807cf805030c00f0022230080b61985311655400a02180e658212005e19b00220032014400aace90095c4c0046204006b463c013a689bc04c008ce3da54bd5646146c8d6135403d0dbb172cb871fa839843442465e49455d4b475f50df8b00099cc38aca8ddec6261625c527da99dbd13b8d202c4a4e41594d435b4f40dc4fc019812ad38b3eb336cd91272f3399ce8199959d80801b8c8c914307980a0b86800c400e97458004c3174f0a91151d0b19d8846c88000
2. Download file
3. Open file
4. Open table of contents
5. Click bookmark
6. Notice that all bookmarks jump to the last page

**Expected behavior**
Bookmarks should link to the page that they were placed on

**Screenshots**
<img width="1062" alt="Screenshot 2024-04-02 at 6 04 03 PM" src="https://github.com/diegomura/react-pdf/assets/459337/245a46b2-7830-483c-9b2f-59fef6139490">

**Desktop (please complete the following information):**

- OS: MacOS
- Browser: Any
- React-pdf version: v3.4.2

danomatic:
The issue goes away if I revert to `v2.3.0`. All `^v3` versions that I tried had the bug.

danomatic:
What's so strange is this seemed to work with other v3 versions, but after about 24 hours of total dependency hell and battling with installing different version, for the life of me I can't find one that doesn't have broken bookmarks. This is breaking a production application in a crucial way. Any help our guidance would be appreciated.

asthaawasthi-laitkor:
> The issue goes away if I revert to `v2.3.0`. All `^v3` versions that I tried had the bug.

Hi @danomatic which node version you are working with.

hormiai76:
I have the same problem. I'm using node v18.14.0

danomatic:
For now I'm avoiding the issue by pinning `@react-pdf/pdfkit` to the last working version in yarn:

```
  "resolutions": {
    "@react-pdf/pdfkit": "3.1.0"
  }
```

sbuys:
Anyone know a version 3 release that works?

sbuys:
[...]

Issue #2682, opened 2024-03-21
Outline Links link to the page instead of the actual element


typesafeui:
**Describe the bug**
When using bookmarks the generated outlines link to the page containing the bookmarked element instead of the element it self causing the page to zoom out completely in Firefox (changing the zoom value of the user) while in chrome it just redirects to the top left of the  page

PS. Might have to do with this line of code: https://github.com/diegomura/react-pdf/blob/1f8952e6bc137e02eaf96486559571a8622943b6/packages/pdfkit/src/outline.js#L28
**To Reproduce**
1. create bookmarks int he pdf
2. render as pdf
3. click one of the outlines
4. See that the link does not work

**Expected behavior**
The user zoom doesn't change and it redirects to the right element

**Screenshots**
N/A

**Desktop (please complete the following information):**

- OS: Windows
- Browser Chrome
- React-pdf version 3.4.0

PR #2715, opened 2024-04-12
Add back functionality for bookmark props

mgmolisani:
This restores functionality that seemed to have accidentally erased in a [minor version update](https://github.com/diegomura/react-pdf/pull/2532/commits/5139b942eec7674a97195f90b04f4badeeae283d#diff-b715dda1c414782f098d4d383a10d77477178e2f015970666e180a0c20b0da44) to converge the forked `pdfkit` vendor package. This resulted in all bookmarks being created in `fit` mode on the last page not respecting any of the props.

I additionally added some comments to roughly show where the custom logic of this file exists to give some signal that this should not be altered in future fork maintenance.

I tested this locally in a simple repro and observed both the raw output and the PDF in Adobe Reader were again properly linking to the page and respecting props.

There are no unit tests for the custom logic in this vendor package hence I did not add any new ones.

Resolves https://github.com/diegomura/react-pdf/issues/2700

### DIFF
--- a/packages/pdfkit/src/mixins/outline.js
+++ b/packages/pdfkit/src/mixins/outline.js
@@ -9,7 +9,9 @@ export default {
     this.outline.endOutline();
     if (this.outline.children.length > 0) {
       this._root.data.Outlines = this.outline.dictionary;
-      return (this._root.data.PageMode = 'UseOutlines');
+      /* Custom fork start */
+      return (this._root.data.PageMode = this._root.data.PageMode || 'UseOutlines');
+      /* Custom fork end */
     }
   }
 };
--- a/packages/pdfkit/src/outline.js
+++ b/packages/pdfkit/src/outline.js
@@ -1,11 +1,32 @@
+/* Custom fork start */
+const DEFAULT_OPTIONS = {
+  top: 0,
+  left: 0,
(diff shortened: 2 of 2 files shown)


typesafeui:
Nice! this probably closes #2682 too 🙌

AgustinBilldin:
pls merge

EvHaus:
Confirmed this fixes #2700.

typesafeui:
@diegomura @wojtekmaj can we merge this?

diegomura (commented):
Thanks for pushing this. I really need to start using the base pdfkit thought 😞 Do you think you can try finding a way to fix this on react-pdf side?

danomatic:
Polite bump. I've been pinned to an old version of `@react-pdf/pdfkit` for months to avoid the bugs this resolves.

mgmolisani:
> Thanks for pushing this. I really need to start using the base pdfkit thought 😞 Do you think you can try finding a way to fix this on react-pdf side?

I think since this is a regression from the original behavior in this lib and is blocking consumers from updating, its probably best to get this merged and then consider making changes to the vendor source package. I think blocking new features might be appropriate to steer the ship in that direction but fixing the regression should be priority #1.

PR #2640, opened 2024-02-14
feat: add fallback font support

nikgraf:
Allows to add multiple fallback fonts and will fall back to the next font and eventually last font in case the glyph is not available.

I added an example for testing:

Screenshot 2024-02-14 at 17 31 59

Happy to update the PR in case something is missing e.g. test. Just need to know what and where.

PR #2723, opened 2024-04-24
Bump types package

hendrikmolder:
We updated the `fontFamily` types in #2640 but forgot to bump the version of the types package.


diegomura (approved):
utACK

wojtekmaj:
This fix was incorrect, but #2727 fixes it for me.

PR #2727, opened 2024-04-25
fix: font fallbacks types

hendrikmolder:
## Context

#2640 Added support for font fallbacks, but we updated the types incorrectly.

`FontSource` should still accept just a `string`.

However, in stylesheets we should be able to specify an array of font families (as shown in the font fallbacks example [here](https://github.com/diegomura/react-pdf/blob/6bf9d990723db757232c42e3a040bbfaa5ceb308/packages/examples/src/fontFamilyFallback/index.jsx#L18)).

## Changes

- Update `FontSource` and `Style` type declarations accordingly
- Bump `types` version

### DIFF
--- a/.changeset/rare-moles-raise.md
+++ b/.changeset/rare-moles-raise.md
@@ -1,0 +1,5 @@
+---
+"@react-pdf/types": minor
+---
+
+accept string array for `fontFamily` in stylesheets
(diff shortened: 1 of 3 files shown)


wojtekmaj (approved):
Works for me!

git-ekuo:
@hendrikmolder @wojtekmaj , thanks for the work! May I know if this would be merged soon? I've been looking forward to using the font fallback feature :)

.changeset/rare-moles-raise.md
```diff
@@ -0,0 +1,5 @@
+---
+"@react-pdf/types": minor
```
diegomura, line 2:
```suggestion
"@react-pdf/types": patch
```

Issue #2730, opened 2024-04-26
3.4.4 font changes no longer apply


Mudbill:
**Describe the bug**

I updated the library from 3.4.2 to 3.4.4 and `fontFamily` no longer applies. I see that [there are font related changes](https://github.com/diegomura/react-pdf/pull/2640) in these updates, so perhaps it is related to that.

**To Reproduce**

```jsx
<View>
  <Text>This is normal text</Text>
  <Text style={{ fontFamily: "Helvetica-Bold" }}>This should be bold</Text>
  <Text style={{ fontWeight: 700 }}>This should probably also be bold</Text>
  <Text style={{ fontFamily: "Courier" }}>This should use the Courier font</Text>
</View>
```

All of the above appear as the same, default Helvetica 400.

**Expected behavior**

Font changes should apply correctly.

Note: Using `fontWeight` before 3.4.2 didn't work either. I was only able to set bold font using Helvetica-Bold font family. Maybe Helvetica-Bold is not registered as a bold variant of Helvetica, but rather its own family?

**Screenshots**

![image](https://github.com/diegomura/react-pdf/assets/8832665/13cde087-9f20-4b1e-8f08-10f76d41e76c)

**Desktop (please complete the following information):**

- OS: macOS
- Browser: Chrome
- React-pdf version: 3.4.4

bdkopen:
@Mudbill I located the issue and opened #2731 to address the font family regression. This does not include the `fontWeight` issue introduced prior to 3.4.2.

joacub:
this also breaks registering fonts like SFMono-Semibold.ttf, it never loads pdf get stuck rendering

bdkopen:
> this also breaks registering fonts like SFMono-Semibold.ttf, it never loads pdf get stuck rendering

@joacub Can you provide an example that works in 3.4.2 but does not work in 3.4.4? This may not be a related issue. I was able to load a SFMono-Semibold font in both 3.4.2 and 3.4.4 fine. Note that I found many SFMono files are an "otf" extension, not "ttf"  like you wrote in your message.

DaveFPath:
Last week I updated to 3.4.4, and I got a report from a client saying that all the bold text in all the pdf's were no longer bold. Upon reading up on this thread and others, I believe my problem is the same as this one.

```
  companyName: {
    fontWeight: "bold",
    fontFamily: "Helvetica-Bold",
  },
```

[...]

PR #2731, opened 2024-04-27
fix: available Font Family Regression

bdkopen:
Changes in #2640 caused a regression by not providing proper support for [font families made available within the library](https://react-pdf.org/fonts#register). This has caused any font families that are built-in to the library to default to Helvetica.

# Changes
- Properly handles `run.attributes.font` as an array so any strings properly`getOrCreateFont` their associated fonts.
- Update the test page to include fonts made available within the library.

# Related
Issue: #2730

# Example
Before the change:
image

After the change:
image

### DIFF
--- a/packages/examples/src/fontFamilyFallback/index.jsx
+++ b/packages/examples/src/fontFamilyFallback/index.jsx
@@ -16,7 +16,12 @@ const styles = StyleSheet.create({
   },
   regular: {
     fontFamily: ['Roboto', 'NotoSansArabic'],
+    fontSize: '14',
     fontWeight: 900,
+  },
+  default: {
+    fontFamily: ['Courier-Bold', 'NotoSansArabic'],
+    fontSize: '14',
   },
 });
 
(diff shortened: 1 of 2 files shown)


bdkopen:
@diegomura tagging to notify that this fixes a recent regression. Please let me know if you need anything else with this. Thanks!

diegomura (approved):
utACK

coren-frankel:
My project has v3.4.4 and I just dug through my node modules to find that these changes aren't in the latest release. Please halp!

bdkopen:
> My project has v3.4.4 and I just dug through my node modules to find that these changes aren't in the latest release. Please halp!

@coren-frankel these changes are merged, but they won't be available until @diegomura releases the next version.

EvHaus:
FYI -- this change causes the following failure for me if `font-family` is a string, rather than array:

> run.attributes.font.map is not a function. (In 'run.attributes.font.map((font) => typeof font === \"string\" ? getOrCreateFont(font) : font)', 'run.attributes.font.map' is undefined)","stack":"Error: PDF creation failed. Please contact us!

PR #2736, opened 2024-05-01
feat: Use monochrome emojis contained in the font family if no emoji source available

siketyan:
This pull request proposes to add fallback to monochrome emojis if no emoji image source available. Though react-pdf does not support colour emojis yet, it is useful for monochrome or grayscale documents if we can fallback to the monochrome emojis.

### DIFF
--- a/packages/layout/src/text/fontSubstitution.js
+++ b/packages/layout/src/text/fontSubstitution.js
@@ -60,7 +60,7 @@ const fontSubstitution =
         break;
       }
 
-      const chars = string.slice(run.start, run.end);
+      const chars = [...string.slice(run.start, run.end)];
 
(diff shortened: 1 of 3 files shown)


packages/layout/src/text/fontSubstitution.js
```diff
@@ -60,7 +60,7 @@ const fontSubstitution =
         break;
       }
 
-      const chars = string.slice(run.start, run.end);
+      const chars = [...string.slice(run.start, run.end)];
```
diegomura, line 63:
```suggestion
      const chars = string.slice(run.start, run.end);
```

diegomura (approved):
tACK

PR #2794, opened 2024-06-16
fix undefined glyph getting id

farhoudi:
i'm not js developer but i'm getting this error when i try to create bunch of pdfs from our tauri application.
image
image
it happens only on a few items, not all.
but this change fixes that.


wildseansy:
I also started seeing this issue today with my project.  Please merge this fix soon!

diegomura:
Can you provide a way to reproduce it to understand why it's returning undefined here?

PR #2802, opened 2024-06-19
docs: remove lerna bootstrap command from setup steps

coren-frankel:
While following the `CONTRIBUTING.md` setup instructions, I ran into a warning upon running the `yarn bootstrap` command after `yarn install`:

```bash
$ lerna bootstrap
ERR! bootstrap The "bootstrap" command was removed by default in v7, and is no longer maintained.
ERR! bootstrap Learn more about this change at https://lerna.js.org/docs/legacy-package-management
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

The [Lerna docs](https://lerna.js.org/docs/legacy-package-management#replacing-your-usage-of-lerna-bootstraplerna-link) say:
> Replace lerna bootstrap with npm install (or yarn/pnpm). If you are already performing your package manager's install command somewhere in your workflow before where you had previously called lerna bootstrap, then you can just delete it instead.

I removed the `yarn bootstrap` command from the setup instructions and the `bootstrap` script from `package.json`. I also corrected some typos and grammar and made punctuation more consistent.


diegomura (approved):
utACK

Issue #2813, opened 2024-07-10
Non-deterministic `Document` IDs on subsequent renders


markusylisiurunen:
**Describe the bug**
I encountered an issue with non-deterministic document IDs when trying to write a few snapshot test cases for a project I was working on. I tried setting every prop for both `<Document />` (incl. `creationDate` and `modificationDate`) and `<Page />`. The document's ID is still different on every render.

**To Reproduce**
Try to render any document twice with any content and compare the raw outputs. You will find that the `/ID` line (e.g. `/ID [<a072362df31f6eb25904f2f2fc2166d8> <a072362df31f6eb25904f2f2fc2166d8>]`) differs on subsequent renders.

**Expected behavior**
I would expect the ID to stay the same given the same props.

**Desktop:**

- OS: macOS
- `react-pdf` version: 3.4.4

diegomura:
You can now accomplish deterministic ids by setting `creationDate`

PR #2814, opened 2024-07-10
Generate the document's ID on demand instead of the class constructor

markusylisiurunen:
Solves #2813.

Referring to #2813, subsequent `<Document />` renders produces a PDF file with non-deterministic `/ID` even though the `creationDate` and `modificationDate` props were set. If I managed to understand correctly, the ID is generated when the `PDFDocument` class was instantiated. However, some of the `.info` object's fields were updated from outside the class instance during the rendering process:

```js
const setPDFMetadata = (target) => (key, value) => {
  if (value) target.info[key] = value;
};

const addMetadata = (ctx, doc) => {
  const setProp = setPDFMetadata(ctx);

  // ...
  const creationDate = props.creationDate || new Date();
  const modificationDate = props.modificationDate || null;

  // ...
  setProp('CreationDate', creationDate);
  setProp('ModificationDate', modificationDate);
};
```

However, the ID was generated based on the `.info` in the constructor and was not updated after the `.info` object was updated. This means that the ID is generated with the default `CreationDate` which is the current time.

```js
// Initialize the metadata
this.info = {
  Producer: 'PDFKit',
  Creator: 'PDFKit',
  CreationDate: new Date()
};

if (this.options.info) {
  for (let key in this.options.info) {
    const val = this.options.info[key];
    this.info[key] = val;
  }
}

this._id = PDFSecurity.generateFileID(this.info);
```

This PR changes the behaviour so that the ID is not generated in the constructor but rather only when it is needed. Therefore, the modifications in the `.info` object should already be made at that point.

### DIFF
--- a/packages/pdfkit/src/document.js
+++ b/packages/pdfkit/src/document.js
@@ -115,9 +115,6 @@ class PDFDocument extends stream.Readable {
       });
     }
 
-    // Generate file ID
-    this._id = PDFSecurity.generateFileID(this.info);
-
     // Initialize security settings
     // this._security = PDFSecurity.create(this, options);
 
@@ -131,6 +128,10 @@ class PDFDocument extends stream.Readable {
     if (this.options.autoFirstPage !== false) {
       this.addPage();
     }
+  }
+
+  _id() {
+    return PDFSecurity.generateFileID(this.info);
   }
 
   addPage(options) {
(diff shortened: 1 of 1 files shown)


diegomura:
I was trying to unify our pdfkit fork with the real package, and this introduces a new change. Can we do this on react-pdf side by overriding the trailer?

Issue #2659, opened 2024-02-27
Infinite loop hanging megathread (minPresenceAhead, margin, break)


tom2strobl:
# Description

1) There has to be a general flaw in the layout logic resulting in infinite loops that crash browser engines _(and whatever the issue might be it is evident that crashes are the worst expression of failure handling)_
2) (maybe related or not) minPresenceAhead works on the simplest examples but fails to work for me in any real-life scenarios where nested Views (eg inevitable for rendering bullet lists) and page padding (necessary for my case) are used

Note that I use 3.3.8 which afaik should already include https://github.com/diegomura/react-pdf/pull/2400

## Preliminary work

Regarding 1)

- https://github.com/diegomura/react-pdf/issues/1384 is related, simply nothing really happened since 2021
- https://github.com/diegomura/react-pdf/issues/2083 was closed (because OP gone rogue) but seems to be related
- https://github.com/diegomura/react-pdf/issues/2303 was closed but seems to be related
- https://github.com/diegomura/react-pdf/issues/2242 might be related
- https://github.com/diegomura/react-pdf/issues/874 might have been related

Regarding 2)

- https://github.com/diegomura/react-pdf/issues/2595 absolutely is related and showcases it pretty well
- https://github.com/diegomura/react-pdf/issues/955 seems it doesn't work for a lot of folks for a very long time

## Reproducable cases
_(note I'm using @jeetiss REPL for hanging issues since you obviously can't share an url of the official REPL if the process hangs)_

[...]

PR #2822, opened 2024-07-18
fix: prevent infinite loop while splitting pages

mantljosh:
The existing check for whether all children got moved the next page makes sense if we're already partly through the page - but if this happens before we have anything in `currentChildren` then we'll essentially leave a blank page and move everything to the "next" page, which results in an infinite loop (since the next page never gets smaller).

This PR checks for this situation, and in this case will add the current node to the current page and push the remaining nodes to the next page. This ensures that the next page will get smaller by at least one node so that future iteration won't infinitely loop.

I added a test case that reproduces it - the test doesn't fail it will just hang forever, but if it doesn't loop then the test will pass successfully.

Fixes https://github.com/diegomura/react-pdf/issues/2659

### DIFF
--- a/.changeset/fair-rocks-leave.md
+++ b/.changeset/fair-rocks-leave.md
@@ -1,0 +1,5 @@
+---
+"@react-pdf/layout": patch
+---
+
+fix: prevent infinite loop while splitting pages
--- a/packages/layout/src/steps/resolvePagination.js
+++ b/packages/layout/src/steps/resolvePagination.js
@@ -100,13 +100,19 @@ const splitNodes = (height, contentArea, nodes) => {
 
       // All children are moved to the next page, it doesn't make sense to show the parent on the current page
       if (child.children.length > 0 && currentChild.children.length === 0) {
-        const box = Object.assign({}, child.box, {
-          top: child.box.top - height,
-        });
-        const next = Object.assign({}, child, { box });
+        // But if the current page is empty then we can just include the parent on the current page
+        if (currentChildren.length === 0) {
+          currentChildren.push(child, ...futureFixedNodes);
+          nextChildren.push(...futureNodes);
+        } else {
+          const box = Object.assign({}, child.box, {
+            top: child.box.top - height,
+          });
+          const next = Object.assign({}, child, { box });
 
-        currentChildren.push(...futureFixedNodes);
-        nextChildren.push(next, ...futureNodes);
+          currentChildren.push(...futureFixedNodes);
+          nextChildren.push(next, ...futureNodes);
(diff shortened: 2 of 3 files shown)


carlobeltrame (approved):
Change seems correct to me (who wrote the current splitting algorithm), thanks for fixing this! I don't have time to test it manually, unfortunately.

diegomura:
Thanks for this. Unfortunately I have very little time lately and can't test right now. Can someone validate it? 🙏🏻

mantljosh:
@diegomura can this be merged/released?

ArturCzopek:
Would be awesome to release because this is probably the reason why our scenario does not work as well and page is crashing all the time

ArturCzopek:
@mantljosh could you help install your version locally and make it as a permanent dependency? I am not strict front-end dev but I hope that your fix could solve lots of issues we have

BasitAli:
I am still running into an issue where it's stuck in an infinite loop here. Still investigating, but maybe there could be a hard break after certain runs or something so that it doesn't just hang.

ArturCzopek:
For me, wrapping content of component by empty react tags `<>CONTENT</>` or some child by it fix all issues with looping, no idea why but try it

BasitAli:
Yeah, probably some edge case. For me it was a new section in a report I was generating. Moved the section to a new page to resolve. But the data in the report is dynamic, so it's very possible it could happen again. We definitely need some sort of maximum depth check to break out of the loop and fallback to some default behavior for these cases.

diegomura:
This is included yes. Please do!

thedivac:
My dynamically created document is a lot more complex, but this simple case crashes the browser if you copy it into the playground at https://react-pdf.org/repl

```[javascript]
const Quixote = () => (
  <Document>
    <Page style={styles.body}>
      <View
        style={{
          border: 1,
          height: 700,
          marginBottom: 100,
        }}
      >
        <Text style={{ background: "yellow" }}>Test</Text>
      </View>
      <View
        style={{
          border: 1,
          height: 700,
          marginBottom: 100,
        }}
      >
        <Text style={{ background: "yellow" }}>Test</Text>
      </View>
     
    </Page>
  </Document>
);
```

1Jesper1:
@thedivac Your example still let the browser crash. It it the same problem as this?  https://github.com/diegomura/react-pdf/issues/2884#issuecomment-2391880139

DeviousM:
I don't think it's solved. @thedivac's example still causes a freeze/crash, and so does it in my specific, much more complex case. Has anyone found a solution for it yet?

PR #2823, opened 2024-07-18
fix: rollup config

maximuk:
There is an issue with the latest version in my project (Next.js app):
image

After debugging I came to consideration that the issue appears because of rollup config and `module` variable overwrite in `packages/renderer/lib/react-pdf.browser.js` file:
image

This overwrite breaks HMR work and crashes all the page.


diegomura:
Packages no longer has `module`. This should be fixed in next release

Issue #2818, opened 2024-07-15
Cannot render standard fonts locally (Vite, React, TS)


mrjeanp:
**Describe the bug**
Standard fonts aren't being rendered using `fontFamily` on a Vite/React/TS project.

**To Reproduce**

1. Set `fontFamily: "Times-Roman"`  on `<Text />` component. 

Attached screenshots show that loading the fonts from the public folders (with different name) works but directly referencing "Times-Roman" does not render the font. Using the REPL works though, am I missing something?.

**Expected behavior**
To correctly render standard font that is pre-installed in the system.

**Screenshots**
<img width="415" alt="image" src="https://github.com/user-attachments/assets/24b3693f-3b33-4c3f-a22e-a2abd12ce236">
<img width="595" alt="image" src="https://github.com/user-attachments/assets/03bd9148-d315-4498-b771-b06db6d1106f">
<img width="864" alt="image" src="https://github.com/user-attachments/assets/1aa63741-f85d-41f3-adbc-c2629cd47e61">
<img width="344" alt="image" src="https://github.com/user-attachments/assets/7b96fbd9-630c-4016-881e-127b2ac64a05">

**Desktop (please complete the following information):**

- OS: [MacOS]
- Browser [chrome]
- React-pdf version [^3.4.4]

knuula:
I am having the same issue here. It is specifically popping up with <Text> elements that are inside of <View> elements that are "inline-flex", "inline", or "inline-block"

knuula:
some more context on this --- the non-standard fonts are rendering in some situations.

if the parent <View> element is a "flex" for example, the non-standard font is rendering correctly.

if the parent <View> element is a "inline-flex", the non-standard font will default to some basic font or not show up at all.

mrjeanp:
I cloned the react-pdf-site repo and the REPL page has the same problem on a localhost. I don't understand why, 

localhost:3000/repl
<img width="1728" alt="image" src="https://github.com/user-attachments/assets/88fe301c-269b-4678-aadc-b7d60a8757c3">

react-pdf.org/repl:
<img width="1728" alt="image" src="https://github.com/user-attachments/assets/0b51aab2-0ef3-4296-b9cd-8e5811abf7f3">

mrjeanp:
Submitted a PR to fix this https://github.com/diegomura/react-pdf/pull/2827

PR #2827, opened 2024-07-22
fix: standard fonts not rendering

mrjeanp:
This is a follow up fix to my issue https://github.com/diegomura/react-pdf/issues/2818

Problem:
When styling `<Text/>` with standard `fontFamily` such font doesn't render and instead it falls back to `Helvetica` which is the currently configured default font.

Fix:
I copied the following code and included a condition inside `pickFontFromFontStack` to force selecting the font that the user actually picked.
```js
var standard = ['Courier', 'Courier-Bold', 'Courier-Oblique', 'Courier-BoldOblique', 'Helvetica', 'Helvetica-Bold', 'Helvetica-Oblique', 'Helvetica-BoldOblique', 'Times-Roman', 'Times-Bold', 'Times-Italic', 'Times-BoldItalic'];
```

Please review @diegomura 🙏

### DIFF
--- a/packages/layout/src/text/fontSubstitution.js
+++ b/packages/layout/src/text/fontSubstitution.js
@@ -6,6 +6,8 @@ const fontCache = {};
 
 const IGNORED_CODE_POINTS = [173];
 
+const standard = ['Courier', 'Courier-Bold', 'Courier-Oblique', 'Courier-BoldOblique', 'Helvetica', 'Helvetica-Bold', 'Helvetica-Oblique', 'Helvetica-BoldOblique', 'Times-Roman', 'Times-Bold', 'Times-Italic', 'Times-BoldItalic'];
+
 const getFontSize = (node) => node.attributes.fontSize || 12;
 
 const getOrCreateFont = (name) => {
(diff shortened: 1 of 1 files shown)


theskillwithin:
Will this also fix custom fonts loaded via `Font.register`?

mrjeanp:
@theskillwithin i never experienced issues with `Font.register` so i never attempted to fix it, this will most likely not fix your issue. Out of curiosity, what is your setup?

theskillwithin:
> @theskillwithin i never experienced issues with `Font.register` so i never attempted to fix it, this will most likely not fix your issue. Out of curiosity, what is your setup?

@mrjeanp  For some reason the Font.register only works with external fonts.  

as soon as i try to use an internal font by module loading or pointing to relative path `/fonts/foo.ttf` it does not work

oddly if I point to the same font like: `https://mysite.com/fonts/foo.ttf` this does work

using Remix (loaded via vite)

mrjeanp:
Taken from https://react-pdf.org/fonts#register
image

Try using an absolute path, maybe use something like `path.join(__dirname, '../fonts/foo.ttf')`

PR #2835, opened 2024-08-01
Allow src and href attributes to also be numbers

karlwithak:
In the `link` annotation, we can see that numbers are accepted to allow linking directly to page numbers.

https://github.com/diegomura/react-pdf/blob/3f5bf67d84ee784b6685cf603f04ac5f568a146d/packages/pdfkit/src/mixins/annotations.js#L57

This PR updates the type definitions of the `src` and `href` attribute on the `Link` component to allow passing in page numbers.


diegomura:
You linked a pdfkit line, but that does not mean `@react-pdf/renderer` allows passing numbers as src or href. In which case we would want this?

PR #2836, opened 2024-08-01
fix: check SVG node type is not currently supported

obi-awyss:
The check tested the existence of `renderFns` instead of `renderFn`.


diegomura (approved):
utACK

PR #2013, opened 2022-09-08
Adding Form Annotation support

axel7083:
Since pdfkit support Form Annotation I added the logics to support them in react-pdf. I've never contributed to this project, so I am not aware if I was suppose to add this feature in the way I did. I also added the types support for TS.

See issue https://github.com/diegomura/react-pdf/issues/42

Here is an example
```Javascript
<Document>
    <Page>
      <View style={{backgroundColor: 'rgba(182,28,28,0.62)', width: '30%', height: '100%'}}>
          <Form>
              <FormField name="user-info" style={{flexDirection: 'column'}}>
                  <Text>FormText</Text>
                  <FormText name={'name'} value={"hello"} align={'center'} style={{height: '50px'}} password />

                  <Text>FormCombo</Text>
                  <FormCombo
                      name={'combo'}
                      select={['', 'option 1', 'option 2']}
                      value={''}
                      defaultValue={''}
                      style={{height: '20px'}}
                  />

                  <Text>FormList</Text>
                  <FormList name={'list'} select={['', 'option 1', 'option 2']} value={''} defaultValue={''} style={{height: '50px'}}/>

                  <Text>FormPushButton</Text>
                  <FormPushButton name={'bouton'} label={'push button'} style={{height: '50px'}}/>
              </FormField>
          </Form>
      </View>
    </Page>
  </Document>
```

And the result [example.pdf](https://github.com/diegomura/react-pdf/files/9530609/example.pdf)

PR #2845, opened 2024-08-18
feat: added Form Annotation support

natterstefan:
## Tasks

- [x] fix dependency cycle (also mentioned [here](https://github.com/diegomura/react-pdf/pull/2013#discussion_r1014974134)). @diegomura: I am not sure how though... #need-help
- [ ] discuss component names (see https://github.com/diegomura/react-pdf/pull/2845#discussion_r1721040015, and https://github.com/diegomura/react-pdf/pull/2845#discussion_r1783965707)
- [x] fix macOS (https://github.com/diegomura/react-pdf/pull/2845#discussion_r1734780553)

## Related PR

- https://github.com/diegomura/react-pdf-site/pull/145

## Notes

- based on the September 2022 PR <https://github.com/diegomura/react-pdf/pull/2013> initially started by [@axel7083](https://github.com/axel7083)
- used [@runelk](https://github.com/runelk)'s [branch](https://github.com/runelk/react-pdf) as the base branch for this one
- applied `initForm` fix by [@kjossendal](https://github.com/kjossendal), see https://github.com/diegomura/react-pdf/pull/2013#issuecomment-1998708270

## Docs

- <https://pdfkit.org/docs/forms.html>

## Example PDF

[Example 2.0.pdf](https://github.com/user-attachments/files/16785205/Example.2.0.pdf)

## Example Package (install via GitHub)

-> <https://github.com/traveltechdeluxe/react-pdf/pull/2>


packages/pdfkit/src/mixins/acroform.js
```diff
@@ -46,6 +46,10 @@ export default {
    * this method to set the default font.
    */
   initForm() {
+    if (!!this._root.data.AcroForm) {
+      // Form is already initialized
+      return this;
+    }
```
natterstefan, line 52: Fixes an issue when using `<Form />` twice on e.g. two pages.

packages/primitives/src/index.js
```diff
@@ -8,6 +8,12 @@ export const Note = 'NOTE';
 export const Path = 'PATH';
 export const Rect = 'RECT';
 export const Line = 'LINE';
+export const Form = 'FORM';
+export const FormField = 'FORM_FIELD';
+export const TextInput = 'TEXT_INPUT';
+export const FormPushButton = 'FORM_PUSH_BUTTON';
+export const Picker = 'PICKER';
+export const FormList = 'FORM_LIST';
```
natterstefan, line 16: What do you think of [keeping the initial names of these components](https://github.com/diegomura/react-pdf/pull/2013/files#r1014982991) to keep them aligned with the form annotation methods of `PDFKit`?

```
formText( name, x, y, width, height, options)
formPushButton( name, x, y, width, height, name, options)
formCombo( name, x, y, width, height, options)
formList( name, x, y, width, height, options)
```
([src](https://pdfkit.org/docs/forms.html#form_annotation_methods))

packages/render/src/primitives/renderNode.js
```diff
@@ -12,8 +12,18 @@ import setLink from '../operations/setLink';
...
+import renderTextInput from './form/renderTextInput';
+import renderFormPushButton from './form/renderFormPushButton';
+import renderPicker from './form/renderPicker';
+import renderFormList from './form/renderFormList';
+
+const isRecursiveNode = (node) =>
+  node.type !== P.Text &&
+  node.type !== P.Svg &&
+  node.type !== P.Form &&
+  node.type !== P.FormField;
```
natterstefan, line 26: I ran into the following error when at least `node.type !== P.FormField`  is not applied: https://github.com/traveltechdeluxe/react-pdf/blob/81b75d26cf57e307256147372718ab8246676af6/packages/render/src/primitives/form/renderTextInput.js#L9-L12

packages/render/src/primitives/renderNode.js
```diff
@@ -12,8 +12,18 @@ import setLink from '../operations/setLink';
 import clipNode from '../operations/clipNode';
 import transform from '../operations/transform';
 import setDestination from '../operations/setDestination';
-
-const isRecursiveNode = (node) => node.type !== P.Text && node.type !== P.Svg;
+import renderForm from './form/renderForm';
+import renderFormField from './form/renderFormField';
```
natterstefan, line 16: TODO

- [ ] fix `Dependency cycle`

packages/examples/src/form/index.jsx
```diff
@@ -0,0 +1,105 @@
...
+                value="bar"
+                align="center"
+                style={{ height: '50px' }}
+                password
+              />
+            </View>
+
+            <Text>Checkbox</Text>
+            {/* TODO: how to check automatically? */}
+            <Checkbox name="checkbox" style={{ height: '20px' }} />
```
natterstefan, line 50: - [ ] example: how to set `checked`?

packages/pdfkit/src/mixins/acroform.js
PhilippBloss: Why not check this inside the render package itself? 
That way, it would not interfere with this issue: https://github.com/diegomura/react-pdf/issues/2613

PhilippBloss: And what about removing the `<Form />` component? Its sole purpose is initializing on the document. Shouldn't it be a property on the `<Document />` then, or maybe even just a check inside the form components?

packages/render/src/primitives/renderNode.js
PhilippBloss: This could be solved by introducing a cleanUp function similar to the render function in the renderNode.js, which would be called after render the component and childs. This way a form field would not need to render its own childs. 

```suggestion
const isRecursiveNode = (node) =>
  node.type !== P.Text &&
  node.type !== P.Svg &&
  node.type !== P.Form;

const cleanUpFns = {
  [P.FormField]: cleanUpFormField
}

// inside renderNode():

  const renderFn = renderFns[node.type];

  if (renderFn) renderFn(ctx, node, options);

  if (shouldRenderChildren) renderChildren(ctx, node, options);

  const cleanUpFn = cleanUpFns[node.type];

  if (cleanUpFn) cleanUpFn(ctx, node, options);

// renderFormField.js:

const renderFormField = (ctx, node, options) => {
  const name = node.props?.name || '';
  
  const formField = ctx.formField(name);
  const option = options;
  if(!option.formFields)
    option.formFields = [formField]
  else
  option.formFields.push(formField)
};

const cleanUpFormField = (_ctx, _node, options) => {
  options.formFields.pop()
}
```

PhilippBloss: On the form text it would then need something like this: `options.formFields?.at(0)`

PhilippBloss:
Right now, push buttons don't have any purpose. They are only there to exist. They cannot hold any value. There should be an option for the user to specify an `8.6.4` action (from [PDF Reference 1.7](https://opensource.adobe.com/dc-acrobat-sdk-docs/pdfstandards/pdfreference1.7old.pdf)) for this button. Even if it's just on a basic level. Going too detailed e.g. with `8.5.3` actions would go beyond the scope of the pr. What do you think?

packages/examples/src/form/index.jsx
natterstefan: Thank you! This works like a charm. Thank you for sharing the code! I'd have no idea how to do this myself currently.

packages/pdfkit/src/mixins/acroform.js
natterstefan: Hey @PhilippBloss, thanks for the input. I did not find the place where I could add this to `<Document />`. But I added it to form components instead now: https://github.com/diegomura/react-pdf/pull/2845/commits/7162e6f4b1bb01f72c26a0a052ef4ac76e8b7971.

packages/examples/src/form/index.jsx
PhilippBloss: Updated the code for an appearance with paths.
I don't own a Mac. Would be thankful if you could try this one and give feedback on it.

natterstefan: > Updated the code for an appearance with paths. I don't own a Mac. Would be thankful if you could try this one and give feedback on it.

👋  I think we are getting there, but not yet.

Web:
image

macOS:

image

PhilippBloss: While tinkering around I noticed my mistake on the first try. I wanted to use the [Zapf Dingbats](https://en.wikipedia.org/wiki/Zapf_Dingbats)-Font for the marks, but put neither a ref nor the resource into the code. I tried to fix that one, because it should be the cleaner one of the two.

By removing `NeedAppearances` from pdfkit I could recreate an error state with the first one and at least could verify that my fix worked on chrome on Windows. From the recreation I would conclude that the apple standard PDF viewer doesn't support `NeedAppearances` which would make it even more important to work right. Thanks for trying out my solutions.

Sidenote: I had the idea to put an option for xmarks in. So that it shows checkmarks default, but the user can decide to change it. I also fixed two points where you accidentally compared a string with a boolean via triple equals which will always result in a false expression.

packages/render/src/primitives/form/renderCheckbox.js
```diff
@@ -0,0 +1,29 @@
+import { parseCheckboxOptions } from '../../utils/parseFormOptions';
+
+const renderCheckbox = (ctx, node, options) => {
+  const { top, left, width, height } = node.box || {};
+
+  // Element's name
+  const name = node.props?.name || '';
+  const formFieldOptions = options.formFields?.at(0);
+
+  if (!formFieldOptions)
```
PhilippBloss, line 10: Checkboxes and TextFields don't need a FormField to them. Would be recommended but not necessary.

packages/render/src/primitives/form/renderTextInput.js
```diff
@@ -0,0 +1,29 @@
+import { parseTextFieldOptions } from '../../utils/parseFormOptions';
+
+const renderTextInput = (ctx, node, options) => {
+  const { top, left, width, height } = node.box || {};
+
+  // Element's name
+  const name = node.props?.name || '';
+  const formFieldOptions = options.formFields?.at(0);
+
+  if (!formFieldOptions)
```
PhilippBloss, line 10: Checkboxes and TextFields don't need a FormField to them. Would be recommended but not necessary.

packages/renderer/index.d.ts
```diff
@@ -224,6 +224,75 @@ declare namespace ReactPDF {
...
+    align?: string;
+    multiline?: boolean;
+    password?: boolean;
+    noSpell?: boolean;
+    format?: TextFieldFormatting;
+  }
+
+  class TextInput extends React.Component<TextInputProps> {}
+
+  interface CheckboxProps extends FormCommonProps, FormCommonProps {
```
PhilippBloss, line 274: Import Twice is not needed here

packages/examples/src/form/index.jsx
natterstefan: Hi @PhilippBloss, thanks for your help. Without your support, this would not work! 💯 

It works now on Web (Screenshot 1) and macOS (Screenshot 2):

### Web
image

### macOS
image

natterstefan:
Hi @diegomura, what do you think about this feature request?

packages/primitives/src/index.js
```diff
@@ -8,6 +8,11 @@ export const Note = 'NOTE';
 export const Path = 'PATH';
 export const Rect = 'RECT';
 export const Line = 'LINE';
+export const FormField = 'FORM_FIELD';
```
diegomura, line 11: Not a strong opinion, but would `Form` be more semantic? For `Field` I imagine like an input, but this is actually a form wrapper

diegomura (commented):
I love this. Thanks so much for putting this together. I'd like to get this merged soon. 

Left a small comment about primitive naming.

Also I think we should add docs in the site (https://github.com/diegomura/react-pdf-site) after this lands. Can you take care of them? 🙏🏻 Not a blocker to merge this once we settle on namings

packages/primitives/src/index.js
natterstefan: Hi @diegomura,

thanks for your response. Let's see what I suggested here in the past: https://github.com/diegomura/react-pdf/pull/2845#discussion_r1721040015. 🤔 

My initial thought was to align the names with the ones used in `pdfkit` ([here](https://github.com/foliojs/pdfkit/blob/831179e0674076a7b7775991620f26065967b250/docs/forms.md#form-annotation-methods)). So `TextInput` becomes `FormText`, `Picker` becomes `FormPicker`, and so on. But you made a valid point [back in 2022](https://github.com/diegomura/react-pdf/pull/2013/files#r1014982991) suggesting we should stick to the native web primitives. I got used to the current names while preparing the PR.

Regarding `FormField`: I chose the same name as `pdfkit` mainly to align our (react-pdf) and their [(pdfkit) docs](https://github.com/foliojs/pdfkit/blob/831179e0674076a7b7775991620f26065967b250/docs/forms.md#field-names) and keep them similar. I don't know if that's practical.

PhilippBloss: What about something like `FormGroup`? Form rather sounds like the single and complete set of inputs, but they rather are a subsection. E.g., you could separate billing address and shipping address into two `groups`.

natterstefan: Hi @PhilippBloss,

your suggestion makes sense to me, considering what pdfkit states in their docs as well:

> Using the formField method you might create a shipping field that is added to the root of the document, an address field that refers to the shipping field as it's parent, and a street Form Annotation that would refer to the address field as it's parent.

What do you think of `fieldset`(<https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset>):

> The <fieldset> HTML element is used to group several controls as well as labels (<label>) within a web form.

packages/renderer/index.d.ts
```diff
@@ -224,6 +224,76 @@ declare namespace ReactPDF {
...
+    nDec?: number;
+    sepComma?: boolean;
+    negStyle?: 'MinusBlack' | 'Red' | 'ParensBlack' | 'ParensRed';
+    currency?: string;
+    currencyPrepend?: boolean;
+  }
+
+  // see http://pdfkit.org/docs/forms.html#text_field_formatting
+  interface TextInputProps extends FormCommonProps {
+    align?: string;
```
PhilippBloss, line 265: This can only be either left, center or right. But it's just a small thing

packages/renderer/index.d.ts
```diff
@@ -224,6 +224,76 @@ declare namespace ReactPDF {
...
+  interface FormFieldProps extends NodeProps {
+    name: string;
+  }
+
+  class FormField extends React.Component<
+    React.PropsWithChildren<FormFieldProps>
+  > {}
+
+  // see http://pdfkit.org/docs/forms.html#text_field_formatting
+  interface TextFieldFormatting {
```
PhilippBloss, line 245: Might adjust this interface name as well

packages/primitives/src/index.js
PhilippBloss: Sounds good as well. Would give the decision to @diegomura

PhilippBloss: Would take over the docs, except @natterstefan has already started with those

packages/renderer/index.d.ts
```diff
@@ -224,6 +224,76 @@ declare namespace ReactPDF {
...
+    onState?: string;
+    offState?: string;
+    xMark?: boolean;
+  }
+
+  class Checkbox extends React.Component<CheckboxProps> {}
+
+  interface PickerAndListProps extends FormCommonProps {
+    sort?: boolean;
+    edit?: boolean;
```
PhilippBloss, line 287: Edit and noSpell are combo only

packages/renderer/index.d.ts
```diff
@@ -224,6 +224,76 @@ declare namespace ReactPDF {
...
+    multiline?: boolean;
+    password?: boolean;
+    noSpell?: boolean;
+    format?: TextFieldFormatting;
+  }
+
+  class TextInput extends React.Component<TextInputProps> {}
+
+  interface CheckboxProps extends FormCommonProps {
+    backGroundColor?: string;
```
PhilippBloss, line 275: The `g` would need to be lower case

GauravB159:
Would love to see this get merged! I'd love to contribute as well if there's a blocker

diegomura:
@PhilippBloss thanks for the review here. Would love to take this to the finish line. Are you still open to take care of the docs? 🙏🏻

packages/primitives/src/index.js
natterstefan: > Would take over the docs, except @natterstefan has already started with those

Hi @PhilippBloss, I haven't taken care of the docs yet. You can take over the docs if you want to or we can share the task. It's fine for me.

Also thanks for the review. I will try to apply your suggestions tomorrow or at least in the next few days.

Thanks for the ping @diegomura.

PhilippBloss:
> @PhilippBloss thanks for the review here. Would love to take this to the finish line. Are you still open to take care of the docs? 🙏🏻

Initially I wanted to wait for a decision on the component names, but I made a todo: https://github.com/diegomura/react-pdf-site/pull/145

natterstefan:
Hi @diegomura,
Hi @PhilippBloss,

I just made the branch compatible with 4.x and took care of @PhilippBloss' feedback. The only that is left now is deciding which names we choose for the components:

### Option 1 - align with pdfkit

- `<FormField />`
- `<FormText />`
- `<FormCombo />`
- `<FormCheckbox />`
- `<FormList />`

### Option 2 - web standards

- `<Fieldset />`
- `<InputText />` (not 100% aligned as it would be `<input type="text" />`)
- `<Select />`
- `<Checkbox />` (not 100% aligned as it would be `<input type="checkbox" />`)
- `<List />` (also not 100% aligned but there are `ol` and `ul` lists)

I favour Option 1 more as there is a connect to the used library used for forms. What do you think?

(previous discussion: https://github.com/diegomura/react-pdf/pull/2845#discussion_r1721040015, and https://github.com/diegomura/react-pdf/pull/2845#discussion_r1783965707)

natterstefan:
Hey @diegomura, Hey @PhilippBloss,

can you take a look at my last thoughts please? Thank you!

natterstefan:
_\*\*Reminder\*\*_

Hey @diegomura, Hey @PhilippBloss,

can you take a look at my last thoughts please? Thank you!

diegomura:
Hi @natterstefan   

Sorry for my slow response here. I took a long vacation on the last couple of weeks. 

Thanks for listing both options. In my opinion something like option 2 is what we need. Being coherent with pdfkit is not really important, as pdfkit is an implementation detail and not something users are really aware of. 

React-pdf always followed react-native primitives, so if an element is already defined there we should use the same naming (ex. TextInput instead of InputText or FormText)

natterstefan:
Hi @diegomura,

no worries. I hope you had a wonderful time. :)

This seems reasonable and let's follow that approach here as well.

It works for text input, but what about other components of this PR?

Checkbox is not part of react native anymore: https://reactnative.dev/docs/checkbox, also forms in general.

diegomura:
Thanks @natterstefan !

Here are my thoughts:

- `FormField`: Why do we need this one? It's not semantic nor clear what this does for the user. Pdfkit docs says that to express hierarchy a `name` can be passed to each form element
- `TextInput`: Like the react-native primitive
- `Select`: not in primitives but good name
- `Checkbox`:  not in primitives but good name
- `List`: Can be merged with Select maybe, and add a type="list" prop? I feel they are basically the same element, it's just rendered differently

natterstefan:
Hi @diegomura,

let's see what I can add to our discussion.

> * `FormField`: Why do we need this one? It's not semantic nor clear what this does for the user. Pdfkit docs says that to express hierarchy a `name` can be passed to each form element

That is true, it's usefulness is "limited" (possibly an uninformed opinion) to creating a hierarchy of nodes as this can be achieved with `name` only as well. Setting defaults (such as the default font) for all child form annotations ([docs](https://github.com/foliojs/pdfkit/blob/master/docs/forms.md)) is another feature. I understand that you question its purpose in the react-pdf context. Maybe if we can answer this you can make the final decision: Does `FormField` help creating a cleaner, more hierarchical PDF? If one prefers another component layer instead of using `name` only then yes. Does it help to simplify shared styling/attributes? Yes, but one can achieve this also by sharing `props` or `styles` (_correct?_). So what about removing it and if people request it add it (again) in another release?

> * `TextInput`: Like the react-native primitive
> * `Select`: not in primitives but good name
> * `Checkbox`:  not in primitives but good name

Works for me.

> * `List`: Can be merged with Select maybe, and add a type="list" prop? I feel they are basically the same element, it's just rendered differently

[They even share the same props](https://github.com/traveltechdeluxe/react-pdf/blob/b6995a3875738165e977c274c76301339e870de2/packages/renderer/index.d.ts#L314-L320) as far as I am aware of. Sounds reasonable. What about `type="combo|list` then as possible `type` values (in sync [with pdfkit](https://github.com/foliojs/pdfkit/blob/master/docs/forms.md#form-annotation-methods))?

natterstefan:
Hey @diegomura, 👋 

Have you seen my reply? I am hopeful that this will be merged asap. :)

natterstefan:
Hi @diegomura, please take a look at my latest changes and edited answer thoughts [here](https://github.com/diegomura/react-pdf/pull/2845#issuecomment-2585908313). //cc @PhilippBloss

chrisdevereux:
would be great to get this reviewed and merged if possible!

natterstefan:
Hey @diegomura. I would be happy if you could give feedback in the next few days. I think some people are waiting for this a well. We also had to use a temporary workaround to be able to use it for our customers already. Please let me know where and how I can help you. Thank you and have a nice day.

jub-nick:
Hey, I would love to see this getting merged. 
@natterstefan Is there any way, that you could expand the scope of this pull request a bit and add the possibility to embed signature fields?
If I understand it correctly pdfkit needs to be extended [here](https://github.com/diegomura/react-pdf/blob/master/packages/pdfkit/src/mixins/acroform.js#L144-L166). According to the [Acrobat Javascript API reference](https://opensource.adobe.com/dc-acrobat-sdk-docs/library/jsapiref/JS_API_AcroJS.html#id619) the field type would be `signature`.
Would love to hear your opinion on this. Thanks in advance!

natterstefan:
Hi @jub-nick,

Thanks for your interest! I appreciate the suggestion to expand the scope of this pull request to include signature fields. It sounds interesting and I am sure it would add even more value to `react-pdf`.

However, at the moment, I suggest adding another PR for your request to get this PR live asap. I am also not 100% sure if I added all supported props of the `signature` field. Having said that here is a draft PR that is ready once this one is merged: https://github.com/traveltechdeluxe/react-pdf/pull/3.

Additionally I added another one for `radiobutton` but that one is not working yet, especially on macOS: https://github.com/traveltechdeluxe/react-pdf/pull/4. I am also more than happy to prepare it further once this PR is merged.

Looking forward to any contributions—thanks again for your input!

packages/primitives/src/index.js
diegomura: `FieldSet` makes sense to me. I'll do it separately so I don't keep blocking this PR merge :)

diegomura (approved):
Thanks for the hard work here! And I apologize again for the delays 😞 It's been very hard to find time lately. I'll take care of the small rename in a separate PR so we can merge this now

If you guys can work on the docs that would be great 🙏🏻

Issue #2825, opened 2024-07-19
MaxListenersExceededWarning: Possible EventEmitter memory leak detected


Torbraw:
**Describe the bug**
When trying to render multiple times the same pdf with renderToBuffer() in a loop in a Nest.js app (with express), I get the following warnings in the console. Note that all the render() methods yield the same error:

```
(node:41468) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 unhandledRejection listeners added to [process]. Use emitter.setMaxListeners() to increase limit
(Use 'node --trace-warnings ...' to show where the warning was created)
(node:41468) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 uncaughtException listeners added to [process]. Use emitter.setMaxListeners() to increase limit
```

If I add `process.on('warning', e => console.warn(e.stack));` to have more infos here's the results:

```
MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 unhandledRejection listeners added to [process]. Use emitter.setMaxListeners() to increase limit
    at genericNodeError (node:internal/errors:984:15)
    at wrappedFn (node:internal/errors:538:14)
    at _addListener (node:events:593:17)
    at process.addListener (node:events:611:10)
    at loadAssembly (C:\Workspace\flexyb-pay\node_modules\yoga-layout\binaries\wasm-async-node.js:11:63)
    at Object.loadYoga (C:\Workspace\flexyb-pay\node_modules\yoga-layout\src\entrypoint\wasm-async-node.ts:25:29)
    at _callee$ (C:\Workspace\flexyb-pay\node_modules\@react-pdf\layout\lib\index.cjs:698:34)
    at tryCatch (C:\Workspace\flexyb-pay\node_modules\@babel\runtime\helpers\regeneratorRuntime.js:45:16)
    at Generator.<anonymous> (C:\Workspace\flexyb-pay\node_modules\@babel\runtime\helpers\regeneratorRuntime.js:133:17)
    at Generator.next (C:\Workspace\flexyb-pay\node_modules\@babel\runtime\helpers\regeneratorRuntime.js:74:21)
    at asyncGeneratorStep (C:\Workspace\flexyb-pay\node_modules\@babel\runtime\helpers\asyncToGenerator.js:3:17)
    at _next (C:\Workspace\flexyb-pay\node_modules\@babel\runtime\helpers\asyncToGenerator.js:17:9)
[...]

PR #2864, opened 2024-09-17
fix: load yoga binaries only once

nnaku:
This PR should fix the #2825 

It's not necessary to load and compile yoga-layout WebAssembly binaries more than once. 
https://github.com/facebook/yoga/issues/1320#issuecomment-1616187392

Current master is a bit broken but this is tested top of the [latest release commit](https://github.com/diegomura/react-pdf/commit/6bf9d990723db757232c42e3a040bbfaa5ceb308)


diegomura (approved):
utACK

Thanks!

PR #2871, opened 2024-09-22
feat!: drop cjs support

diegomura:
For more details look at: https://github.com/diegomura/react-pdf/pull/2711#issuecomment-2366834537


wojtekmaj:
So much cleaner, absolutely nuts!

PR #2877, opened 2024-09-23
chore: bump jay-peg to 1.1.0

Friendseeker:
Thanks for releasing jay-peg 1.1.0!


diegomura:
Thanks for all the fixes there! 
Anyone could probably use the latest changes on jay-peg by just bumping the dep as version in package.json should be compatible. Just mentioning because next release will be 4.0.0 and bumping (depending on your setup) might not be as straighforward as it seems. There we deprecate commonjs from the lib

diegomura (approved):
utACK

Issue #2884, opened 2024-09-25
Infinite loop when splitting pages


cosdon:
**Describe the bug**
Some layouts loop when calling splitPage; all the nodes are transferred to the next page.

**To Reproduce**

Apologies for the layout, it's reverse engineered from a customer's PDF that showed the problem.
This snippet will hang the renderer
```

const style_1 = { padding: "30px 30px 70px 30px" };

const style_2 = { marginBottom: 24 };

const style_3 = {
  fontFamily: "Noto Sans",
  color: "#A1A1A1",
  textTransform: "uppercase",
  fontSize: 9,
  fontWeight: 700,
};

const style_4 = {
  display: "flex",
  flexDirection: "row",
  alignItems: "center",
  flexWrap: "nowrap",
};

const style_5 = {
  fontFamily: "Noto Sans",
  color: "#333333",
  fontSize: 18,
  fontWeight: 600,
};

const style_6 = { flex: 1 };

const style_7 = {
  display: "flex",
  flexDirection: "row",
  paddingTop: 10,
};

const style_8 = {
  display: "flex",
  flexDirection: "row",
  alignItems: "stretch",
};

const style_9 = { marginRight: 8, marginBottom: 3 };

const style_10 = { width: 4, marginRight: 2 };

const style_11 = { backgroundColor: "#F2A24A" };

const style_12 = {
  display: "flex",
  flexDirection: "row",
  alignItems: "center",
};

const style_13 = {};

const style_14 = {
  fontFamily: "Noto Sans",
  color: "#333333",
  fontSize: 9,
  fontWeight: 400,
};

const style_15 = { fontWeight: 700, marginLeft: 6 };

const style_16 = { color: "#5C5C5C", fontWeight: 500 };

const style_17 = { backgroundColor: "#151515" };

const style_18 = {
  fontFamily: "Noto Sans",
  color: "#333333",
  fontSize: 12,
  fontWeight: 400,
};

const style_19 = { fontWeight: 700 };

const style_20 = {
  width: "100%",
  height: 2,
  backgroundColor: "#E8E8E8",
  marginTop: 4,
  marginBottom: 8,
};

const style_21 = { marginBottom: 12 };

const style_22 = { fontSize: 18 };

const style_23 = { fontSize: 18, marginVertical: 18 };

const style_24 = {
  fontFamily: "Noto Sans",
  color: "#333333",
  fontSize: 9,
  fontWeight: 400,
  margin: "4px 0",
};

[...]

PR #2895, opened 2024-10-03
Take padding into consideration when deciding to break on minPresenceAhead

cosdon:
Fixes https://github.com/diegomura/react-pdf/issues/2884

Changes splitPage to pass the box paddingTop through to shouldBreak to be used when deciding if breaking a node will improve its presence. Without this shouldBreak can return true for a node that's already as high as it can go on the page, causing an infinite loop.

### DIFF
--- a/packages/layout/src/node/shouldBreak.js
+++ b/packages/layout/src/node/shouldBreak.js
@@ -26,7 +26,7 @@ const getEndOfPresence = (child, futureElements) => {
   return Math.min(afterMinPresenceAhead, endOfFurthestFutureElement);
 };
 
-const shouldBreak = (child, futureElements, height) => {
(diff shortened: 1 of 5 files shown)


BI-JBrown:
We've been experiencing the issue in question and I can confirm this PR solves the problem. We've added a workaround for now to use margin instead of padding but it's a little hacky. Would love to see this one get merged.

Andrei101Alex:
We've also been experiencing the same issue when it comes to images that push other nodes outside the view in multiple wrapped pages. Accounting for padding in the breaking logic would fix this issue as we use padding to style our wrapped page breaks. This PR should get more attention

georeith:
We're also seeing infinite loops and page hanging. I removed the padding values from our PDF and this fixed them.

PR #2906, opened 2024-10-14
Background image

naresh-tts:


packages/layout/src/image/ninePatch.js
```diff
@@ -0,0 +1,366 @@
+// import { createCanvas, loadImage } from "canvas";
+import { Canvas, loadImage, Image } from "skia-canvas";
+
+class NinePatch {
```
diegomura, line 4: Can you explain what is this?

naresh-tts: @diegomura  I have created background image with ninepatch that automatically create background canvas with shadow for background image we have to just pass image for that and it look lile below.
image
image
image

packages/layout/package.json
```diff
@@ -34,9 +34,11 @@
     "@react-pdf/stylesheet": "^4.2.4",
     "@react-pdf/textkit": "^4.4.1",
     "@react-pdf/types": "^2.4.1",
+    "canvas": "^2.11.2",
     "cross-fetch": "^3.1.5",
     "emoji-regex": "^10.3.0",
     "queue": "^6.0.1",
+    "skia-canvas": "https://github.com/samizdatco/skia-canvas/releases/download/v1.0.1/skia-canvas-v1.0.1-linux-x64-glibc.tar.gz",
```
diegomura, line 41: Why isn't this on npm?

packages/layout/src/image/ninePatch.js
naresh-tts: Yes that is box shadow but we have create image look like box shadow and provide that image as background so that's why it's related to background image.

packages/layout/package.json
naresh-tts: Before it available on npm but latest version is still not available on npm and it's recommendation to use this version from github once it available in npm we will update to get from npm.

packages/layout/src/image/ninePatch.js
diegomura: The fact it's an image it's an implementation detail, but the style is box shadow, not background image. Regardless, I don't think this is the best way of creating box shadows. It should be possible to do it using PDF graphic scripting, not with images

diegomura:
[This](https://github.com/diegomura/react-pdf/pull/2906#discussion_r1911874193) is a concern. It does not seem this is adding background-image support as the title says

PR #2921, opened 2024-10-29
chore: update node versions

farseekers:


diegomura:
Why dropping 21?

farseekers:
It looked like the pattern was to test LTS versions and the latest - 21 didn't fall within that.

Issue #2931, opened 2024-11-13
Tspan inside if Text Element is broken


fabian-ehringer-sii:
**Describe the bug**
When I place multiple `Tspan` elements inside of a `Text` element, the SVG is not rendered correctly.
The `Tspan` elements are rendered on top of each other.

**To Reproduce**
```javascript
const BrokenTspan = () => (
  <Document>
    <Page>
      <Svg>
        <Text x="10" y="30">
          <Tspan>test</Tspan>
          <Tspan style={{ fill: "red" }}>test</Tspan>
          <Tspan style={{ fill: "black" }}>1234</Tspan>
        </Text>
      </Svg>
    </Page>
  </Document>
);

ReactPDF.render(<BrokenTspan />);
```
![image](https://github.com/user-attachments/assets/6943a290-8d53-4428-b5ab-00292c829bf7)

Reproducible example in REPL: https://react-pdf.org/repl?code=3187b0760ce02e004042027101ac0a6600a8400e04330c02f0c00500944407ca405030c00f002223002b80b6e9495df6300157007354bdfbf06019401bb0f11324654003d60ac2008802300064d300279680ccfa162a539f2528a9a0300f458f180b9718bfc31a218036a88400de4130006600967e7e005c309a08a800260600bea9b6f6504e5e6e7c1e9ed604be01c1a191d1719a00467eb8c0c86919da004c26002c3945ee1239aa3cf9928eb2f2834e42a20a4e2cec5c60036400dc3434004aa8f550024c00620074896049a808240c8828e8b9308e944b344000

**Expected behavior**
```xml
<svg xmlns="http://www.w3.org/2000/svg" width="400" height="400">
<text x="10" y="30">
  <tspan>test</tspan>
  <tspan fill="red">test</tspan>
  <tspan fill="black">1234</tspan>
</text>
</svg>
```

![image](https://github.com/user-attachments/assets/ce288cf8-d839-4813-ad22-1f21f2e30cc0)

**Desktop (please complete the following information):**

- OS: MacOS
- Browser: chrome
- React-pdf version: 4.0.0

mantosh-mariner:
Hi, I’m Mantosh , and I’m interested in contributing to this project. I came across issue #1234 and would like to work on it. Could you please assign it to me? 

Thank you for your time and for maintaining this project. Looking forward to your reply!

PR #2961, opened 2024-11-18
fix: calculate x position of tspan elements without explicit x prop

klingebiel-sii:
Fixes https://github.com/diegomura/react-pdf/issues/2931

If the `x` attribute of a `<tspan>` has not been defined explicitly, the current behavior causes the each `<tspan>` to inherit the `x` value of it's parent `<text>` element.

```js
<Svg>
    <Text x="10" y="30">
        <Tspan>One</Tspan>
        <Tspan>Two</Tspan>
        <Tspan>Three</Tspan>
    </Text>
</Svg>
```

**Is**:
image

**Should**:
image

This PR fixes this issue by...
- resetting the inherited `x` value to the original `x` value of the `<tspan>`
- aggregating an offset for each children of a `<text>` node and using it as the child node's `x` value, if necessary

### DIFF
--- a/packages/layout/src/svg/inheritProps.js
+++ b/packages/layout/src/svg/inheritProps.js
@@ -43,6 +43,13 @@ const inheritProps = (node) => {
   const children = node.children.map((child) => {
     const props = Object.assign({}, inheritedProps, child.props || {});
     const newChild = Object.assign({}, child, { props });
+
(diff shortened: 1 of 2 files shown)


fabian-ehringer-sii:
Hi @diegomura, can you please look at this PR. This issue is currently blocking us from using react-pdf. Thanks a lot!! :)

diegomura (approved):

Issue #2335, opened 2023-06-15
Missing dependency on "pako"


texonidas:
**Describe the bug**
While using @react-pdf/renderer in a project with Yarn Plug'n'Play enabled I received several warnings about "pako":
```
The Yarn Plug'n'Play manifest forbids importing "pako" here because it's not listed as a
  dependency of this package:
```

**To Reproduce**
1. Install Yarn 2+ and enable Plug'n'Play in the repo
2. Run `yarn add @react-pdf/renderer`
3. Import @react-pdf/renderer in the root of the project
4. Run the project

**Expected behavior**
Successful project build

**Desktop**
 - OS: Windows
 - @react-pdf/renderer version: 3.1.12

Joshuabaker2:
I just ran into this, I had to add the following to my `.yarnrc.yml` file: 

```
packageExtensions:
  "@react-pdf/pdfkit@*":
    dependencies:
      "pako": "*"
```

zenricatlasopen:
> I just ran into this, I had to add the following to my `.yarnrc.yml` file:
> 
> ```
> packageExtensions:
>   "@react-pdf/pdfkit@*":
>     dependencies:
>       "pako": "*"
> ```

got this error:

"Cannot read properties of undefined (reading 'call')"

Joshuabaker2:
> got this error:
> 
> "Cannot read properties of undefined (reading 'call')" when importing the package

That's because the EventEmitter isn't added in correctly, you have to also add the following to your vite config file via a polyfill: 

```js
const nodePolyfills = require("rollup-plugin-node-polyfills");

module.exports = {
  resolve: {
    alias: {
      events: require.resolve("rollup-plugin-node-polyfills/polyfills/events"),
    },
  },
  plugins: [
    nodePolyfills(),
  ]
};

```

markcnunes:
> I just ran into this, I had to add the following to my `.yarnrc.yml` file:
> 
> ```
> packageExtensions:
>   "@react-pdf/pdfkit@*":
>     dependencies:
>       "pako": "*"
> ```

Thanks, @Joshuabaker2 , for sharing your fix. I had a similar problem with Pnpm.

I’ve been using `@react-pdf/renderer` without any issues, just adding it to package.json worked fine. But after I added `@flatfile/react` to my project, I started getting an error about not finding `pako`:

"Module not found: Package path ./lib/zlib/zstream.js is not exported from package"
https://github.com/diegomura/react-pdf/issues/2795

[...]

PR #3134, opened 2025-03-25
fix(pdfkit): add pako as a direct dependency

ilia-os:
From https://github.com/diegomura/react-pdf/issues/2335 

`yarn modern` doesnt see `pako` as a dependency and throws errors like this:
```
The Yarn Plug'n'Play manifest forbids importing "pako" here because it's not listed as a dependency of this package:
 ```

Adding `pako` as a direct dependency may help.. 
Such package.json modification doesnt lead to yarn.lock change - so there souldnt be any impact

`pako@~1.0.5` is a dependency introduced by `browserify-zlib`
`pako@^0.2.5` is a dependency introduced by `unicode-trie`, which is not a direct dependency and doesnt seem to break yarn

### DIFF
--- a/packages/pdfkit/package.json
+++ b/packages/pdfkit/package.json
@@ -35,7 +35,8 @@
     "fontkit": "^2.0.2",
     "jay-peg": "^1.1.1",
     "linebreak": "^1.1.0",
-    "vite-compatible-readable-stream": "^3.6.1"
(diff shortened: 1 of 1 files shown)


diegomura:
Not sure I get why. pdfkit does not use pako directly for what I can tell

diegomura:
Fork no longer existing in this repo. pdfkit already resolves this

PR #3142, opened 2025-04-05
fix: undefined and null bookmark prop values are rendered instead of being skipped

EvHaus:
[This change](https://github.com/diegomura/react-pdf/commit/481b536f4ad145fb227829399b85a35838a506f8#diff-314d8c0c863d18f6faee1ffac3994f6e232801400e3f13ba9564a8fac61f42f8R34) introduced a regression where a `null` or `undefined` value for the `bookmark` props would be rendered like this

Screenshot 2025-04-04 at 10 34 49 PM

instead of being skipped. This PR fixes that and introduced a test to catch that.

### DIFF
--- a/packages/layout/tests/steps/resolveBookmarks.test.ts
+++ b/packages/layout/tests/steps/resolveBookmarks.test.ts
@@ -18,7 +18,72 @@ describe('layout resolveBookmarks', () => {
 
     const result = resolveBookmarks(root);
 
-    expect(result).toEqual(root);
+    expect(result).toEqual({
+      type: 'DOCUMENT',
+      props: {},
+      children: [
+        {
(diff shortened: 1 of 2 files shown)


packages/layout/tests/steps/resolveBookmarks.test.ts
```diff
@@ -18,7 +18,72 @@ describe('layout resolveBookmarks', () => {
 
     const result = resolveBookmarks(root);
 
-    expect(result).toEqual(root);
+    expect(result).toEqual({
```
EvHaus, line 21: Unrelated to the fix, but updated this test to test against a new variable instance. The previous test would always pass because the original `root` variable would get mutated.

PR #3151, opened 2025-04-19
fix: treat wrap={undefined} like an omitted wrap prop

EvHaus:
## Summary

Similar to #3149 and #3142, this fixes a regression from [the layout TypeScript conversion](https://github.com/diegomura/react-pdf/commit/481b536f4ad145fb227829399b85a35838a506f8#diff-787134450138de88787ea4fed051cb798a2e3a5816955746ab7730617ae6d6e3R11).

After that change, an explicit `wrap={undefined}` was treated as falsy (no wrapping), while an omitted `wrap` prop still defaulted to `true`. Those should behave the same.

### Behaviour

| Usage | Result |
| --- | --- |
| `<View />` | `true` |
| `<View wrap={undefined} />` | `true` |
| `<View wrap={null} />` | `false` |
| `<View wrap={false} />` | `false` |
| `<View wrap={true} />` | `true` |

`null` stays falsy (disables wrapping). Only `undefined` / omitted use the default.

Also adds unit tests for `getWrap`.

## Test plan

- [x] `yarn vitest run packages/layout/tests/node/getWrap.test.ts`
- [ ] Confirm wrapping still works for views with no `wrap` prop
- [ ] Confirm `wrap={false}` still prevents page breaks inside the view

### DIFF
--- a/.changeset/fix-undefined-wrap.md
+++ b/.changeset/fix-undefined-wrap.md
@@ -1,0 +1,5 @@
+---
+'@react-pdf/layout': patch
+---
+
+Treat an explicit `wrap={undefined}` like an omitted `wrap` prop (default `true`). `null` and `false` still disable wrapping.
--- a/packages/layout/src/node/getWrap.ts
+++ b/packages/layout/src/node/getWrap.ts
@@ -8,7 +8,9 @@ const getWrap = (node: SafeNode) => {
 
   if (!node.props) return true;
 
-  return 'wrap' in node.props ? node.props.wrap : true;
+  return 'wrap' in node.props
+    ? node.props.wrap === undefined || node.props.wrap === true
+    : true;
 };
 
 export default getWrap;
--- a/packages/layout/tests/node/getWrap.test.ts
+++ b/packages/layout/tests/node/getWrap.test.ts
@@ -1,0 +1,70 @@
+import { describe, expect, test } from 'vitest';
+
+import getWrap from '../../src/node/getWrap';
+
+describe('node getWrap', () => {
+  test('Should return false by default for non-wrap types', () => {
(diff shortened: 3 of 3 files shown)


diegomura:
Is it correct that `undefined` was already returning true? I'm not sure about null though. That's a falsy value. Makes more sense to me to false for null than true

EvHaus:
> Is it correct that `undefined` was already returning true?

The current behaviour is:

1. `<View />` is treated as `true`
2. `<View wrap={undefined} />` is treated as `false`
3. `<View wrap={null} />` is treated as `false`
4. `<View wrap={false} />` is treated as `false`
5. `<View wrap={true} />` is treated as `true`

I think 1, 4 and 5 make sense. But 2 and 3 are unexpected to me.

> I'm not sure about `null` though. That's a falsy value. Makes more sense to me to false for null than true

I see where you're coming from however given that "no value" (case 1) and `undefined` (case 2) will evaluate to `true` I think having `null` evaluate to `true` is a more developer-friendly/consistent experience.

Let me know if you feel strongly about it and I'll make the PR change accordingly. Thanks @diegomura!

diegomura:
> <View wrap={true} /> is treated as false

I don't think this is right. This case is treated as `true`. I believe that's a typo?

> But 2 and 3 are unexpected to me.

`null` and `undefined` are falsy values. It doesn't make sense to me to be treated as true.

EvHaus:
My bad on the typo. Updated comment and PR description. Kept `null` handling as falsy. However, I would still argue `undefined` should be handled as `true`. Otherwise you end up with:

- `<View />` treated as `true`
- `<View wrap={undefined} />` treated as `false`

Which I think is really unexpected/clumbsy behaviour. If you still disagree, feel free to close the PR.

PR #3183, opened 2025-06-30
fix: resolve images with space in filename

### LINKED ISSUE #3182: Cannot load images with space in filename
Problem is with `getAbsoluteLocalPath` with turns spaces in pathname into `%20`

https://github.com/diegomura/react-pdf/blob/ee5c96b80326ba4441b71be4c7a85ba9f61d4174/packages/image/src/resolve.ts#L34-L55

thgh:
Fixes #3182

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -82,7 +82,7 @@ jobs:
     strategy:
       matrix:
         # We aim to test all maintained LTS versions of Node.js as well as the latest stable version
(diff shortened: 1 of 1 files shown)


diegomura:
Can you please add a test that validates the fix? 🙏🏻

thgh:
I added several tests and changed the fix to something simpler.

Also fixed the github action 😉

PR #3189, opened 2025-07-14
fix(types): allow recursive array of styles

williamsjokvist:
This eases the `style` param typing for all components, allowing [Style[], undefined, Style] etc. to be passed. Which is a common pattern in e.g. React Native. 

Desired behavior: 
```ts
function PaddedView({ style, ...rest }: ViewProps) {
  return (
    <View
      style={[
        {
          padding: 12,
        },
        style
      ]}
      {...rest}
    />
  )
}

```

Before this would give an error, despite the JS working fine. 
```
Type 'Style | Style[] | undefined' is not assignable to type 'Style'.
Type 'undefined' is not assignable to type 'Style'.
```

To circumvent this, I've seen people do these cumbersome runtime checks, which is not necessary with vanilla JS: 
```ts
function Root({ style, ...rest }: ViewProps) {
  return (
    <View
      style={[
        {
          padding: 12,
        },
        ...(Array.isArray(style) ? style : [style ?? {}])
      ]}
      {...rest}
    />
  )
}

```

In my own projects I've added this piece of module augmentation, which is the typing included in this PR:

```ts
declare module '@react-pdf/renderer' {
  interface RecursiveArray<T>
    extends Array<T | ReadonlyArray<T> | RecursiveArray<T>> {}

  type StyleProp<T> = T | RecursiveArray<T | undefined>;

  interface ViewProps {
    style?: StyleProp<Style | Style[]>;
  }
  ...
}
```


tifye (approved):
Looks good to me

diegomura (commented):
I find cumbersome to have `StyleProp<Style | Style[]>` repeated everywhere. Maybe we can just export `StyleProp` to be that instead of using generics?

williamsjokvist:
For sure @diegomura, I think I thought different components had different style interfaces, like `TextStyles`, `ViewStyles` where this generic would make sense, but I see that's not the case. Will fix and update the PR 👍

williamsjokvist:
Hi, I have updated the PR, please have another look @diegomura 

Looks like there's already test coverage for this, it's only unsupported by the typing
https://github.com/diegomura/react-pdf/blob/16cf5bd7006ee1f64d37d3b9ee72e576ee03e343/packages/layout/tests/steps/resolveStyles.test.ts#L56

PR #3190, opened 2025-07-15
Always return both height and width when measuring text

### LINKED ISSUE #2974: RangeError: Invalid array length at Array.push (<anonymous>)
**Describe the bug**
Updating to Version 4.x results in Error:

RangeError: Invalid array length
    at Array.push (<anonymous>)
    at y (yoga-wasm-base64-esm.js:1273:1)
    at 00047d12:0xd3fc
    at 00047d12:0xb7e
    at 00047d12:0x9b62
    at 00047d12:0xa5a5
    at 00047d12:0x5a9c
    at 00047d12:0x769c
    at 00047d12:0xc3cf
    at 00047d12:0x76d7
    

The error occurs when generating a PDF with a very large volume of data (array.length > 5000). Even after taking more than 3 minutes, it eventually causes the error.

We tested using manual pagination, such as chunks, by defining the number of items per page, and it worked in a time 10 times faster. However, this solution is not viable because we cannot fix the number of items per page, as each generated PDF contains fields with variable sizes.

**Expected behavior**
It is expected that the error mentioned above does not occur and that the PDF generation has better performance using its native pagination functions.

- OS: [MacOS, Windows]
- Browser [chrome, safari]
- React-pdf/renderer at anything over 3.4.5

carlobeltrame:
Fixes #2974
Returning an undefined width translated into a NaN width when passing the measurement result back to yoga. This NaN width would then be converted to 0 and a warning as follows would be emitted: Measure function returned an invalid dimension to Yoga: [width=nan, height=21.779297]
The warnings in yoga's javascript bindings are stored into a long array buffer. If a document is very large (> 100 pages) and contains a lot of text nodes, this can produce millions of yoga warnings, inflating this array buffer, until at some point the browser refuses to push another character onto the end of the buffer. This results in the error `RangeError: Invalid array length at Array.push ()`

### DIFF
--- a/.changeset/lemon-dolphins-hunt.md
+++ b/.changeset/lemon-dolphins-hunt.md
@@ -1,0 +1,5 @@
+---
+"@react-pdf/layout": patch
+---
+
+Fix yoga error Invalid array length at Array.push(<anonymous>)
--- a/packages/layout/src/text/measureText.ts
+++ b/packages/layout/src/text/measureText.ts
@@ -26,7 +26,7 @@ const measureText =
     if (widthMode === Yoga.MeasureMode.Exactly) {
       if (!node.lines) node.lines = layoutText(node, width, height, fontStore);
 
-      return { height: linesHeight(node) };
+      return { height: linesHeight(node), width };
     }
 
     if (widthMode === Yoga.MeasureMode.AtMost) {
--- a/packages/layout/tests/text/measureText.test.ts
+++ b/packages/layout/tests/text/measureText.test.ts
@@ -1,0 +1,58 @@
+import { describe, expect, test } from 'vitest';
+
+import * as P from '@react-pdf/primitives';
+import FontStore from '@react-pdf/font';
+import { SafeTextNode } from '../../src';
+import measureText from '../../src/text/measureText';
+
+const TEXT =
+  'Life can be much broader once you discover one simple fact: Everything around you that you call life was made up by people that were no smarter than you';
+
+const fontStore = new FontStore();
+
+const createTextNode = (
+  value: string,
+  style = {},
+  props = {},
+): SafeTextNode => ({
+  style,
+  props,
+  type: P.Text,
+  children: [{ type: P.TextInstance, value }],
+});
+
+describe('measureText', () => {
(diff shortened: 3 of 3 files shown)


sagar-jani:
Thanks for the PR, i can't find this file in `node_modules` - `packages/layout/src/text/measureText.ts`, is it part of `@react-pdf` ?

carlobeltrame:
> Thanks for the PR, i can't find this file in `node_modules` - `packages/layout/src/text/measureText.ts`, is it part of `@react-pdf` ?

It is, otherwise I couldn't have created this PR. Please double check you have installed the package correctly according to the documentation.

Chivas1717 (approved):
great fix! let's merge asap

carlobeltrame:
Thanks everyone for testing the fixes!
@diegomura would you care to review? These problems have reportedly affected a number of users now, and part of the problems were caused by my PR #2400, so I'd like to get this situation sorted out if at all possible.

KaramveerSidhu:
Thanks for the fix @carlobeltrame. I can also confirm that we faced the same problem when generating PDFs greater than 100 pages, and patching the fix, works perfectly. However patching in production is difficult to maintain, and we are waiting for this PR to be merged.

Stith87:
Any plans to get this merged in soon?  This error is causing significant issues for my org and would love to get this fix in. @diegomura

brianshultz:
@diegomura mind taking a look at this one? Have a few customers that are having issues as a result of this bug. Would appreciate a speedy merge!

mohsom:
pls merge

roger-serra-leanon:
Fix worked for me. But I am not sure if related, it's only working if I use a standard page size like 'letter', 'a4', etc. The process just hangs if a custom number is set. Using @react-pdf/renderer": "4.3.0 and react 19.

carlobeltrame:
> Fix worked for me. But I am not sure if related, it's only working if I use a standard page size like 'letter', 'a4', etc. The process just hangs if a custom number is set. Using @react-pdf/renderer": "4.3.0 and react 19.

@roger-serra-leanon interesting. Can you provide a minimal reproduction example, e.g. using https://react-pdf.org/repl ?

PR #3203, opened 2025-08-10
fix: prevent unhandled promise rejections when loading emoji images

100Gyeon:
**Problem:**
When emoji images fail to load from the data source, the rendering process breaks with unhandled promise rejections.

**Solution:**
Added error handling to the `fetchEmojis` function in `@react-pdf/layout`.  
Emoji images that fail to load now gracefully fall back to text rendering, preventing the entire render process from breaking.

**Changes:**
* Added `.catch()` handler to prevent unhandled promise rejections
* Characters without valid emoji images now render as plain text

### DIFF
--- a/packages/layout/src/text/emoji.ts
+++ b/packages/layout/src/text/emoji.ts
@@ -52,10 +52,15 @@ export const fetchEmojis = (string: string, source?: EmojiSource) => {
       emojis[emoji] = { loading: true };
 
       promises.push(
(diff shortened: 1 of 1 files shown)


carlobeltrame:
In case you are using the twemoji images, see also my proposed improvement to the docs on how to set up twemoji integration:
https://github.com/diegomura/react-pdf-site/pull/157

Issue #3223, opened 2025-10-06
Compatibility with React 19 - TypeError: Cannot read properties of null (reading 'props') and Reconciler Incompatibility


ZandercraftGames:
### Description

After upgrading to React v19.2.0, `react-pdf` outputs a TypeError to the console, stating that props on this line is null:
```JS
. . .
  const render = async function (compress) {
    if (compress === void 0) {
      compress = true;
    }
    const props = container.document.props || {}; <--- **ERROR**
    const {
      pdfVersion,
      language,
      pageLayout,
      pageMode,
      title,
      author,
      subject,
      keyboards,
      creator = 'react-pdf',
      producer = 'react-pdf',
      creationDate = new Date(),
      modificationDate
    } = props;
    const ctx = new PDFDocument({
. . .
```

Additionally, a reconciler error occurs seemingly at random. I'm assuming it is happening at unload (see below).

### Steps to reproduce

Set up a React 19 project.
Set up a PDFViewer rendering a document. See error in console.

```TSX
<PDFViewer showToolbar={false} className="h-full w-full">
    {template}
</PDFViewer>
```

### Expected behavior

Document rendered without errors.

### Actual behavior

Document rendered in window properly, but `TypeError: Cannot read properties of null (reading 'props')` error printed to console.

### Additional information

<img width="901" height="569" alt="Image" src="https://github.com/user-attachments/assets/e8b39ddf-41ce-4f67-9eab-4c8a03c9388b" />

```
TypeError: Cannot read properties of null (reading 'props')
    at a (react-pdf.browser.js:140:38)
    at Object.toBlob (react-pdf.browser.js:198:15)
    at react-pdf.browser.js:314:109
    at Het.r.start (index.js:145:17)
    at r.<computed> [as splice] (index.js:59:12)
    at Array.l (react-pdf.browser.js:314:19)
    at t (react-pdf.browser.js:122:62)
    at reconciler-31.js:10:74372
    at Wn (reconciler-31.js:10:74737)
    at hu (reconciler-31.js:10:68185)
    at cs (reconciler-31.js:10:67882)
    at ee (reconciler-31.js:10:8613)
    at U (reconciler-31.js:10:7438)
    at wu.flushSyncWork (reconciler-31.js:10:89114)
    at Object.updateContainer (reconciler-31.js:19:1899)
    at Object.i [as updateContainer] (react-pdf.browser.js:133:14)
    at react-pdf.browser.js:354:25
[...]

PR #3224, opened 2025-10-09
feat: add React 19.2.X support (@react-pdf/reconciler)

ZandercraftGames:
Addresses reconciler incompatibility issue in #3223 

Opening this up to kick things off with some initial fixes to support the reconciler changes in React 19.2.0.

Unfortunately, I'm still getting the `TypeError: Cannot read properties of null (reading 'props')` errors experienced in that issue and some occurrences of [Invalid hook call. Hooks can only be called inside of the body of a function component.](https://react.dev/errors/321), which might still need to be addressed in other components of react-pdf for full 19.2.X compatibility.

The Invalid Hook error seems to be new to React 19.2.0, however, on version 19.1.0, the props error occurs running the latest NPM version of `@react-pdf/renderer`.

Additionally, this makes the version check in the reconciler's `index.ts` more flexible for future releases and revisions to react-reconciler as necessary.


agodard:
Hello, do you know if there is any planned release with this update?

ghigt:
We would also need it, could this be merged?

gajus:
@diegomura Would it be possible to get your attention here?

diegomura (approved):

Issue #3237, opened 2025-10-30
Fix for "Cannot read unitsPerEm"


Pasquale007:
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `@react-pdf/textkit@6.0.0` for the project I'm working on.

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/@react-pdf/textkit/lib/textkit.js b/node_modules/@react-pdf/textkit/lib/textkit.js
index 7ca4f07..4e3f61e 100644
--- a/node_modules/@react-pdf/textkit/lib/textkit.js
+++ b/node_modules/@react-pdf/textkit/lib/textkit.js
@@ -2813,7 +2813,7 @@ const fontSubstitution = () => ({ string, runs }) => {
             // If anything that would impact res has changed, update it
             if (font !== lastFont ||
                 fontSize !== lastFontSize ||
-                font.unitsPerEm !== lastFont.unitsPerEm) {
+                font?.unitsPerEm !== lastFont?.unitsPerEm && lastFont) {
                 if (lastFont) {
                     res.push({
                         start: lastIndex,
@@ -2831,7 +2831,7 @@ const fontSubstitution = () => ({ string, runs }) => {
             index += char.length;
         }
     }
-    if (lastIndex < string.length) {
+    if (lastIndex < string.length && lastFont) {
         const fontSize = getFontSize(last(runs));
         res.push({
             start: lastIndex,
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

chris-farmact:
@Pasquale007 do line breaks (e.g. from text areas or forced in React `{"\n"}`) still work for you?

Pasquale007:
@chris-farmact yes {"\n"} still breaks the line as well as normal text breaks

Issue #3111, opened 2025-03-03
Cannot read properties of undefined (reading 'unitsPerEm')" in @react-pdf/renderer v4


Mario8419:
**Describe the bug**
After upgrading @react-pdf/renderer from v3.4.5 to v4, the following error occurs:

Cannot read properties of undefined (reading 'unitsPerEm') following no pdf rendering

e.g of font register

import font from '../../utils/montserrat-regular.ttf'
import fontBold from '../../utils/montserrat-bold.ttf'

 Font.register({
    family: 'Montserrat',
    fonts: [
      { src: font },
      { src: fontBold, weight: 600 }
    ]
  })

ct-dixit:
the same issue comes up, Is there any solution of this issue?

mAkifArs:
The same issue here.  my font is Inter btw.

diegomura:
Do you get any more extensive error message? Kind of hard to replicate or figure out with so little information

mAkifArs:
It happens on 
`// Setup rendering queue
  useEffect(() => {
    const renderQueue = queue({ autostart: true, concurrency: 1 });

    const queueDocumentRender = () => {
      setState((prev) => ({ ...prev, loading: true }));

      renderQueue.splice(0, renderQueue.length, () =>
        state.error ? Promise.resolve() : pdfInstance.current.toBlob(),
      );
    };

    const onRenderFailed = (error) => {
      console.error(error);
      setState((prev) => ({ ...prev, loading: false, error }));
    };

    const onRenderSuccessful = (blob) => {
      setState({
        blob,
        error: null,
        loading: false,
        url: URL.createObjectURL(blob),
      });
    };

    pdfInstance.current = pdf();
    pdfInstance.current.on('change', queueDocumentRender);
    if (document) {
      pdfInstance.current.updateContainer(document);
    }

    renderQueue.on('error', onRenderFailed);
    renderQueue.on('success', onRenderSuccessful);

    return () => {
      renderQueue.end();
      pdfInstance.current.removeListener('change', queueDocumentRender);
    };
  }, []);`


this part of renderers 
 const onRenderFailed = (error) => {
      console.error(error);
      setState((prev) => ({ ...prev, loading: false, error }));
    };
part of code. 

added the image of console also. 

![Image](https://github.com/user-attachments/assets/08c8b351-ae31-41b0-a9c1-4aad24684f41)

m4xleb:
I have the same error since my last update. 

[...]

PR #3240, opened 2025-11-03
fix(textkit): cannot read units for Em

Pasquale007:
Addressing issue:
https://github.com/diegomura/react-pdf/issues/3237
and:
https://github.com/diegomura/react-pdf/issues/3111

### DIFF
--- a/.changeset/five-rockets-battle.md
+++ b/.changeset/five-rockets-battle.md
@@ -1,0 +1,5 @@
+---
+"@react-pdf/textkit": patch
+---
+
+fix: cannot read units for Em


packages/textkit/src/engines/fontSubstitution/index.ts
```diff
@@ -87,7 +87,7 @@ const fontSubstitution =
       }
     }
 
-    if (lastIndex < string.length) {
+   if (lastIndex < string.length) {
```
diegomura:
```suggestion
    if (lastIndex < string.length) {
```

PR #3266, opened 2025-12-28
chore: remove huge snapshots

diegomura:
Decided to remove these due to the complexity of the snapshots. While the test might be good, it will be impossible to maintain moving forward after any small change.


diegomura:
@carlobeltrame I just re-noticed your comment 

> Especially the textkit layout integration test took a while to write, but should be of great importance to detect unwanted changes in the future.

Sorry. I explained the reason why this change in the PR description but please let me know if you disagree

carlobeltrame:
@diegomura I see. I understand that you don't want a huge snapshot in the library that breaks on every small change. However, users of the library want react-pdf to not change behaviour with every small update. There have been a number of unintentional regressions in the past which I was trying to nail down, and I still think react-pdf urgently needs some kind of integration test.

I chose the snapshot test because it allows me to test the layout part of the pipeline independently of pdfkit. Instead, we could add more end-to-end tests with some document and an expected pdf (effectively also a snapshot), but this would break even more often (every time the layout pipeline is changed PLUS every time pdfkit is changed).

I also chose a snapshot test, because **snapshots are not meant to be edited and reviewed manually line-by-line, but rather just indicate that something has changed**.
That is, I didn't expect anyone to read, understand or manually update the snapshot contents at any point. When something in the code is modified and the snapshot is expected to change, vitest snapshots can be updated automatically at the press of a button (<kbd>u</kbd> when running `yarn run test`) or using a command-line flag (`yarn run test -u` or `yarn run test --update`).
In code review, a changed snapshot should be interpreted as "**something in the layout part of the pipeline has been changed, did you expect that?**" and not as "please view these thousands of changed margin values due to the one-line change". If you want, you can skim over the changes inside the snapshot.

I could have added snapshot tests for narrower scopes, such as only a single engine of the layout pipeline. But some of the regressions I wanted to cover were caused by changes in the order of the engines, and this wouldn't be caught by narrower snapshot tests. Therefore I think the integration test has a value, and the narrower snapshots could be used in addition to it if desired.

My experience with snapshot tests in other projects has been positive, especially because vitest makes it very easy to use them. I can accept if you don't want to use snapshot tests here at all. But I would like to request that some form of integration test is added to catch unintended layout regressions more thoroughly (as in, at all). Otherwise I as a developer using react-pdf just have to write more integration tests or end-to-end tests myself that can catch when an update of react-pdf breaks my app. I'd expect a mature, stable library to never break my application in non-major updates.

PR #3279, opened 2026-01-21
fix(render): border colors handle alpha values correctly

### LINKED ISSUE #1491: RGBA color doesn't work correct with border
**Describe the bug**
When I use border color is `rgba(60, 217, 48, 0.49);`
then it shows like the image below
image

**Expected behavior**
I wish it can show like text

exoego:
Closes #1491

- Use `parseColor()` to extract RGB value and opacity correctly, like `renderBackground` does so                                                                                         
- Apply `strokeOpacity()` for each border side individually
- Add test like `renderBackground.test.ts`

I have verified the change works fine with my work project.

### DIFF
--- a/.changeset/fast-paws-build.md
+++ b/.changeset/fast-paws-build.md
@@ -1,0 +1,5 @@
+---
+"@react-pdf/render": minor
+---
+
+fix border colors with opacity rendered incorrectly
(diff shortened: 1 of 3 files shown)


sorawee:
This works really well! Thank you :)

DrJume:
LGTM?

sorawee:
Hi @diegomura !

I’ve been experiencing the bug that this PR seems to address, and have been using the fork with this PR included as a workaround. From what I can tell, the PR seems to be in a good shape, so I was wondering if it might be ready for review/merge when you have a chance.

Totally understand if you have other priorities though.

Appreciate your time!

PR #3280, opened 2026-01-21
feat(render): support hsla and hwba color models

### LINKED ISSUE #1575: Color opacity not working for hsla
**Describe the bug**
`hsla` is so much easier to work with than the other CSS color types, and while `hsl` does work with `react-pdf`, the opacity does _not_ work, for some reason, even though opacity _does_ work with e.g. `rgba` colors. 😟

**To Reproduce**

1. Go to [snippet at REPL](https://react-pdf.org/repl?code=3187b0760ce02e00408a057025803c450298c0bc300500943807cf805030c00f0022230080b6198531165d400a02180e658212005e19b006f00da00580032c80343001b3c80ba017dda74e54024a37e1838ecad0027801b31e3c49d394003882150938005c300390f00461020960898de0af60e50208e5e8a11a67ee8518cb1e10e3ad6006650a9f13a004e487c0016393071941a1a793010f9c0d80044a5508e101e00f41d8e963cc01800d6485098600074a08c1d7c1d72b21daab28df11ddaa654006a481800ee3516d61276e99c7e7d037cf9200860002600c24120f95edec510bd788a3000acb200a44a3fa03ca6369010c235270b986ee302bdfc8160a84d22718144623e00133011890b44c10a2532b787178d449d1223100a5b1b8fc5a2b224b24321c55789ac1ce20e800a8603c8eb541c540e96d769cea07578023588ae80c662b1d8040037190c800250c1f4a05c1a000c4c6f9162dc30f93c15110a87416156aab210000)
2. See that the red square is not transparent, but should be.

**Expected behavior**

I expect that the opacity for `hsla` works, just like it does if I instead set the `backgroundColor` to for example `#f399` or `rgba(150, 0, 0, 0.4)`. I _especially_ expect it to work since `hsl` _does_ work. At first I thought it was maybe the `hsl` color format that was not supported, but then realized it actually is, which made me even more disappointed that the opacity _doesn't_ work. So close, but yet so far 😢😅

**Screenshots**

❌`hsla`:

✔`rgba`:

exoego:
Closes https://github.com/diegomura/react-pdf/issues/1575

- Bumps color-string so it can parse hsl(a) and hwb(a), and it also supports TypeScript natively
    - Now `^2.1.4` both in render and stylesheet
- Removes outdated own type def for color-string
- Adds hsl-to-rgb and hwb-to-rgb, since color-string does not provides such conversion


diegomura (approved):
utACK

Amazing!

PR #3281, opened 2026-01-21
feat: support external SVG via Image tag

### LINKED ISSUE #1250: Support SVG Base64 Images

One of the ways we have tried to render SVG charts (generated from highcharts or other react charting libraries) in react-pdf is to use a base64 encoded data URI. This is easier than trying to convert an svg element to a react-pdf Svg component, because the browser has built in methods for encoding dataURI's (see code example bellow). 

```
export function encodeSvgString(svg: string) {
  const decoded = unescape(encodeURIComponent(svg));
  const b64String = btoa(decoded);
  const imgSource = `data:image/svg+xml;base64,${b64String}`;
  return imgSource;
}
```

However, the `<Image />` component doesn't support dataURI sources when the type is `svg+xml` instead of jpeg or png. A plain html `` element can render both svg and jpg/png dataURI's, so it would be helpful if the `<Image />` component could support svg dataURI's as well. 

At first I thought it was maybe the length of the base64 string, since charts are complex the string can get pretty long, but I tried with a simplified one and it still doesn't render, and the error I see (with both the large and small dataURIs) is a warning `Not valid image extension` coming from resolveBase64Image because isValidFormat is false. 

I'm not sure how complicated it would be, but if possible it would be helpful to support base64 images with the `data:image/svg+xml;base64` format because it's a simpler way (at least as a user) to incorporate charts from react charting libraries without having to convert an svg element into the `<Svg />` component. 

I have tried converting the svg element into the Svg component ( still working out some bugs #1234 ) I also tried using canvas to convert the svg into a png or jpeg data URI which works but we loose a lot of image quality because it's rasterized instead of a vector image.

**Additional context**

Here is an svg converted into base64 (it's just 2 rectangles from a bar chart) it renders in the browser 
`data:image/svg+xml;base64,PHN2ZyB4bWxuczp4bGluaz0iaHR0cDovL3d3dy53My5vcmcvMTk5OS94bGluayIgdmVyc2lvbj0iMS4xIiBjbGFzcz0iaGlnaGNoYXJ0cy1yb290IiBzdHlsZT0iZm9udC1mYW1pbHk6SGVsdmV0aWNhO2ZvbnQtc2l6ZToxMnB4OyIgeG1sbnM9Imh0dHA6Ly93d3cudzMub3JnLzIwMDAvc3ZnIiB3aWR0aD0iNjAwIiBoZWlnaHQ9IjYwMCIgdmlld0JveD0iMCAwIDYwMCA2MDAiPgogIDxnIGNsYXNzPSJoaWdoY2hhcnRzLXNlcmllcyBoaWdoY2hhcnRzLXNlcmllcy0wIGhpZ2hjaGFydHMtYmFyLXNlcmllcyBoaWdoY2hhcnRzLWNvbG9yLTAiIGRhdGEtei1pbmRleD0iMC4xIiBvcGFjaXR5PSIxIiB0cmFuc2Zvcm09InRyYW5zbGF0ZSg1OTAsNTA5KSByb3RhdGUoOTApIHNjYWxlKC0xLDEpIHNjYWxlKDEgMSkiIGNsaXAtcGF0aD0idXJsKCNoaWdoY2hhcnRzLW8xa2JoMHotMzEzLSkiPgo8cmVjdCB4PSIzNDQuNSIgeT0iMjkxLjUiIHdpZHRoPSI1MCIgaGVpZ2h0PSIyMDQiIGZpbGw9IiNBOURBQ0MiIHN0cm9rZT0iI2ZmZmZmZiIgc3Ryb2tlLXdpZHRoPSIxIiBvcGFjaXR5PSIxIiBjbGFzcz0iaGlnaGNoYXJ0cy1wb2ludCBoaWdoY2hhcnRzLWNvbG9yLTAiPjwvcmVjdD4KPHJlY3QgeD0iMTIwLjUiIHk9IjUxLjUiIHdpZHRoPSI1MCIgaGVpZ2h0PSI0NDQiIGZpbGw9IiNBOURBQ0MiIHN0cm9rZT0iI2ZmZmZmZiIgc3Ryb2tlLXdpZHRoPSIxIiBvcGFjaXR5PSIxIiBjbGFzcz0iaGlnaGNoYXJ0cy1wb2ludCBoaWdoY2hhcnRzLWNvbG9yLTAiPjwvcmVjdD4KPC9nPgo8L3N2Zz4=`

### LINKED ISSUE #1622: Support .svg files in <Image>
Currently react-pdf supports .png, .jpg and base64 images, but .svg format is not.
React-pdf `<Svg>` inline tag allows to describe a svg image, but in use cases, a svg file is an asset on the project (Example: logo.svg) loaded using webpack or other tool

In Image tag, the resolve method for extension should handle .svg extension as other cases. The html `` natively supports, like this `My logo`

- An alternative is to copy the content of all .svg files in assets and transform to react-pdf svg tags or;
- Try to transform .svg image to base64 format (in some scenarios is not possible)

exoego:
Closes #1622
Closes #1250

## Flow

1. Fetch the .svg image (base64 supported too)
2. Parse it as DOM using DOMParser in browser (or own parser in Node.js)
3. Convert the DOM into react-render `<SVG>` tag.

## Other changes

During tests, I found some features are not working properly, so fixed those.
E.g. `rx` in rect, stop in gradient.

## Attributions to SVG files
I've borrowed [a complex SVG ](https://lukaszadam.com/images/free-illustrations/video-call.svg) which is CC0 license (MIT license compatible) from this web page https://lukaszadam.com/illustrations
`<style>` tag in the file are converted to inline style since `<SVG>` seems not to support `<style>` yet. 

Other SVG files are manually crafted.

Screenshot of example :
image


packages/image/package.json
```diff
@@ -24,7 +24,8 @@
   },
   "dependencies": {
     "@react-pdf/png-js": "^3.0.0",
-    "jay-peg": "^1.1.1"
+    "jay-peg": "^1.1.1",
+    "linkedom": "^0.18.12"
```
diegomura: This lib is [quite heavy](https://bundlephobia.com/package/linkedom@0.18.12). I generally try not to add more overhead. Is the rationale here that web (where bundle size is more important) won't need it and it's fine for Node?

Do we actually need a DOMParser polyfill here? Won't an xml to json solution be more lightweight + simplify the logic?

packages/image/src/svg/parse.ts
```diff
@@ -0,0 +1,228 @@
...
+const CrossPlatformDomParser: typeof DOMParser = (() => {
+  if (typeof DOMParser !== 'undefined') {
+    return DOMParser;
+  }
+
+  // In Node.js, use linkedom (only import when needed)
+  // Use BROWSER constant to ensure this code is dead-code eliminated in browser build
+  if (!BROWSER) {
+    // eslint-disable-next-line @typescript-eslint/no-require-imports
+    const linkedom = require('linkedom');
```
diegomura: I think it should be fine to just use import syntax. Rollup should be clever enough to strip it from the browser build.

packages/image/src/types.ts
```diff
@@ -6,7 +6,32 @@ export interface Image {
...
+  width: number;
+  height: number;
+  data: string;
+  format: 'svg';
+  viewBox?: Viewbox;
+  children: SvgNode[];
+  key?: string;
+}
+
+export type AnyImage = Image | SvgImage;
```
diegomura: Lets rename

```suggestion
export type Image = RasterImage | SvgImage;
```

packages/image/src/svg/parse.ts
diegomura, line 1: Nit. Maybe just have all this logic inside the `svg.ts` file and avoid having the `svg` dir. I'd like to keep the `src` file structure flat as it's a simple package. Maybe you can create a `dom-parser.ts` file with the polyfill

packages/image/src/index.ts
```diff
@@ -1,5 +1,12 @@
 import resolveImage from './resolve';
 
-export type { Image, ImageSrc } from './types';
+export type {
+  Image,
+  ImageSrc,
+  SvgImage,
+  AnyImage,
+  SvgNode,
+  Viewbox,
```
diegomura, line 9: Do we need to expose these 2?

packages/layout/src/steps/resolveSvg.ts
```diff
@@ -116,6 +133,23 @@ const parseProps =
...
+  // If both are specified or neither is specified, no change needed
+  if ((rx && ry) || (!rx && !ry)) return node;
+
+  const newProps = Object.assign({}, node.props, {
+    rx: rx ?? ry,
+    ry: ry ?? rx,
+  });
+
+  return Object.assign({}, node, { props: newProps });
+};
```
diegomura, line 153: Nice

packages/layout/src/steps/resolveSvg.ts
```diff
@@ -186,11 +220,55 @@ const parseText =
...
+      gradientTransform: parseGradientTransform,
+      // Stop properties
+      offset: parsePercent,
+      stopColor: transformColor,
+      stopOpacity: parsePercent,
+    } as any,
+    node.props || {},
+  );
+
+  return Object.assign({}, node, { props });
```
diegomura, line 259: I ignore if this is correct 😄

packages/layout/src/text/getAttributedString.ts
```diff
@@ -105,15 +108,15 @@ const getFragments = (
   for (let i = 0; i < instance.children.length; i += 1) {
     const child = instance.children[i];
 
-    if (isImage(child)) {
+    if (isImage(child) && isRasterImage(child)) {
```
diegomura, line 111: Why can't we inline svgs in text? Just not built?

packages/layout/src/steps/resolveSvgImages.ts
```diff
@@ -0,0 +1,63 @@
...
+}
+
+/**
+ * Resolve SVG images step.
+ *
+ * Converts ImageNodes containing SVG data into SvgNodes,
+ * allowing them to be processed by the existing SVG rendering pipeline.
+ */
+function resolveSvgImages(node: SafeNode): SafeNode {
+  if (isImage(node) && isSvgImage(node.image)) {
```
diegomura: NIt. Maybe just have all checks in `isSvgImage`. there's no case where `isSvgImage` should be true but `isImage` false

packages/layout/src/steps/resolveSvgImages.ts
diegomura, line 1: Thinking out loud: I wonder if we shuold not include this logic on `resolveAssets` or `resolveSVG` step... to me it's kind of in the middle. I don't mind having it as a separate step either if you still prefer it. I just don't wanna end having a pipeline of too many steps (in the future)

diegomura (commented):
@exoego amazing work! I appreciate this a lot. 

I left some comments, most nit :) Let me know what you think about them, and feel free to push back on any. I think this is almost ready but I'd like to collect your thoughts on these points first

packages/image/package.json
diegomura: Btw, new `math` module has a parse svg logic, dependency free: https://github.com/diegomura/react-pdf/blob/master/packages/math/src/parseSvg.ts . We can extract it and reuse it maybe?

diegomura: Although that does not have support for style tags and maybe other things a well. Maybe we can reuse your solution here there. I'd like to avoid duplicates

packages/layout/src/text/getAttributedString.ts
exoego: Unfortunately, inline SVG images in `<Text>` are not currently supported.
The `resolveSvgImages` step runs before `resolveTextLayout` (right-to-left composition), converting SVG Image nodes into SvgNodes.
By the time `getAttributedString` processes Text children, SVG images are no longer Image nodes.
Even if caught earlier, PDFKit's inline attachment API requires a raster image buffer (Buffer), while SVG image data is a string of vector drawing instructions.

packages/image/src/types.ts
exoego: Renamed in https://github.com/diegomura/react-pdf/pull/3281/commits/08a850f9744feb961f59d8c77540cff62bc19e8b

packages/image/src/svg/parse.ts
exoego: https://github.com/diegomura/react-pdf/pull/3281/commits/de20194aebd02a3e692fcab98f038325ce7ed7a2

packages/image/package.json
exoego: math's parseSvg seemingly does not support `<!CDATA[` section, comment, single-quote attributes or maybe more.
So, added small parser https://github.com/diegomura/react-pdf/pull/3281/commits/de20194aebd02a3e692fcab98f038325ce7ed7a2

packages/image/src/svg/parse.ts
exoego: This was fixed by replacing linkedom with own parser.

packages/layout/src/steps/resolveSvgImages.ts
exoego: Merged into `resolveSVG` https://github.com/diegomura/react-pdf/pull/3281/commits/02ab3e97ebb5969def979b7a7fe0333fbad5e2cf
The conversion from SVG Image to SvgNode is naturally part of SVG resolution.
It runs right before SVG processing, so keeping them in one step avoids an extra pipeline stage.

`resolveAsset`s is async and focused on fetching images, fonts, etc.
But SVG image conversion is a synchronous tree transformation.
Also, resolveAssets runs twice in the pipeline, so merging into it would run the conversion twice unnecessarily.

diegomura (approved):
utACK 

Thanks!

packages/image/package.json
diegomura: Thanks. I'll extract and reuse that parser for math and other modules that require going from svg to react-pdf primitives

PR #3286, opened 2026-01-23
feat: wrapping text around floating box

### LINKED ISSUE #573: Wrapping text around images.
I would like to use float to wrap text around images. Is this possible with current implementation? I don't want to use `flex-flow: row` to have image and text in one row. because I want to wrap text around an image. The text could be 10 pages long like a lot of articles in news do. 

```js
<View>
  <Image
    src="https://avatars1.githubusercontent.com/u/5691926?s=460&v=4"
    style={{ width: 100, float: 'left' }}
  />
  <Text>
    Lorem ipsum dolor sit amet, consectetur adipiscing elit. Suspendisse
    ut ligula ullamcorper, sodales ante dignissim, commodo lorem. In a
    egestas ex. Sed rhoncus lorem posuere massa rutrum consectetur. In
    at augue vitae risus fermentum volutpat sagittis nec ex.
  </Text>
</View>
```

Using `flex`, but it's not possible to wrap text under the image with it.

exoego:
Closes https://github.com/diegomura/react-pdf/issues/573

View float | Image float | Margin around Float | clear 
---|---|---|---
image | image | image | image

- Implements float: left | right` and `clear: left | right | both | none` by leveraging `excludeRects` in textkit
- We may support [shape-outside](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/shape-outside) to support non-rectangular shape in future, but it requires heavy change in textkit, so only rectangle in this PR.


diegomura:
This is great. I'll try to find some time to review it more carefully

PR #3287, opened 2026-01-24
feat: Svg rotate transformation with explicit origin

### LINKED ISSUE #3244: Svg rotate transformation ignores explicit origin
**Describe the bug**
The `rotate` transformation of SVG elements can accept up to 3 arguments: https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Attribute/transform#rotate

The first one is the angle, while the other two can be used to specify the coordinates of the point from where the element should be rotated.

Currently react-pdf ignores those arguments: https://github.com/diegomura/react-pdf/blob/4aad0f971a484b75c3ff8de6729de15bc6ffe587/packages/render/src/operations/transform.ts#L21

This means that you are unable to rotate elements at a custom point (for example the center of the element), instead it will always rotate around `(0, 0)`.

There was already one bug report about this before, which got closed, because the reproduction was misunderstood: https://github.com/diegomura/react-pdf/issues/2275
In the issue, the blue and red rectangle should **not** align, because the red rectangle should be rotated around `(3, 3)` while it gets rotated around `(0, 0)` like the blue rectangle.

**To Reproduce**
[react-pdf REPL](https://react-pdf.org/repl?code=3187b0760ce02e00408a057025803c450298c0bc300500943807cf805030c00f0022230080b6198531165d400a02180e659a004f003619b006f61622003a000efc3005f769d395006a483007718d3c54a8a231c8026009c7aea460faa8eeb9f3aa019401b9f272f7e54392123e7e7eb6e6500016005c300044008c00acc08cb10034c1212e1118487c11503109c9a919597ecaca99e5301e3aba0042202892b100b489301d30f1000c3dbdb15535ea6a23ea54004a18c050d5232ded89b1f33542d84b2be32e61911b7d5bdbea397905fb83abe5d0962000d61800ea48e11192bdb2bd89c3479cd777e2b1001188878c05ba1c7e30001992044220dae822484c043b6007a7625cfc68b44f450f1030011c103c4b161e4200832290e00c398603c582f4d2bd2c6f8a63339a43288b0bb7260eb589f3b9bb57825514713be4a0fb0976cfef7278bcde1f2f9b2fc30b846c40d67b060e5e31022980c8c17bd96eae7062c896f50e2609374232cc0c9b820c0f4f3060a13c04088992cdb469a6b320fa91600260348d0551b0e71451b38ff2a56758b27b99af850244087d7c72846d069a5500166fa42a0ba88142758c0d8dca0cebc297120468d65adf1fb5b822203f7d31bced77f63d3000331a52730404216043cc04047eefa733fa766806078e678c7343fcc8c77cab1a3d6513e9d3c8553328bc17a1b0ecd816997bf02a0164f30bf7c6fa55ec24ef4f82b1f8ab1e1205ad2c7ad6205c3016d1229ca776dbb3188e20ca83453c3e190f28d0ed0f42c2d0de004318d0ba0186615876008001b86d3214048160430975c0dc130c45ec300c0a0591803259b208604500418804ce0a1310501a090325661a4c0229405cd18301d21f1015056e3e0dd0f40061100441d48a00188a1532ccd53d4451cc730ec3e08a789641282cce0ac9b3ec268a028040461ecc725267394328602b06c5b2449f1c48c050189e220b281e0443c8c0001253046020052584c12c67328000ac1068090284843d3582cb32f2a72a0b02b21946a26d69941280b81a00031590c90f4304b0f02a110541d02c03106a8000)

**Expected behavior**
In this example the red and blue rectangles should **not** align.

This is what the actual svg rendered by a browser looks like:
https://codepen.io/kaesekuchendev/pen/EaKVrbw

**Screenshots**
Current react-pdf:
Image

How the browser renders the same svg:
Image

**Desktop**

- OS: Linux
- Browser Firefox
- React-pdf version 4.2.3

JakabLaszlo1124:
## New functionality

The Svg component's rotate transformation can handle rotation around an explicit origin point. So the following parameterizations are now supported:
- `rotate(angle)` : The original option. It rotates the Svg around the origin (0, 0) point with `angle` degree .
- `rotate(angle, originX, originY)`: The new option.  It rotates the Svg around the (`originX`, `originY`) point with `angle` degree.

## Issues
Closes #3244

### DIFF
--- a/.changeset/slimy-jeans-wait.md
+++ b/.changeset/slimy-jeans-wait.md
@@ -1,0 +1,6 @@
+---
+"@react-pdf/stylesheet": minor
+"@react-pdf/render": minor
+---
+
+feat: Svg rotate transformation with explicit origin


diegomura:
Sorry, I implemented an almost exact solution already 😅

PR #3338, opened 2026-04-05
fix: prevent Thai character text truncation by pre-normalizing ำ (U+0E33)

### LINKED ISSUE #3295: Thai character causes text to be truncated
**Describe the bug**

The Thai character `ำ` causes text truncation.

I presume that there's normalization process somewhere, where a single character `ำ` becomes two characters `ํ` + `า`. However, the string length is still from pre-normalization, causing the truncation.

For a short concrete example, consider `กำก`, which has three characters: `ก` + `ำ` + `ก`. The PDF instead shows `กํา` where the last character `ก` is missing.

Apparently, this is because the normalization causes the text to become 4 characters: `ก` + `ํ` + `า` + `ก`. So if the engine still uses the string length 3, the last `ก` would be omitted.

**Workaround**

In case anyone encounters the same issue, one possible workaround is to create a wrapper of `Text` that replaces `ำ` with `ํ` + `า` explicitly first, so that the normalization doesn't cause the length desync issue.

```
import { Text as RawText } from "@react-pdf/renderer";

interface TextProps {
  content: string;
}

export default function Text({ content }: TextProps) {
  return <RawText>{
    [...content].map((c) => {
      if (c === 'ำ') {
        return "ํา";
      } else {
        return c;
      }
    }).join("")
  }</RawText>
}
```

**To Reproduce**

Consider the following program:

```
import ReactDOM from 'react-dom';
import { Font, PDFViewer } from '@react-pdf/renderer';
import { Document, Page, Text } from '@react-pdf/renderer';

Font.register({
  family: "Sarabun",
  src: "/fonts/Sarabun/Sarabun-Regular.ttf",
});

// Create Document Component
const MyDocument = () => (
  <Document>
    <Page size="A4" style={{ fontFamily: "Sarabun" }}>
      <Text>ก</Text>
      <Text>กำ</Text>
      <Text>กำก</Text>
      <Text>กำกก</Text>
      <Text>กำกกก</Text>
      <Text>กำกกกำ</Text>
      <Text>กำกกกำก</Text>
      <Text>กำกกกำกก</Text>
      <Text>กำกกกำกกก</Text>
    </Page>
  </Document>
);

const App = () => (
  <PDFViewer>
    <MyDocument />
  </PDFViewer>
);

ReactDOM.render(<App />, document.getElementById('root'));
```

where `public/fonts/Sarabun` is the font directory for the Thai font named Sarabun (though any other Thai font will do too). Sarabun can be [downloaded from Google Fonts](https://fonts.google.com/specimen/Sarabun).

**Expected behavior**

It should display:

```
ก
กำ
กำก
กำกก
กำกกก
กำกกกำ
กำกกกำก
กำกกกำกก
กำกกกำกกก
```

Instead, it displays:

```
ก
กํ
กํา
กําก
กํากก
กํากกก
กํากกกํ
กํากกกํา
กํากกกําก
```

**Screenshots**

Image

- OS: MacOS
- Browser Chrome and Firefox
- React-pdf version 2.0.0

cyphercodes:
## Problem

The Thai character ำ (U+0E33) causes text truncation issues as described in #3295. 

When the font engine lays out text containing this character, it decomposes ำ into two separate glyphs: ํ (U+0E4D) + า (U+0E32). This causes a mismatch between:
- String indices (based on the original 3-character string like "กำก")
- Glyph indices (based on the 4 glyphs the font engine creates)

This mismatch causes the text layout engine to calculate incorrect break points, resulting in the last characters being truncated.

## Solution

Pre-normalize Thai text in the transformText function to replace ำ (U+0E33) with its decomposed form ํ + า before font layout. This ensures the string length matches the glyph count, preventing truncation.

## Changes

- Modified packages/layout/src/text/transformText.ts to add normalizeThaiText function
- Added comprehensive test cases in packages/layout/tests/text/transformText.test.ts

## Testing

The fix was verified with test cases covering:
- Thai text with composed characters
- Multiple composed characters in sequence
- Mixed Thai and Latin text
- Various text transformations (uppercase, lowercase, capitalize)

Fixes #3295

### DIFF
--- a/packages/layout/src/text/transformText.ts
+++ b/packages/layout/src/text/transformText.ts
@@ -2,6 +2,25 @@ import { capitalize, upperFirst } from '@react-pdf/fns';
 import { SafeStyle } from '@react-pdf/stylesheet';
 
 /**
+ * Pre-normalize Thai text to handle characters that font engines decompose.
+ * 
+ * Certain Thai characters like ำ (U+0E33) are decomposed by font engines into
+ * multiple glyphs (ํ U+0E4D + า U+0E32), causing a mismatch between string
+ * indices and glyph indices. This leads to text truncation issues.
+ * 
+ * By pre-normalizing these characters, we ensure the string length matches
+ * the glyph count, preventing truncation.
+ * 
+ * @param {string} text
+ * @returns {string} pre-normalized text
+ */
+const normalizeThaiText = (text: string): string => {
+  // Replace Thai character ำ (U+0E33) with its decomposed form ํ + า
+  // This prevents font engine normalization from causing index mismatches
+  return text.replace(/\u0E33/g, '\u0E4D\u0E32');
+};
+
+/**
  * Apply transformation to text string
  *
  * @param {string} text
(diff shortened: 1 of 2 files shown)


sorawee:
As the person who came up with the workaround in #3295 (and thus the fix in this PR), I really think this is just a workaround and not a real solution. We should understand why the desync occurs in the first place and fix that.

You might also want to disclose that your contribution is AI-generated?

diegomura:
Thanks. We have to do this indeed but it's not just thai and it's not just that char. Latest textkit version addresses this so I'll close

Issue #2542, opened 2024-01-22
Switch to Taffy for layout


thecodrr:
Yoga is fine but it only supports Flex layout whereas Taffy has now added support for Flex, Block & Grid layouts with more on their TODO list. I think react-pdf will greatly benefit from the migration.

diegomura:
Interesting. Thanks for posting this. Didn't know about Taffy. I see it's written in rust. Is there a wasm port published somewhere so I can test?

thecodrr:
@nicoburns is working on a WASM build here: https://github.com/DioxusLabs/taffy/pull/394/

diegomura:
Will try to be on the loop and try it out once available

nicoburns:
Currently the packaging/publishing as a module is the blocker for this. Does react-pdf need to run it's layout (so Yoga/Taffy) in both node and browsers? And if so, how are you abstracting away the differences? Rollup?

diegomura:
Thanks for reaching out @nicoburns ! Yes, we run it in both node and browser. Particularly yoga does not need any abstraction to run in both, but several packages of this repo does, and we use rollup for building different versions for each platform

PR #3349, opened 2026-04-07
feat: try taffy

diegomura:


carlobeltrame:
Very exciting! Refs #2542
Is there anything I can do to help out? Could I help by testing or rebasing this?

diegomura:
@carlobeltrame I didn't went too far, but as far as I remember discussing it with claude I came to the conclusion the lib is not yet there. Had a couple of issues to resolve. But I should've written which 😄 At that point I got a bit of cold feet and decided to wait. But I'm open to hear if that's not the case.

I'd be also interested on learning how this would affect bundlesize, specially on web builds. I'd love for react-pdf to be as thin as possible

PR #3367, opened 2026-04-12
perf: Reduce rendering time by ~26% (p95 885→652ms) across the layout and text pipeline

exoego:
I have benchmarked on a 15-page "resume", profiled and optimized a bit.

### DIFF
--- a/packages/layout/src/steps/resolveDimensions.ts
+++ b/packages/layout/src/steps/resolveDimensions.ts
@@ -1,5 +1,6 @@
 import * as P from '@react-pdf/primitives';
-import { isNil, compose } from '@react-pdf/fns';
+import { isNil, matchPercent, compose } from '@react-pdf/fns';
+import * as Yoga from 'yoga-layout/load';
 import FontStore from '@react-pdf/font';
 
 import getMargin from '../node/getMargin';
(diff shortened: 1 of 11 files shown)


packages/layout/src/steps/resolveDimensions.ts
```diff
@@ -65,66 +20,183 @@ import {
...
+  const { style } = node;
+  const height = isPage(node) ? node.box?.height : style?.height;
+
+  // Dimensions
+  applyYogaValue(yogaNode, 'setHeight', height);
+  applyYogaValue(yogaNode, 'setWidth', style.width);
+  applyYogaValue(yogaNode, 'setMinWidth', style.minWidth);
+  applyYogaValue(yogaNode, 'setMaxWidth', style.maxWidth);
+  applyYogaValue(yogaNode, 'setMinHeight', style.minHeight);
+  applyYogaValue(yogaNode, 'setMaxHeight', style.maxHeight);
```
diegomura, line 127: what's the performance difference between this and what we were doing before with each function? This looks to me more like a code design change

packages/textkit/src/engines/bidi/index.ts
```diff
@@ -12,6 +26,18 @@ const bidiEngine = () => {
...
+    if (direction !== 'rtl' && !mayContainRTL(string)) {
+      const runs: Run[] = [
+        {
+          start: 0,
+          end: string.length,
+          attributes: { bidiLevel: 0 },
+        },
+      ];
+      return { string, runs } as AttributedString;
+    }
```
diegomura, line 39: I assume the expensive part is `bidi.getEmbeddingLevels`? Wonder why. I'm not a big fan of these heuristics with early return. I'm not sure this means we can avoid bidi reordering

packages/layout/src/steps/resolvePagination.ts
```diff
@@ -49,6 +49,12 @@ const relayoutPage = compose(
   resolvePageStyles,
 );
 
+// Lightweight relayout for pagination splits. Styles and inheritance are
+// already resolved from the initial layout pass — only yoga dimensions
+// and text layout need recalculation. This is safe because splitNode and
+// splitView preserve fully-resolved styles on all copied nodes.
```
diegomura, line 55: This is not true for dynamic elements

packages/render/src/operations/transform.ts
```diff
@@ -70,7 +70,10 @@ const applyTransformations = (ctx: Context, node: SafeNode) => {
   const { props, style } = node;
   const origin = [node.origin.left, node.origin.top];
   const propsTransform = 'transform' in props ? props.transform : undefined;
-  const operations = style?.transform || propsTransform || [];
+  const styleTransform = Array.isArray(style?.transform)
+    ? style.transform
+    : undefined;
+  const operations = styleTransform || propsTransform || [];
```
diegomura, line 76: ?

packages/textkit/src/layout/generateGlyphs.ts
```diff
@@ -58,14 +64,28 @@ const layoutRun = (string: string) => {
...
-      'ltr',
-    );
+    const fontObj = font[0];
+
+    // Check cache before calling expensive font.layout()
+    let fontCache = layoutCache.get(fontObj);
+    if (!fontCache) {
+      fontCache = new Map();
+      layoutCache.set(fontObj, fontCache);
+    }
```
diegomura, line 74: Why do we need this? Font objects are unique and shared across attributed strings

packages/textkit/src/layout/generateGlyphs.ts
```diff
@@ -58,14 +64,28 @@ const layoutRun = (string: string) => {
...
+      // passing LTR to force fontkit to not reverse the string
+      glyphRun = fontObj.layout(
+        runString,
+        undefined,
+        undefined,
+        undefined,
+        'ltr',
+      );
+      if (fontCache.size >= MAX_CACHE_SIZE) fontCache.clear();
+      fontCache.set(runString, glyphRun);
```
diegomura, line 87: What is this trying to catch? As far as I can tell glyphRun caching will only trigger if we are running layout on the same text string, which not sure how likely it is to justify having this complexity + memory overhead

diegomura (commented):
@exoego I appreciate the effort here but it's a bit too many changes :) Some of which as I commented I don't really see how they have any impact. Would be great to work on performance (I been trying to improve it as well), but bit by bit, not in a big aggregator PR.

The major performance bottleneck is pagination, but the real improvement won't be avoiding 1 or 2 layout steps but at this point I need to fully redesign the algorithm to be O(N). But I been struggling to do so

exoego:
i'll split these

PR #3369, opened 2026-04-12
fix(textkit): fix glyphIndexAt regression

exoego:
<del>Adapt `advanceWidthBetween` to new `glyphIndexAt` semantics from #3358.</del>
Fix `glyphIndexAt` regression from #3358.
`advanceWidthBetween` was missed, causing width=0 for characters at run boundaries when scriptItemizer splits CJK text by script (Han/Hiragana/ Katakana). This broke text wrapping for any multi-script text.

Found this regression while working on #3288


diegomura:
> The refactored glyphIndexAt uses reverse-loop search instead of direct array lookup, requiring callers to use end - 1 (as slice.ts was updated).

Good catch. Wonder if we should fix `glyphIndexAt` requiring to pass end - 1 instead. All methods here don't require that. Was not intentional

packages/textkit/src/run/advanceWidthBetween.ts
```diff
@@ -13,7 +13,7 @@ import { Run } from '../types';
 const advanceWidthBetween = (start: number, end: number, run: Run) => {
   const runStart = run.start || 0;
   const glyphStartIndex = Math.max(0, glyphIndexAt(start - runStart, run));
-  const glyphEndIndex = Math.max(0, glyphIndexAt(end - runStart, run));
+  const glyphEndIndex = Math.max(0, glyphIndexAt(end - runStart - 1, run) + 1);
```
exoego: This is the fix

exoego:
@diegomura Ah, ok. Fixed `glyphIndexAt` instead.

ecrousseau:
I have (I think?) a similar issue - my issue is with rendering Arabic, but the bug seems to also be in `glyphIndexAt` and to be related to the backwards processing. 

I get an error like
```
file:///Users/edward.rousseau/src/clean/node_modules/@react-pdf/textkit/lib/textkit.js:992
            if (addedGlyphs.has(glyph.id))
                                      ^

TypeError: Cannot read properties of undefined (reading 'id')
    at file:///Users/edward.rousseau/src/clean/node_modules/@react-pdf/textkit/lib/textkit.js:992:39
    at Array.map (<anonymous>)
    at reorderLine (file:///Users/edward.rousseau/src/clean/node_modules/@react-pdf/textkit/lib/textkit.js:984:35)
    at Array.map (<anonymous>)
    at reorderParagraph (file:///Users/edward.rousseau/src/clean/node_modules/@react-pdf/textkit/lib/textkit.js:1012:51)
    at Array.map (<anonymous>)
    at Array.<anonymous> (file:///Users/edward.rousseau/src/clean/node_modules/@react-pdf/textkit/lib/textkit.js:1023:39)
    at file:///Users/edward.rousseau/src/clean/node_modules/@react-pdf/textkit/node_modules/@react-pdf/fns/lib/index.js:67:24
    at file:///Users/edward.rousseau/src/clean/node_modules/@react-pdf/textkit/lib/textkit.js:1834:199
    at layoutText (file:///Users/edward.rousseau/src/clean/node_modules/@react-pdf/layout/lib/index.js:1677:19)

Node.js v22.9.0
```

Minimal example to reproduce:
```
import { renderToBuffer } from '@react-pdf/renderer';
import { Document, Page, Text, Font } from '@react-pdf/renderer';
import React from 'react';

Font.register({
  family: 'Noto Sans Arabic',
  src: 'https://cdn.jsdelivr.net/npm/@fontsource/noto-sans-arabic/files/noto-sans-arabic-arabic-400-normal.woff',
});

const doc = React.createElement(
  Document,
  null,
  React.createElement(
    Page,
    null,
    React.createElement(
      Text,
      { style: { fontFamily: 'Noto Sans Arabic', direction: 'rtl' } },
      'نائب وزير'
    )
  )
);

console.log('Rendering...');
renderToBuffer(doc);
```

I'm not sure if the change proposed here will fix my issue. 

For anyone else with the same issue as me: downgrading to `@react-pdf/textkit` 6.1.1 works for now.

PR #3372, opened 2026-04-14
Requested Feature #3341 = breakWhenNeeded

caticHarun:
closes #3341

This feature adds a property "breakWhenNeeded".

If breakWhenNeeded=true and element cannot fit on the page, then it adds a page break before

The main difference between wrap=false and breakWhenNeeded=true is that in wrap={false} the element is limited on one page. Whereas in breakWhenNeeded=true, it only adds a break before and it continues on as it should


diegomura:
I don't think I understand the difference between break and breakWhenNeeded (nor I like the name of the prop too much :))

caticHarun:
break=true always makes a page break

But breakWhenNeeded adds a page break if the element is going to wrap on another page. But if that element is bigger than the whole page (e.g if is is bigger than the A4 Format), then it continues normally (with a page break before)

caticHarun:
Here is the use case I am going for. 

Imagine an invoice style application, that makes pdfs

You have headers, products and lastly total.

If you have a couple of items, then you want the total on the same page.
image


But if you have a bit more items, the total gets wrapped on another page
image


The goal is to add a page break if it wraps, and if it does not, then leave it on the same page

Currently you can do this with wrap=false, but the issue is, when the element is too big to fit on one page, it tries to squeze everything on one paper which can lead to this:

image

And with breakWhenNeeded
image

caticHarun:
Name of the prop is not important, it can always be changed :)

caticHarun:
let me know how to proceed

piotrbaldych-cr:
@diegomura this is such a simple change but such important add-on for multiple use-cases like tables or long lists of complex elements. What can we do to make it merged in?

@caticHarun thank you so much for this!

raffaeldp:
Thank you so much for this PR. I would in fact need exactly this for my app. It would be so great to see it merged :)

Also, if you do not like the name @diegomura, this could maybe become a default behavior of wrap=false. Like if we do not have enough space on the current page, we go to the next one, but if the component is bigger than the whole next page, we break and go to the next one ? 
That could be a solution if there is no usecase of not breaking when component is bigger than one page.

FranckG28:
Would be a very interesting feature !

caticHarun:
I also agree it should be the default behaviour of wrap={false] @diegomura

HugoHartmann067:
This feature would be very useful for generating PDFs with dynamic content, as the current wrap={false} behavior doesn't handle elements larger than a page well. 
Is there any update on getting this merged?

CodeSmith32:
I feel like the mentioned current behavior of `wrap={false}` &ndash; when the contents are longer than a page &ndash; is so bad, it's not even reliable enough to be worth using. I'd almost just rather see this `breakWhenNeeded` feature get absorbed into the `wrap` prop's behavior.

i.e., Instead of creating a new prop, let's just fix the broken `wrap` prop.

I understand this might be considered a 'breaking change', because it will introduce different behavior than before, but IMO, there should *never* be contents squashed on a page because something didn't render correctly. To me, that's a bug. Frustratingly, it happens a lot in this package, and to me, that's because there are a *lot of bugs*.

Props to @caticHarun for fixing this bug with `wrap`!

raffaeldp:
Is this project dead ? @diegomura :'(

PR #3407, opened 2026-04-28
Fix Arabic ligature bidi reordering in textkit

### LINKED ISSUE #3406: Arabic ligatures can be dropped during textkit bidi reordering
## Summary

Arabic ligatures can be dropped or assigned to the wrong visual run during `@react-pdf/textkit` bidi reordering.

This shows up in Arabic PDFs generated through `@react-pdf/renderer` when a shaped Arabic ligature appears more than once in the same line, especially near mixed LTR spans such as company names or percentages.

## Reproduction text

```txt
توجد شراكة استراتيجية مع شركات متعددة. تتكون هيكلية ملكية الشركة من ثلاثة مساهمين، حيث يمتلك المساهم الأول (saudi lend gate) نسبة 50% من الأسهم.
```

The important cases are:

- `شراكة` and `شركات` both use a Cairo Arabic `شر` ligature glyph.
- The same line also contains LTR text, `(saudi lend gate)`, and a percentage, `50%`.

## Actual result

`reorderLine` dedupes ligatures by `glyph.id`. In Cairo, multiple logical `شر` clusters can resolve to the same glyph id, so one logical occurrence is treated as a duplicate and omitted.

Before: Arabic ligature dropped by glyph-id dedupe

## Expected result

Each logical ligature cluster should be emitted once, even if another word uses the same ligature glyph id. Ligature continuation codepoints within the same source glyph should still be deduped.

After: repeated ligatures preserved

## Root cause

In `packages/textkit/src/layout/bidiReordering.ts`, the current implementation:

1. Builds output runs by slicing reordered visual indices with the original logical run boundaries.
2. Dedupes ligatures with `addedGlyphs.has(glyph.id)`.

That makes `glyph.id` too broad as a dedupe key. It removes a second occurrence of the same ligature glyph from a different source cluster. For mixed bidi text, original run boundaries can also no longer describe the correct visual run ownership after reordering.

The safer key is the source run plus the source glyph-array index. That identifies true ligature continuations while allowing the same glyph id to appear in another word.

## Proposed fix

I opened a PR that:

- resolves each visual index back to its owning source run and source glyph index
- emits output runs by contiguous source-run ownership after bidi reordering
- dedupes ligature continuations by source glyph index, not by `glyph.id`
- updates `stringIndices` and `glyphIndices` for the reordered runs
- adds regression tests for repeated ligature glyph ids and Arabic + LTR mixed bidi text

e-Naeim:
Fixes #3406.

## What changed

This updates textkit bidi reordering so reordered runs are built from the source run that owns each visual character index.

Ligature continuations are now deduped by source glyph-array index within the owning run, instead of by `glyph.id`.

## Why

Arabic fonts can reuse the same ligature glyph id for different logical clusters. Cairo, for example, can emit the same `شر` ligature glyph id in separate words such as `شراكة` and `شركات`.

The previous `addedGlyphs.has(glyph.id)` logic treated the second logical cluster as a duplicate and omitted it. With mixed bidi text, slicing reordered visual indices by the original logical run boundaries can also assign glyphs to the wrong visual run.

## Examples

Reproduction text:

```txt
توجد شراكة استراتيجية مع شركات متعددة. تتكون هيكلية ملكية الشركة من ثلاثة مساهمين، حيث يمتلك المساهم الأول (saudi lend gate) نسبة 50% من الأسهم.
```

Before:

Before: Arabic ligature dropped by glyph-id dedupe

After:

After: repeated ligatures preserved

## Validation

- Added regression coverage for repeated ligature glyph ids from different source clusters.
- Added regression coverage for Arabic text mixed with an LTR span.
- Ran a standalone smoke test of the updated `bidiReordering` behavior with both regression inputs.
- Generated before/after PDFs through a downstream `@react-pdf/renderer` stack using Cairo to verify the visible Arabic output.

I could not run the full upstream workspace locally on Windows because the repository contains an existing test fixture filename under `packages/image/tests/assets` that cannot be checked out on NTFS. The PR branch itself is created directly against `master` and only changes `packages/textkit`.

### DIFF
--- a/packages/textkit/src/layout/bidiReordering.ts
+++ b/packages/textkit/src/layout/bidiReordering.ts
@@ -31,12 +31,12 @@ const getReorderedIndices = (string: string, segments) => {
   return indices;
 };
 
-const getItemAtIndex = (runs: Run[], objectName: string, index: number) => {
+const getRunItemAtIndex = (runs: Run[], index: number) => {
   for (let i = 0; i < runs.length; i += 1) {
     const run = runs[i];
-    const updatedIndex = run.stringIndices[index - run.start];
     if (index >= run.start && index < run.end) {
-      return run[objectName][updatedIndex];
+      const glyphIndex = run.stringIndices?.[index - run.start];
+      if (glyphIndex !== undefined) return { run, glyphIndex };
     }
   }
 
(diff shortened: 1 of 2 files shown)


omgovich:
@diegomura Hi! Any chance it could be shipped in the near future?

PR #3409, opened 2026-05-03
refactor(pdfkit): align reference.js with upstream and switch Buffer polyfill to node-stdlib-browser

exoego:
- This PR keeps `src/reference.js` byte-identical with upstream foliojs/pdfkit (no source patches, no shim) as part of #2613  

> [Reference](https://github.com/foliojs/pdfkit/blob/master/lib/reference.js) file differences

The previous [rollup-plugin-polyfill-node](https://www.npmjs.com/package/rollup-plugin-polyfill-node) (unmaintained for ~2.5 years) bundled an outdated inline Buffer impl whose `Buffer.concat` rejected bare `Uint8Array`.
Fontkit v2.x's `subset.encode()` returns a Uint8Array, since they are migrating codebase to Uint8Array, so generating a PDF with embedded fonts in the browser threw "argument must be an Array of Buffers".

So, switched to newer and more popular [node-stdlib-browser](https://www.npmjs.com/package/node-stdlib-browser) + [@rollup/plugin-inject ](https://www.npmjs.com/package/@rollup/plugin-inject).
The newly bundled `buffer@5.7+` accepts `Uint8Array` per current Buffer spec.

The new test, which runs agains the bundle, covers the path.


diegomura:
Managed to push the needed changes to upstream pdfkit. Fork no longer needed

PR #3411, opened 2026-05-07
fix(types): inline Primitive enum values for isolatedModules compatibility + sync enums

benasher44:
## Problem

The `Primitive` enum in `packages/types/primitive.d.ts` initializes its values from `@react-pdf/primitives` runtime imports:

```ts
import * as P from '@react-pdf/primitives';

export enum Primitive {
  G = P.G,
  Svg = P.Svg,
  // ...
}
```

This breaks under TypeScript's `isolatedModules` mode (and `verbatimModuleSyntax`), because enum members cannot be initialized from imported values when each file must be independently transpilable.

Fixed by inlining the string literal values directly + added a test to ensure these stay in sync


benasher44:
Hi @diegomura! I wanted to see if you'd be open to merging this and #3412. We're currently maintaining local patches for these.

PR #3412, opened 2026-05-07
fix(ESM): convert type declarations to ESM exports

benasher44:
## Summary

Replaces the legacy `export = ReactPDF` / `declare namespace ReactPDF` pattern in `index.d.ts` with proper ESM export declarations.

The CJS-style export assignment causes **TS1203** for consumers using `module: NodeNext` or `module: Node16`:

> Export assignment cannot be used when targeting ECMAScript modules.

Since v4 already dropped CJS from the runtime (`feat!: drop cjs support` in #2871), this aligns the type declarations with the actual module format.

## Changes

- Remove `declare class ReactPDF` and `export = ReactPDF`
- Remove `declare namespace ReactPDF` wrapper
- Add `export` to all interfaces and type aliases (previously non-exported namespace members)
- All `export class`, `export const`, and `export function` declarations are unchanged (just un-indented)

### DIFF
--- a/packages/renderer/index.d.ts
+++ b/packages/renderer/index.d.ts
@@ -18,776 +18,851 @@ import {
   HitSlop,
 } from '@react-pdf/types';
 
-declare class ReactPDF {
-  static default: typeof ReactPDF;
+export interface Styles {
+  [key: string]: Style;
+}
+export interface OnRenderProps {
+  blob?: Blob;
 }
 
-export = ReactPDF;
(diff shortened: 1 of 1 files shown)


benasher44:
Recommend viewing the diff with whitespace hidden

MikeAski:
@benasher44 Thank you so much for your contribution

@diegomura Any ETA for a merge & release of this crucial fix?

PR #3421, opened 2026-05-12
feat(textkit): expose line-breaking strategy

wojtekmaj:
This exposes `lineBreakStrategy` on `<Text />` and in textkit layout options so callers can choose between `best-fit` and `knuth-plass` line breaking.

The default remains Knuth–Plass with best-fit fallback. Following the review feedback, this revision removes the automatic long-text threshold, so existing documents keep their current typography and users opt into the performance tradeoff explicitly:

```jsx
<Text lineBreakStrategy="best-fit">...</Text>
```

Why: in a downstream PDF generator, customer-editable body copy can contain very long paragraphs. Profiling showed the hot path was textkit's Knuth–Plass active-node search. For a 50k-character paragraph, the linebreaker built ~20,650 syllables and ~30,431 nodes; node creation took ~6ms, but Knuth–Plass took ~769ms and the heap jumped to ~277MB.

Measured downstream with best-fit selected for long body text:

```text
Knuth–Plass, 200k chars: ~31.3s wall, ~43.5s CPU, ~3.38GB maxRSS
Best-fit, 200k chars:     ~2.3s wall,  ~3.0s CPU,  ~0.5GB maxRSS
```

The generated PDF size was essentially unchanged (~180KB), so the win comes from avoiding intermediate layout work rather than output compression.

The tests use a fixture where Knuth–Plass and best-fit produce different line breaks, verifying both explicit best-fit selection and the unchanged default.

### DIFF
--- a/.changeset/tiny-hats-brake.md
+++ b/.changeset/tiny-hats-brake.md
@@ -1,0 +1,8 @@
+---
+'@react-pdf/layout': minor
+'@react-pdf/renderer': minor
+'@react-pdf/textkit': minor
+'@react-pdf/types': minor
+---
+
+Use faster best-fit line breaking for long non-justified text and expose a lineBreakStrategy option.
--- a/.changeset/tiny-hats-brake.md
+++ b/.changeset/tiny-hats-brake.md
@@ -1,0 +1,8 @@
+---
+'@react-pdf/layout': minor
+'@react-pdf/renderer': minor
+'@react-pdf/textkit': minor
+'@react-pdf/types': minor
+---
+
+Expose a lineBreakStrategy option for choosing faster best-fit line breaking.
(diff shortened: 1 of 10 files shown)


diegomura:
@wojtekmaj doesn't this mean long text line breaking will be of lower quality than short text one? K&P produces much better results than the best fit fallback

wojtekmaj:
Yes, it's an opinionated compromise between quality and performance.

diegomura:
I see. I'm not convinced to merge this as is tbh. The threshold it's not obvious from outside and its value kind of arbitrary. Also for context, react-pdf was sponsored for a long time by an online magazine that focused a lot on editorial accuracy and K&P was one thing they felt very strong about. This will break theirs (and anyone with the same needs) in non obvios ways.

Maybe what we can support is adding the ability to set line-breaking algorithm via props on `<Text />`? It would give users much flexibility too. Some might just want performant solution everywhere

diegomura:
Ideally K&P can me more performant as well. This solution was taken from an old repo in GH, and never look back at it except for some minor fixes. There might be improvements to make

wojtekmaj:
I'll see what I can do. In the meantime, I think it's better to compromise than making the script just... Stop responding.

wojtekmaj:
Rendered obsolete by #3489

PR #3429, opened 2026-05-21
fix(image): replace deprecated url parser

elitan:
## Summary
- replace legacy url.parse usage in @react-pdf/image local path detection
- preserve local relative/absolute path support and reject remote URLs for local reads
- add coverage for file:// image URLs

## Test plan
- yarn workspace @react-pdf/primitives build
- yarn workspace @react-pdf/svg build
- yarn workspace @react-pdf/image typecheck
- yarn vitest packages/image/tests/resolve.test.ts --run

### DIFF
--- a/.changeset/quiet-paths-smile.md
+++ b/.changeset/quiet-paths-smile.md
@@ -1,0 +1,5 @@
+---
+"@react-pdf/image": patch
+---
+
+Replace deprecated url.parse usage in local image path resolution.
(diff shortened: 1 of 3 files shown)


antoinerousseau:
This would also suppress this warning in recent Node.js versions:

```
(node:4) [DEP0169] DeprecationWarning: `url.parse()` behavior is not standardized 
and prone to errors that have security implications.
Use the WHATWG URL API instead. CVEs are not issued for `url.parse()` vulnerabilities.
```

PR #3450, opened 2026-07-07
fix(layout): don't squash a container's sole child at a page break (#3449)

### LINKED ISSUE #3449: Content taller than one page is squashed/overlapped at the bottom instead of flowing onto the next page
# Content overlaps itself at a page break instead of moving to the next page

## Describe the bug

When a block lands near the bottom of a page with *just barely not enough* room to fit, react-pdf force-fits it into the little space left instead of moving it to the next page. Its lines are then drawn on top of each other ("overlapping / scrunched at the page break"), and the rest of the page below is wasted.

It shows up most with two-column rows and `wrap={false}` blocks near a page boundary (headers, footers, itinerary-style entries, table rows).

> The pagination logic lives in **`@react-pdf/layout`**, which `@react-pdf/renderer` depends on. The renderer version can look unrelated (e.g. renderer `4.5.1` → layout `4.6.1`); the bug is in layout **`4.6.1`** and is present on the latest published version.

## How to reproduce

Paste into the [react-pdf REPL](https://react-pdf.org/repl) and look at the **bottom of page 1**. (A plain long `<Text>` is *not* enough to trigger it — it paginates cleanly. The trigger is a `wrap={false}` block that is the **sole child of its container** and lands at the bottom of an already-populated page.)

```jsx
const Block = ({ label, value }) => (
  <View style={{ marginTop: 5 }} wrap={false}>
    <Text style={{ fontSize: 8, fontWeight: 700 }}>{label}</Text>
    <Text style={{ fontSize: 8 }}>{value}</Text>
  </View>
);

const Entry = ({ time, title, a, w, p }) => (
  <View>
    <View style={{ flexDirection: "row", borderBottomWidth: 1, paddingVertical: 8 }} wrap={false}>
      <View style={{ width: 130, marginLeft: 20 }}>
        <Text style={{ fontSize: 11, marginBottom: 5 }}>{time}</Text>
        <Text style={{ fontSize: 7 }}>PENDING</Text>
      </View>
      <View style={{ flex: 1, marginRight: 20 }}>
        <Text style={{ fontSize: 10 }}>{title}</Text>
        <Text style={{ fontSize: 8, marginTop: 5 }}>2 Adults</Text>
        <Block label="Address" value={a} />
        <Block label="Website" value={w} />
        <Block label="Phone" value={p} />
      </View>
    </View>
  </View>
);

const Doc = () => (
  <Document>
    <Page size="A4" style={{ paddingTop: 40, paddingBottom: 90, paddingHorizontal: 40, flexDirection: "column" }}>
      <View fixed>
        <Text style={{ fontSize: 22 }}>Daily Itinerary</Text>
      </View>
      <View style={{ flex: 1 }}>
        {/* spacer: pushes the 2nd entry to the bottom of page 1 */}
        <View style={{ height: 500 }} />
        <View>
          <View minPresenceAhead={100}>
            <View style={{ marginTop: 10, marginBottom: 8 }}>
              <Text style={{ fontSize: 9, fontWeight: 700 }}>TUESDAY, APR 7</Text>
            </View>
            <Entry time="12:25 AM" title="Juvia"
              a="1111 Lincoln Road, Miami Beach, Florida, 33139"
              w="http://www.juviamiami.com/" p="+1 305-763-8272" />
          </View>
          <View>
            <Entry time="01:05 AM" title="Miami Beach Golf Club"
              a="2301 Alton Road, Miami Beach, Florida, 33140"
              w="http://www.miamibeachgolfclub.com/" p="+1 305-532-3350" />
          </View>
        </View>
      </View>
      <View style={{ position: "absolute", bottom: 0, left: 40, right: 40, height: 50 }} fixed>
        <Text style={{ fontSize: 8 }}>EXCLUSIVE RESORTS</Text>
      </View>
    </Page>
  </Document>
);

ReactPDF.render(<Doc />);
```

The overlap only happens when the entry lands with just barely not enough room, and the exact height that does so depends on font metrics (they differ slightly between the browser REPL and a server render). The verified window is about height: 490–510; if 500 renders cleanly for you, try 490 or 510.

Expected behavior

The entry should move to the next page and render normally — never overlap its own lines / the footer.

Screenshots

Before (published 4.6.1) — page 1: the "01:05 AM / Miami Beach Golf Club" entry is squashed onto the bottom of page 1; Address sits on top of the street, Website on the URL, Phone on the number.

Image

After (fix in #3450) — page 1 / page 2: page 1 ends cleanly after "Juvia"; the golf entry moves to page 2 and renders normally.

Image

Root cause

In resolvePagination's splitNodes, when a node is split and all of its children move to the next page (leaving an empty shell), the parent is kept on the current page whenever currentChildren.length === 0. That check is meant to mean "the page is empty," but currentChildren is the local list for the node's own container — it is 0 simply because the node is the first child of its container, not because the page is empty. So a node that is the sole child of its container but lands on an already-populated page is force-fit into the space left, and Yoga compresses its lines on top of each other.

Environment

- @react-pdf/renderer: ^4.5.1 (uses @react-pdf/layout 4.6.1)
- Node.js server-side rendering (also reproduces in the in-browser REPL)

Proposed fix

#3450 — decide based on whether the page is actually empty above the node (threaded through the split recursion), so a node with content above it moves to the next page; it is only kept (and allowed to overflow) when the page is genuinely empty, which also avoids an infinite page loop. Includes regression tests.

billinspiratobell:
Fixes #3449.

## Problem

Content **overlaps itself at a page break** — a block that lands near the bottom of a page is force-fit into the few points of space left, drawing its lines on top of each other, instead of moving to the next page. It shows up with two-column rows / `wrap={false}` blocks (headers, footers, itinerary-style entries) near a page boundary, and is present on the latest published `@react-pdf/layout` (`4.6.1`).

A full step-by-step reproduction (paste-into-REPL) is in [`REPL.md`](./REPL.md).

## Root cause

In `resolvePagination`'s `splitNodes`, when a node is split and **all** of its children move to the next page (leaving an empty shell), the parent is kept on the current page whenever `currentChildren.length === 0`:

```js
if (child.children.length > 0 && currentChild.children.length === 0) {
  if (currentChildren.length === 0) {
    currentChildren.push(child, ...futureFixedNodes); // kept on the current page
  } else {
    // move to the next page
  }
}
```

`currentChildren.length === 0` is meant to mean *"the page is empty, so keep this here (even if it overflows) rather than emit an empty page / loop forever."* But `currentChildren` is the **local** list for the node's own container — it is `0` simply because the node is the **first child of its container**, which is not the same as the page being empty. So a node that is the sole child of its container but lands on an already-populated page is force-fit into the space left, and Yoga compresses its lines on top of each other.

## Fix

Decide based on whether the page is actually **empty above the node**, threaded through the split recursion as `pageEmpty`:

- Page already has content above the node → **move** the whole node to the next page.
- Page is genuinely empty and the node cannot fit → keep it (allow overflow), which preserves termination and avoids an infinite page loop.

The change is contained to `splitNodes` / `split` / `splitView` / `splitChildren`.

## Tests

Added to `packages/layout/tests/steps/resolvePagination.test.ts`:

- **Overlap regression** — asserts zero text-on-text overlap for a `wrap={false}` two-column entry that is the sole child of its container, swept across several page heights. Fails before this change (7/6/5/3 overlapping spans), passes after.
- **Loop guard** — a tall fixed-header case where a node fits the content area but never fits the space left under the header; verifies pagination still terminates (no infinite page loop).

All **416** layout tests pass. Also verified end-to-end by rendering a real document that reproduced the overlap: **8 overlapping spans → 0** after the fix (the entry moves cleanly to the next page).

### DIFF
--- a/.changeset/pagination-page-break-overlap.md
+++ b/.changeset/pagination-page-break-overlap.md
@@ -1,0 +1,5 @@
+---
+'@react-pdf/layout': patch
+---
+
+Fix content overlapping itself at a page break. A node that is the sole child of its container and lands near the bottom of a page (e.g. a `wrap={false}` block, or an element taller than the space left) was force-fit into the remaining space instead of moving to the next page, drawing its lines on top of each other. Pagination now bases that decision on whether the page is actually empty above the node, which also avoids an infinite page loop. Closes #3449.
--- a/REPL.md
+++ b/REPL.md
@@ -1,0 +1,140 @@
+# Reproducing #3449 — content overlapping itself at a page break
+
+This document shows how to reproduce, by hand, the pagination bug fixed in this
+branch: a block that lands near the bottom of a page is **squashed into the few
+points of space left** — its lines drawn on top of each other — instead of
+moving cleanly to the next page.
+
+- **Affected package:** `@react-pdf/layout` (the pagination step). It is what
+  `@react-pdf/renderer` uses to break a document into pages, so the bug shows up
+  through the renderer even though the renderer version number looks unrelated
+  (e.g. renderer `4.5.1` depends on layout `4.6.1`).
+- **Still present on the latest published version** (layout `4.6.1`).
+
+## Quick reproduction (react-pdf REPL)
+
+1. Open <https://react-pdf.org/repl>.
+2. Replace the editor contents with the snippet below.
+3. Look at the **bottom of page 1**.
+
+```jsx
+const Block = ({ label, value }) => (
+  <View style={{ marginTop: 5 }} wrap={false}>
+    <Text style={{ fontSize: 8, fontWeight: 700 }}>{label}</Text>
+    <Text style={{ fontSize: 8 }}>{value}</Text>
+  </View>
+);
+
+const Entry = ({ time, title, a, w, p }) => (
+  <View>
+    <View style={{ flexDirection: "row", borderBottomWidth: 1, paddingVertical: 8 }} wrap={false}>
+      <View style={{ width: 130, marginLeft: 20 }}>
+        <Text style={{ fontSize: 11, marginBottom: 5 }}>{time}</Text>
+        <Text style={{ fontSize: 7 }}>PENDING</Text>
+      </View>
+      <View style={{ flex: 1, marginRight: 20 }}>
+        <Text style={{ fontSize: 10 }}>{title}</Text>
+        <Text style={{ fontSize: 8, marginTop: 5 }}>2 Adults</Text>
+        <Block label="Address" value={a} />
+        <Block label="Website" value={w} />
+        <Block label="Phone" value={p} />
+      </View>
+    </View>
+  </View>
+);
+
+const Doc = () => (
+  <Document>
+    <Page size="A4" style={{ paddingTop: 40, paddingBottom: 90, paddingHorizontal: 40, flexDirection: "column" }}>
+      <View fixed>
+        <Text style={{ fontSize: 22 }}>Daily Itinerary</Text>
+      </View>
+      <View style={{ flex: 1 }}>
+        {/* spacer: pushes the 2nd entry to the bottom of page 1 */}
+        <View style={{ height: 500 }} />
+        <View>
+          <View minPresenceAhead={100}>
+            <View style={{ marginTop: 10, marginBottom: 8 }}>
+              <Text style={{ fontSize: 9, fontWeight: 700 }}>TUESDAY, APR 7</Text>
+            </View>
+            <Entry time="12:25 AM" title="Juvia"
+              a="1111 Lincoln Road, Miami Beach, Florida, 33139"
+              w="http://www.juviamiami.com/" p="+1 305-763-8272" />
+          </View>
+          <View>
+            <Entry time="01:05 AM" title="Miami Beach Golf Club"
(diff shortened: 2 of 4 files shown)


billnbell3:
OK this is ready. See my https://react-pdf.org/repl paste in the code above. It also fixes the recursion. test case is there too.

billnbell3:
CAN WE GET THIS MERGED?

billnbell3 (approved):
ANYONE ELSE look at this?

billnbell3:
please merge this - it is critical fixes.

diegomura:
This should be obsolete after pagination overhaul done in https://github.com/diegomura/react-pdf/pull/3502

billnbell3:
This is still not fixed. See the fix I put in.

the legacy path is still the default and still ships the bug, so the fix isn't obsolete until the cutover major lands.

// layout 5.2.0, lib/index.js
const wantsNextPagination = (root) => root.children.some((page) =>
  page.props?.experimentalPagination || page.props?.layout);
const paginationStep = (root, fontStore) => wantsNextPagination(root)
  ? resolvePagination(root, fontStore)      // new @react-pdf/paginate engine
  : resolvePagination$1(root, fontStore);   // legacy — still has your bug

PR #3457, opened 2026-07-22
fix(layout): strip default-ignorable characters instead of rendering fallback glyphs

hendrikmolder:
Fixes rendering of Unicode Default_Ignorable_Code_Points (bidi controls, zero-width chars, BOM, variation selectors, …) which currently show up as visible fallback `.notdef` glyphs in generated PDFs.

## Problem

Characters like `U+202A` (LRE), `U+202C` (PDF), `U+200F` (RLM), zero-width spaces/joiners, and the BOM are default-ignorable — browsers render nothing for them. In react-pdf they appear as garbage, typically a stray glyph overprinted on the first letter of the following text (plus a swallowed adjacent space).

This is common with text (e.g. names) for RTL locales, which often arrive wrapped in bidi embedding controls. Example: `"\u202aExample Name\u202c\u200f"` renders with a stray glyph stacked on the first letter.

## Root cause

`packages/layout/src/text/ignoreChars.ts` only stripped a 3-entry list (`U+2028`, `U+2029`, `U+2060`). All other default-ignorables reached font substitution, missed the font cmap, and were drawn as fallback glyphs. fontkit already handles these correctly via `isDefaultIgnorable` / `hideDefaultIgnorables`, but that path is not reached because react-pdf performs its own per-codepoint font substitution first.

## Changes

- Strip characters the fragment font cannot render when they are Unicode Default_Ignorable_Code_Points, mirroring fontkit's `isDefaultIgnorable` ranges.
- Preserve existing `U+2028` / `U+2029` separator stripping (kept via an explicit extra set) to avoid a regression.
- Characters the font *does* provide a glyph for are left untouched.
- Adds unit tests in `packages/layout/tests/text/ignoreChars.test.ts` and a changeset.

## Testing

`yarn vitest run packages/layout` — all 417 layout tests pass (incl. 6 new). Lint + typecheck clean.

### DIFF
--- a/.changeset/strip-default-ignorable-chars.md
+++ b/.changeset/strip-default-ignorable-chars.md
@@ -1,0 +1,5 @@
+---
+"@react-pdf/layout": patch
+---
+
+fix(layout): strip default-ignorable characters (bidi controls, zero-width chars, BOM) so they are not rendered as fallback `.notdef` glyphs
--- a/packages/layout/src/text/ignoreChars.ts
+++ b/packages/layout/src/text/ignoreChars.ts
@@ -1,27 +1,107 @@
 import { Font, Fragment } from '@react-pdf/textkit';
 
-const IGNORABLE_CODEPOINTS = [
-  8232, // LINE_SEPARATOR
-  8233, // PARAGRAPH_SEPARATOR
-  8288, // WORD_JOINER
-];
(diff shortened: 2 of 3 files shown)


diegomura:
Is it intended for this to be in draft?

PR #3459, opened 2026-07-24
Fix bidi reordering for indices absorbed into glyph clusters

loaialaddien:
## Summary

Prevent bidi reordering from dereferencing a missing glyph mapping when complex-script shaping absorbs a source position into an existing glyph cluster.

Cases covered:

- `abc र्क ابج def` — regression case.
- `abc र्क def` — complex-script control.
- `abc ابج def` — RTL control.
- `abc किंतु ابج def` — additional mixed-script control.

Without the fix, the regression case throws at `glyph.id`.


diegomura:
Can you check if this is fixed on upstream changes?

PR #3463, opened 2026-07-31
Fonts: allow registration of fonts from `Buffer` and `blob:` URLs

stephenjudkins:
This allows us to save round-tripping base64 de/encoding of binary font data.

### DIFF
--- a/packages/font/src/types.ts
+++ b/packages/font/src/types.ts
@@ -40,8 +40,12 @@ export type FontSourceOptions = {
   postscriptName?: string;
 } & RemoteOptions;
 
+export type BufferFontSource = {
+  buffer: Buffer;
+};
+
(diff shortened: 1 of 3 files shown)


packages/font/src/types.ts
```diff
@@ -40,8 +40,12 @@ export type FontSourceOptions = {
   postscriptName?: string;
 } & RemoteOptions;
 
+export type BufferFontSource = {
+  buffer: Buffer;
+};
```
diegomura, line 45: Can we use Uint8Array instead of Buffer? Buffer is a subclass of Uint8Array anyways, but the former exist in node and web while the latter is Node only

stephenjudkins: ✅ done

PR #3484, opened 2026-08-12
fix(layout): don't leave a wrapping parent at the end of a page with only its fixed child

zewish:
## Summary

Found this while exporting a print menu, where a heading was alone at the bottom of one page and appeared again above its products on the next one. We use a patch-package patch until this is (hopefully 🤞) merged and released.

## Minimal reproduction

The page is 300pt tall and the first block takes 255pt of it. That leaves 45pt, which is
enough for the 20pt header but not for the 40pt row after it.

```jsx
<Document>
  <Page size={{ width: 200, height: 300 }}>
    <View style={{ height: 255 }} />
    <View>
      <View fixed style={{ height: 20 }} />
      <View wrap={false} style={{ height: 40 }} />
      <View wrap={false} style={{ height: 40 }} />
      <View wrap={false} style={{ height: 40 }} />
    </View>
  </Page>
</Document>
```

**Old behavior:**
- Page 1 ends with the header and nothing under it
- Page 2 starts with the same header, followed by all three rows

**New behavior:**
- Page 1 ends after the first block, and the header and its rows stay together on page 2

### DIFF
--- a/.changeset/ripe-aliens-melt.md
+++ b/.changeset/ripe-aliens-melt.md
@@ -1,0 +1,5 @@
+---
+'@react-pdf/layout': patch
+---
+
+Don't leave a wrapping parent at the end of a page with only its fixed child
--- a/packages/layout/src/steps/resolvePagination.ts
+++ b/packages/layout/src/steps/resolvePagination.ts
@@ -114,7 +114,10 @@ const splitNodes = (height: number, contentArea: number, nodes: SafeNode[]) => {
       const [currentChild, nextChild] = split(child, height, contentArea);
 
       // All children are moved to the next page, it doesn't make sense to show the parent on the current page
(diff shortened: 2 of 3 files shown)


zewish:
Hey @diegomura, I would love if you're able to give me some feedback if this PR is useful to the project or not.
Feel free to close it down if it doesn't make sense.

diegomura:
> Page 1 ends after the first block, and the header and its rows stay together on page 2

Hey @zewish . Not sure how this is the expected result if the fixed view of height 20 fits in the 1st page

zewish:
Thanks for the feedback. We have since switched to a different PDF generation approach which doesn’t use react-pdf at all, so we no longer have need for this patch anyways.

PR #3489, opened 2026-08-13
perf(textkit): improve Knuth-Plass line breaking performance

wojtekmaj:
This improves textkit's Knuth-Plass performance for long paragraphs without changing line-breaking behavior.

The Knuth-Plass paper treats line numbers as equivalent once the line widths become identical. The current implementation keeps separating them, so newly inserted candidates are reconsidered at the same breakpoint and the active-node search grows approximately quadratically on the measured corpus.

Changes:

- group equivalent line-number candidates after the last distinct available width
- keep distinct candidates while variable line widths still apply
- add regression fixtures for alignments, variable widths, hyphenation, and best-fit fallback

Measured on an Apple arm64 Mac with Node 24.17.0, using isolated processes, two warmups, and seven measured runs:

```text
500 words:  5.43ms ->  1.54ms wall, 114MB -> 111MB maxRSS
1k words:  15.62ms ->  4.17ms wall, 123MB -> 115MB maxRSS
2k words:  50.72ms ->  4.44ms wall, 222MB -> 118MB maxRSS
4k words: 183.62ms -> 12.76ms wall, 334MB -> 124MB maxRSS
8k words: 749.56ms -> 19.23ms wall, 693MB -> 141MB maxRSS
```

At 8,000 words, candidate evaluations fell from ~36.6M to ~852k and active-node creation fell from ~1.49M to ~34.8k. Candidate evaluations changed from approximately quadratic growth (empirical exponent 2.01) to approximately linear growth (1.04) across the measured matrix.

Breakpoint hashes match at every benchmark size. A separate 192-case baseline/optimized matrix also matches across constant and variable widths, tolerances, hyphenation settings and penalties, and failure cases.

Validation:

- `yarn workspace @react-pdf/textkit test --run`: 794 tests passed
- `yarn workspace @react-pdf/textkit typecheck`: passed
- `yarn workspace @react-pdf/textkit build`: passed
- changed-file ESLint and Prettier: passed
- full suite: 2,276 tests passed; four unrelated pdfkit font tests could not resolve their fixture because the checkout path contains spaces
- root typecheck still has existing `@react-pdf/render` bookmark and SVG type errors; textkit typecheck passes

### DIFF
--- a/packages/textkit/tests/engines/linebreaker.test.ts
+++ b/packages/textkit/tests/engines/linebreaker.test.ts
@@ -10,6 +10,36 @@ const width = 50;
 describe('linebreaker', () => {
   const linebreaker = linebreakerFactory({});
 
+  const createAttributedString = (
+    align: 'left' | 'right' | 'center' | 'justify',
+    syllables: string[],
+  ) => {
+    const string = syllables.join('');
+    const indices = Array.from({ length: string.length }, (_, index) => index);
+
+    return {
+      string,
+      syllables,
+      runs: [
+        {
+          start: 0,
+          end: string.length,
+          attributes: { align, font: [font] },
+          stringIndices: indices,
+          glyphIndices: indices,
+          positions: Array.from(string, (character) => ({
+            xAdvance: character === ' ' ? 3 : 6,
+            yAdvance: 0,
+            xOffset: 0,
+            yOffset: 0,
+            advanceWidth: character === ' ' ? 3 : 6,
+          })),
+          glyphs: [],
+        },
+      ],
+    };
+  };
+
   test('should break lines and adds hyphens only where indicated', () => {
     const attributedString = {
       string: 'Potentieel broeikasgasemissierapport',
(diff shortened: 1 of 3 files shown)


packages/textkit/tests/engines/linebreaker.test.ts
```diff
@@ -296,6 +326,86 @@ describe('linebreaker', () => {
...
+        'gamma delta ',
+        'epsilon zeta ',
+        'eta theta ',
+        'iota kappa ',
+        'lambda mu',
+      ]);
+    },
+  );
+
+  test('should preserve breaks with variable line widths', () => {
```
diegomura, line 349: This test and the ones below don't use `align` variable. Arent we executing the same tests 4 times?

wojtekmaj: Yep, you're right. This fixture doesn't create penalty nodes, so align has no effect here. Collapsed it to one constant-width case in 5961a4a6.

diegomura:
Thanks @wojtekmaj !

PR #3534, opened 2026-08-26
fix(mermaid): render inside Web Workers and honor the theme prop

diegomura:
Two fixes to `@react-pdf/mermaid`, split out of the docs-site branch.

### Web Worker rendering

elkjs (the layout engine behind mermaid) decides what it is from globals: `self` defined and no `document` means "I am the elk worker script", so it hijacks `self.onmessage` and exports no in-process worker. Inside a Web Worker that describes the *host*, not elk, so the first diagram threw and took the host's message handling with it.

`inWorkerScope` stubs a `document` around the render call to send elk down its library branch, and defines an own writable `self` so the restore beautiful-mermaid does on the way out doesn't hit a worker's getter-only `self`. Both are torn down in a `finally`.

### `theme` prop

Named themes are palettes in beautiful-mermaid, not a render option, so passing `theme` did nothing. `themeColors` expands the name into colors before rendering, and those colors now also feed `preprocessSvg` and the fallback text color.

### Tests

`tests/workerScope.test.ts` simulates a worker global (no `document`, getter-only `self`) and asserts a diagram renders and the scope is restored. `tests/render.test.ts` covers theme expansion.

### DIFF
--- a/.changeset/khaki-hoops-repeat.md
+++ b/.changeset/khaki-hoops-repeat.md
@@ -1,0 +1,7 @@
+---
+'@react-pdf/mermaid': patch
+---
+
+Fix rendering inside a Web Worker. elkjs read `self` without `document` as "I am the elk worker script", hijacked the host's `onmessage` and exported no layout engine, so the first diagram threw.
+
+Fix `theme` prop being ignored. Named themes are palettes in beautiful-mermaid, not a render option, so they are now expanded into colors before rendering.
--- a/.changeset/khaki-hoops-repeat.md
+++ b/.changeset/khaki-hoops-repeat.md
@@ -2,6 +2,6 @@
 '@react-pdf/mermaid': patch
 ---
 
-Fix rendering inside a Web Worker. elkjs read `self` without `document` as "I am the elk worker script", hijacked the host's `onmessage` and exported no layout engine, so the first diagram threw.
+Fix diagrams failing to render inside a Web Worker.
 
-Fix `theme` prop being ignored. Named themes are palettes in beautiful-mermaid, not a render option, so they are now expanded into colors before rendering.
+Fix the `theme` prop being ignored.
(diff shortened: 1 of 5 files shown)


packages/mermaid/src/render.ts
```diff
@@ -9,12 +9,50 @@ export interface MermaidRenderOptions {
...
+
+/**
+ * The layout engine behind mermaid, elkjs, reads its own globals to decide what
+ * it is: `self` defined and no `document` means "I am the elk worker script", so
+ * it hijacks `self.onmessage` and exports no in-process worker. Inside a Web
+ * Worker that describes the host, not elk, and rendering throws — taking the
+ * host's own message handling with it. A stub `document` sends elk down its
+ * library branch, and an own writable `self` absorbs the restore beautiful-
+ * mermaid does on the way out (a worker's `self` is getter-only).
+ */
```
diegomura, line 25: simplify this. use simpler language

.changeset/khaki-hoops-repeat.md
```diff
@@ -0,0 +1,7 @@
+---
+'@react-pdf/mermaid': patch
+---
+
+Fix rendering inside a Web Worker. elkjs read `self` without `document` as "I am the elk worker script", hijacked the host's `onmessage` and exported no layout engine, so the first diagram threw.
+
+Fix `theme` prop being ignored. Named themes are palettes in beautiful-mermaid, not a render option, so they are now expanded into colors before rendering.
```
diegomura: simplify this as well

packages/mermaid/src/render.ts
diegomura: Simplified in 70b67cac.

.changeset/khaki-hoops-repeat.md
diegomura: Simplified in 70b67cac.

Issue #3532, opened 2026-08-26
@react-pdf/font started to depend on pdfkit@0.20.1 and triggers a TypeError: Invalid URL


Lucas-C:
**Describe the bug**
Upgrading `@react-pdf/font` from `4.0.9` to `4.1.0` or `4.1.1` caused a `TypeError: Invalid URL`

**To Reproduce**
It is difficult for me to share a reproducible code snippet / example.

In our project, `@react-pdf/font` is a dependency of `@react-pdf/renderer@4.6.0(react@19.2.8)` and `@react-pdf/types@2.13.1`.
We use TypeScript, pnpm, and our code is transpiled using ESBuild.
When invoking Node on the transpiled JS bundle, we get this stracktrace:
```
TypeError: Invalid URL
    at new URL (node:internal/url:840:25)
    at Object.<anonymous> (/app/node_modules/.pnpm/pdfkit@0.20.1/node_modules/pdfkit/lib/mixins/pdfa.js:3:33)
    at Module._compile (node:internal/modules/cjs/loader:1871:14)
    at Object..js (node:internal/modules/cjs/loader:2002:10)
    at Module.load (node:internal/modules/cjs/loader:1594:32)
    at Module._load (node:internal/modules/cjs/loader:1396:12)
    at wrapModuleLoad (node:internal/modules/cjs/loader:255:19)
    at Module.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:154:5)
    at node:internal/main/run_main_module:33:47
```

It seems that `@react-pdf/font@4.1.0` started to depend on `pdfkit@0.20.1`, and `TypeError: Invalid URL` is raised from there:
https://github.com/foliojs/pdfkit/blob/v0.20.1/lib/mixins/pdfa.js#L3
```javascript
export const ICC_PROFILE_PATH = new URL(
  './data/sRGB_IEC61966_2_1.icc',
  import.meta.url,
).href;
```

**Expected behavior**
No exception raised

**Desktop (please complete the following information):**

- OS: Linux
- React-pdf version:  `@react-pdf/renderer@4.6.0(react@19.2.8)`

alainrk:
We hit the same error after upgrading `@react-pdf/renderer` from 4.3.2 to 4.5.1. 
The failure is still there with the current `pdfkit` master (0.20.2), so newer renderer versions are affected too.

Here is a minimal reproduction and some details that may help.

## Reproduction

[...]

PR #3550, opened 2026-09-01
fix: bump pdfkit to 0.20.2 for tracer-safe standard fonts

mrowe:
## What

Bumps the exact `pdfkit` pin in `@react-pdf/font` and `@react-pdf/renderer` from `0.20.1` to `0.20.2`, with a patch changeset for both packages. The lockfile regeneration also let yarn re-normalize one previously merged babel entry.

## Why

pdfkit 0.20.0/0.20.1 map `#standard-fonts/*` to `.cjs` under the `require` condition but `.mjs` under `default`. Node resolves the `.cjs` files at runtime, while file tracers and bundlers — e.g. `@vercel/nft`, which builds Next.js `output: 'standalone'` deployments — resolve the `.mjs` twins. The traced output therefore omits the modules the Node build actually lazy-loads, the build succeeds, and the first server-side render throws:

```
Error: Cannot find module '/app/node_modules/pdfkit/js/standard-fonts/Helvetica.cjs'
```

Helvetica is the default font, so any document without registered fonts hits this immediately. We hit this in production after upgrading `@react-pdf/renderer` 4.5.1 → 4.9.0 (the 4.8.0 fork→upstream pdfkit swap): guest receipt PDFs 500'd in the deployed container while local dev and the build itself were green.

pdfkit 0.20.2 fixes this upstream (foliojs/pdfkit#1782) by resolving the subpath to one file under every condition, so tracers pack what runtime loads. `0.20.1 → 0.20.2` is patch-only with no API change.

Verified in our Next.js standalone deployment: with 0.20.2 (via npm `overrides`) the traced output contains all 14 `standard-fonts/*.cjs` files and server-side rendering works; with 0.20.1 it contains none of them.

Note this is unrelated to #3532 (`TypeError: Invalid URL` from `import.meta.url` under ESBuild-to-CJS transpilation) — different mechanism, not fixed by this bump.

https://claude.ai/code/session_013Z4s94hfd1MWA6TnLmRXEb

### DIFF
--- a/.changeset/bump-pdfkit-standard-fonts.md
+++ b/.changeset/bump-pdfkit-standard-fonts.md
@@ -1,0 +1,6 @@
+---
+'@react-pdf/font': patch
+'@react-pdf/renderer': patch
+---
+
+Bump pdfkit to 0.20.2 so file tracers and bundlers pick up the CommonJS standard-font modules the Node build actually loads, instead of failing at first render with `Cannot find module .../standard-fonts/Helvetica.cjs` in traced deployments (e.g. Next.js standalone output).
--- a/packages/font/package.json
+++ b/packages/font/package.json
@@ -26,7 +26,7 @@
     "@react-pdf/types": "^2.14.0",
     "fontkit": "^2.0.2",
     "is-url": "^1.2.4",
-    "pdfkit": "0.20.1"
+    "pdfkit": "0.20.2"
   },
   "files": [
     "lib"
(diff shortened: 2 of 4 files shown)


VladSez:
@mrowe why this was closed?

ptts:
Same question here. This looks like a clean patch bump, and it fixes a real runtime break, so closing it without a note is confusing. Was there a regression it caused, or is the pdfkit bump being handled somewhere else? master still pins pdfkit@0.20.1.

VladSez:
cc @diegomura

mrowe:
Sorry, just had to clear some things up internally. Should be good to go now.

Issue #1252, opened 2021-04-23
not working on webp image


quanvu1204:
**Describe the bug**

**To Reproduce**
Steps to reproduce the behavior including code snippet (if applies):

_You can make use of [react-pdf REPL](https://react-pdf/repl) to share the snippet_

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**
 - OS: [e.g. MacOS, Windows]
 - React-pdf version [e.g. v1.1.0]

lucassshanks:
bumping this 🙏

BStephenBB:
I'd be up for working on this if someone can point me in the right direction and the maintainers of the project are okay with webp support being added.

It looks like a webp equivalent of `image/src/jpeg.js` and `image/src/png.js` would need to be created and then integrated into `resolve.js`. It seems the first step for doing that would be to find a jb lib with the desired webp related functionality? @diegomura thoughts?

diegomura:
@BStephenBB I would love webp support of course!

The approach you mentioned is what I would've suggested. When finding (or building) a webp js lib we should be very careful about how much code it introduce to the lib

babyccino:
@diegomura have you or anyone else put any work into this? I wouldn't mind having a crack at it if there isn't already someone on the task

diegomura:
I have not. Sorry

imyasha:
@babyccino Are you working on this?

babyccino:
> @babyccino Are you working on this?

No I've never had enough free time to have a go at it. Would be greatly appreciated if it was done though 👍

imyasha:
@babyccino @diegomura 
I hit a PR for this issue please review  this one when you have a min.
https://github.com/diegomura/react-pdf/pull/2463

fasiha:
Per https://github.com/foliojs/pdfkit/issues/1497 Pdfkit can't support webp because PDFs themselves don't support webp. You'll need to convert to a supported image format first 😕

agusdwimilniadi:
use this for convertion https://images.weserv.nl/

DTDucas:
I've opened #3557 for this.

[...]

PR #3557, opened 2026-09-05
feat(image): support WebP images

DTDucas:
PDF has no WebP. An image XObject can carry JPEG (DCTDecode) or flate-compressed raw pixels (FlateDecode, which is what a PNG decodes into), and there is no filter for a VP8/VP8L bitstream. WebP is therefore decoded and re-encoded to PNG while the image is resolved, so pdfkit, svgkit, the render image cache and layout's objectFit keep seeing an ordinary PNG and need no changes.

Browsers already ship a WebP decoder, so <Image src="photo.webp" /> works with no setup: createImageBitmap draws onto an OffscreenCanvas (falling back to a DOM canvas) and toBlob('image/png') re-encodes it.

Node has no image decoder, so the transcoder is injected:

    import sharp from 'sharp';
    import { registerWebpTranscoder } from '@react-pdf/renderer';

    registerWebpTranscoder((data) => sharp(data).png().toBuffer());

Without one, resolving a WebP throws an error naming the fix instead of silently dropping the image. The transcoder's output is sniffed again, so returning JPEG also works, and returning anything else fails with a clear message rather than crashing inside png-js.

getImage is now async. All five callers were already async functions returning it, so nothing else had to change.


DTDucas:
## What

Adds WebP support to `<Image>` / `<ImageBackground>`, for every source form the resolver
already accepts: remote URL, local path, `Buffer`, `Blob`, `data:image/webp` URI and
`{ data, format: 'webp' }`.

Closes #1252.

## Why

WebP is now the default export of most image tooling, and users hit
`Error: Not valid image extension` when they drop one into a document. Closing that gap needs
a decision about *where* the format is handled, which is the interesting part below.

## How

**PDF has no WebP.** An image XObject can carry JPEG (`DCTDecode`) or flate-compressed raw
pixels (`FlateDecode`, which is what a PNG decodes into); there is no filter for a VP8/VP8L
bitstream, and `pdfkit.openImage()` sniffs bytes for PNG/JPEG only. So WebP cannot be passed
through, it must be decoded and re-encoded.

Doing that in `@react-pdf/image`, at resolve time, keeps the change small: everything
downstream (pdfkit, `@react-pdf/svgkit`, the render image cache, layout's `objectFit`) keeps
seeing an ordinary PNG.

```
resolve.ts  getImageFormat()  sniffs `RIFF....WEBP` -> 'webp'
            getImage()        new PNG(await WEBP.transcode(bytes))
webp.ts     WEBP.transcode()  registered transcoder, else browser-native, else throw
```

Where the decoder comes from:

| environment | decoder | setup |
| --- | --- | --- |
| browser | `createImageBitmap` -> `OffscreenCanvas`/`<canvas>` -> `toBlob('image/png')` | none |
| Node | injected by the consumer | `registerWebpTranscoder(fn)` |
| neither | — | throws, naming the fix |

```js
import sharp from 'sharp';
import { registerWebpTranscoder } from '@react-pdf/renderer';

registerWebpTranscoder((data) => sharp(data).png().toBuffer());
```

A missing decoder throws rather than rendering a blank space, so the failure is visible at
development time instead of showing up as a hole in a printed document.

### Bundle cost

No decoder ships with this. Measured with the repo's own `.size-limit.cjs`, building
`packages/renderer` at `56f40997` and again on this branch:

|        | minified + brotlied, with all dependencies |
| ------ | ------------------------------------------ |
| before | 358.64 kB                                  |
| after  | 359.3 kB                                   |
| budget | 550 kB                                     |

That 0.66 kB is plumbing, not a codec: the browser path reuses the decoder the browser already
ships, and the Node path is a registration hook.

I deliberately did **not** bundle a WASM libwebp build (e.g. `@jsquash/webp`): a ~970 KB
package with a ~400 KB wasm blob, async wasm init, and bundler-specific wasm handling pushed
onto every consumer. I also measured the one published pure-TS decoder (`@stacksjs/ts-webp`)
against fixtures before considering it, and it decodes lossless WebP incorrectly, so vendoring
that route did not look safe either. Happy to revisit if you would rather it work on Node with
zero setup.

### Notes on the implementation

The transcoder may return PNG or JPEG; the result is sniffed again, so a transcoder that
returns something else fails with a clear message instead of crashing inside `png-js`.

`getImage` is now `async`. All five callers were already `async` functions that
`return getImage(...)`, so the promise is awaited by the caller's own `await`.

`RasterImage.format` deliberately stays `'jpeg' | 'png'`: by the time an image reaches the
renderer it really is one of those. `'webp'` only appears in `ImageFormat`, the *source*
format.

## Steps to execute

Browser, no setup required:

```sh
yarn install && yarn build
cd apps/examples && yarn dev
# open the "WebP Image" example: the WebP and the PNG reference render identically
```

Node:

```js
import sharp from 'sharp';
import ReactPDF, { registerWebpTranscoder } from '@react-pdf/renderer';

registerWebpTranscoder((data) => sharp(data).png().toBuffer());
await ReactPDF.renderToFile(<Doc />, 'out.pdf');
```

## Tests

`packages/image/tests/webp.test.ts`, 14 tests covering container validation (VP8/VP8L/VP8X vs
PNG/JPEG/SVG/RIFF-but-not-WebP/truncated), every source form, the transcoder receiving the
original bytes, a JPEG-returning transcoder, and the no-transcoder error.

Full suite on this branch: 905 passing, lint clean, size-limit within budget.

Also verified by hand:

- Node: a document with a WebP buffer, a `data:image/webp` URI and a local `.webp` path
  rendered to PDF and rasterised with `pdftoppm`, all three match the PNG reference.
- Browser: the new `apps/examples` entry in Chromium, with no transcoder registered.

## Also included

- Changeset (`minor` for `@react-pdf/image`, `@react-pdf/layout`, `@react-pdf/renderer`,
  `@react-pdf/types`).
- Docs: `apps/site/content/docs/v4/components/image.mdx` and `packages/image/README.md`.

.changeset/webp-image-support.md
```diff
@@ -0,0 +1,20 @@
...
+PDF has no WebP format, so WebP is decoded and re-encoded to PNG while the image is resolved.
+Browsers do this with the decoder they already ship, so `<Image src="photo.webp" />` works with
+no setup. Node has no image decoder, so register one — without it, a WebP image throws instead
+of rendering blank:
+
+```js
+import sharp from 'sharp';
+import { registerWebpTranscoder } from '@react-pdf/renderer';
+
+registerWebpTranscoder((data) => sharp(data).png().toBuffer());
```
diegomura, line 19: First of all thanks for working on this!

However, I have to say, I don't love this. I think if we will have support for webp it should just work or not have it built inside. `registerWebpTranscoder` is just a long path for something that can be done already: convert the image from webp to png outside and pass it as src to an image

DTDucas: That's fair, and it's a reasonable thing to push back on. Let me take the hook out of the
equation and give you the two shapes you described.

**Option A: no public API at all.** My preference, and the smaller diff.

The browser already ships a WebP decoder, so there the format genuinely just works:
`createImageBitmap` -> canvas -> `toBlob('image/png')`. No dependency, no registration, nothing
added to the public API. Node has no image decoder at all, so there it throws with a message
saying exactly what you said above: convert to PNG first and pass that as `src`.

That is "just works" wherever the platform can do it, and an honest, actionable error where it
can't. `registerWebpTranscoder` disappears entirely.

**Option B: make it just work everywhere.**

If you would rather not have the browser/Node asymmetry, Node can decode too without the
browser bundle paying for it. `@cwasm/webp` is libwebp compiled to wasm: 98 KB unpacked, MIT,
one dependency (`@canvas/image-data`), synchronous decode, no native binaries. I measured it
against Pillow-encoded fixtures before proposing it:

| fixture | RGBA max diff | mean |
| --- | --- | --- |
| lossless | 0 | 0.00 |
| lossy (q90) | 132 | 1.65 |

Lossless is pixel-exact and alpha is exact in both, which is what libwebp should give.

It would sit behind a lazy `import()` guarded by the existing `BROWSER` flag, the same way
`fs`/`path`/`url` already are, so the browser build never sees it and Node only loads the wasm
the first time a WebP actually appears. It decodes to raw RGBA, so it needs a small PNG encoder
on top: node `zlib` plus about 50 lines, and since `png-js` is already a dependency the tests
can round-trip the output through it.

The honest caveat: `@cwasm/webp` was last published in 2022. It is wasm-compiled libwebp with
no moving parts, so staleness matters less than usual, but it is a four-year-old dependency and
that is worth weighing.

Happy to push either. If you don't say otherwise I'll go with A, since it removes API surface
rather than adding it.

diegomura: I'm fine with the browser/node asymmetry on code (we have that pattern in other places) but not with the public api. Option A seems fine but I wonder what's the performance of  `createImageBitmap -> canvas -> toBlob('image/png')`. We don't have to throw though, we could include sharp png conversion internally just for node.

Ideally though we would have an isomorphic webp decoder/ PNG encoder (no node nor browser deps) but that seems too much work atm

DTDucas: Measured it. Chromium 145, Apple silicon, median of 7 runs per size, all three encoded at q85
from the same source photo:

| source | createImageBitmap | canvas -> PNG | total | PNG out |
| --- | --- | --- | --- | --- |
| 700x420, 59 KB | 3.4 ms | 6.6 ms | **10 ms** | 205 KB |
| 1920x1080, 221 KB | 13.1 ms | 26.8 ms | **40 ms** | 1068 KB |
| 4000x3000, 505 KB | 42.2 ms | 124.5 ms | **166 ms** | 4497 KB |

The `drawImage` is noise (0.1-0.2 ms), so it is decode plus encode, and it runs once per unique
source because `IMAGE_CACHE` holds the resolved image.

The time looks acceptable to me, but that last column is the real problem: a 505 KB WebP
becomes a 4.5 MB PNG, and pdfkit embeds the IDAT as-is, so the document carries all of it.
Encoding to JPEG instead is both faster and much smaller:

| source | PNG | JPEG (q 0.92) |
| --- | --- | --- |
| 700x420 | 10.0 ms / 205 KB | 10.4 ms / 94 KB |
| 1920x1080 | 39.5 ms / 1068 KB | 29.6 ms / 403 KB |
| 4000x3000 | 179.8 ms / 4497 KB | 97.8 ms / 1175 KB |

So I would pick the output format from the WebP container, which is a header read, no pixel
scan:

- `VP8 ` (simple lossy) has no alpha by definition -> JPEG
- `VP8X` with the alpha flag clear -> JPEG
- anything else (`VP8L`, or `VP8X` with alpha) -> PNG

A lossy WebP is already lossy, so a q92 JPEG costs one more generation rather than introducing
loss where there was none, and it keeps the PDF near the size the source implies. Lossless and
alpha WebP still go through PNG.

On Node: agreed, no throw, and no public API either way. One hesitation about sharp
specifically. It pulls libvips as a native dependency with per-platform prebuilt binaries,
which is a lot to add to `@react-pdf/image` (currently `@react-pdf/svg`, `jay-peg`, `png-js`).
`@cwasm/webp` is libwebp compiled to wasm: 98 KB unpacked, MIT, no native build, synchronous
decode. I checked it against Pillow-encoded fixtures before suggesting it: lossless
pixel-exact, alpha exact, lossy q90 mean error 1.65/255. Its drawback is that it was last
published in 2022. It decodes to raw RGBA, so Node would also need a small PNG encoder, which
is node `zlib` plus about 50 lines, and since `png-js` is already a dependency the tests can
round-trip the output through it.

Either works for me, it is your dependency budget: sharp if you would rather have something
actively maintained, `@cwasm/webp` if you would rather not add a native dep.

On the isomorphic ideal: the wasm decoder is already the isomorphic half, it runs unchanged in
a browser. It is the encode side that is platform-specific, and there the canvas does it for
free and faster than anything we would ship. So this split may be closer to the endpoint than
it looks.

diegomura: What about using `@napi-rs/canvas` instead in node? so we can just share the canvas layer. Not sure it would work, and I realize it's a bit of a more indirect fix, but I'm thinking if that works we could easily add support for [CSS Image filters](https://www.w3schools.com/css/css3_image_filters.asp)

DTDucas: It works. Ran it against the same fixtures as the numbers above, `@napi-rs/canvas` 1.0.8:

| image | node PNG | node JPEG | browser PNG | browser JPEG |
| --- | --- | --- | --- | --- |
| 700x420 | 17.8 ms / 237 KB | 4.7 ms / 94 KB | 10.0 ms / 205 KB | 10.4 ms / 94 KB |
| 1920x1080 | 103 ms / 1183 KB | 20.8 ms / 403 KB | 39.5 ms / 1068 KB | 29.6 ms / 403 KB |
| 4000x3000 | 497 ms / 4159 KB | 74.7 ms / 1175 KB | 179.8 ms / 4497 KB | 97.8 ms / 1175 KB |

`loadImage()` decodes WebP through skia: lossy, lossless and alpha all come through. And
`ctx.filter = 'blur(2px) grayscale(1)'` is accepted, so the CSS image filter idea is real
rather than hopeful.

Two things worth noting from the numbers. Node's PNG encode is roughly 2.8x slower than
Chromium's, while its JPEG encode is faster than Chromium's, so the "JPEG unless the container
says alpha" rule matters more on Node than it does in the browser. And the JPEG output sizes
come out identical in both environments (94 / 403 / 1175 KB), which is a nice property to have.

One caveat, which I think actually argues for your idea. A lossless WebP with semi-transparent
pixels does not round-trip exactly through a canvas: those pixels come back within 2/255,
because canvas stores premultiplied alpha. Opaque pixels are bit-exact. I measured both sides
to check whether that was a skia thing:

node    @napi-rs/canvas : opaque maxdiff 0, semi-transparent maxdiff 2
browser Chromium 145    : opaque maxdiff 0, semi-transparent maxdiff 2

Identical. So it is a canvas property rather than a napi-rs one, and sharing the canvas layer
means both environments produce the same output, which a sharp-based Node path could not
promise.

The cost is that `@napi-rs/canvas` is a native dependency: 148 KB of JS plus a ~27 MB
prebuilt binary per platform. That is the same class of cost as sharp, so it does not dodge the
native-dependency question, but it buys considerably more than sharp would: one drawing and
export path instead of two, and the place where filters would eventually live.

The two APIs are not literally the same shape (`createImageBitmap` + `OffscreenCanvas` +
`convertToBlob` versus `loadImage` + `createCanvas` + `toBuffer`), so "shared" means a thin
adapter of about fifteen lines per environment and one shared draw-and-export path above it.
Still the right seam for filters later.

I am happy to build it this way. The only question I would put back to you: hard dependency, so
Node just works, or optional, so people who never touch WebP do not pay the 27 MB? Your call
and I will push it either way.

diegomura: Hard dependency on node is fine. Bundle size in Node is not that important as it is for browser. Unless you see a different way of having CSS image filters. Worths checking

DTDucas: Checked it, and I don't see a better way for images, but the measurements split the question in
a way I think is worth having on record.

**The colour-matrix half does not need a canvas.** The Filter Effects spec defines `grayscale`,
`sepia`, `saturate`, `hue-rotate`, `invert`, `brightness`, `contrast` and `opacity` as
`feColorMatrix`/`feComponentTransfer`, all of which are affine, so a whole chain composes into
one 4x5 matrix and applies in a single pass. I implemented that in plain JS and diffed it
against Chromium's `ctx.filter` on the same image, chaining seven of them:

grayscale(0.6) sepia(0.3) saturate(1.4) hue-rotate(30deg) invert(0.1) brightness(1.1) contrast(1.2)

max channel diff : 6/255
mean channel diff: 0.524



The residual is Chromium rounding to 8 bits between stages while the composed matrix rounds
once, so if anything the matrix is the more accurate of the two.

**Blur is the exception.** I implemented the three-box approximation exactly as the SVG spec
defines it and compared against `blur(4px)`, interior only so edge padding doesn't pollute it:

max channel diff : 31/255
mean channel diff: 3.571



That is not a match. Chromium runs a real Gaussian rather than the spec's box approximation, so
matching a browser on blur means using the same rasterizer the browser uses. `drop-shadow`
inherits this since it is blur plus offset plus composite.

**But the decisive part is not the filter math, it is that filters need decoded pixels.** Today
PNG and JPEG bytes go to pdfkit untouched. The moment any filter applies, that image has to be
decoded, filtered and re-encoded. Without a canvas, Node would need a WebP decoder, a JPEG
pixel decoder (`jay-peg` parses markers, it doesn't decode pixels), and a PNG/JPEG encoder.
`@napi-rs/canvas` is all three in one dependency, running the same skia code the browser runs,
which also means the two environments produce the same output rather than merely similar
output.

So the canvas is not the expensive route to filters. It is the cheap route to
decode-and-encode, and filters come along with it. Hard dependency on Node it is, and I'll
build it that way.

One thing worth keeping in your back pocket: the matrix half stays useful for content that is
never rasterised. A `filter: grayscale(1)` on a `View` full of vector children could map fill
and stroke colours through the same matrix at draw time, with no rasterisation and no
dependency at all. Different code path from images, but the same matrix code would serve both.

Caveat on all of the above: measured on Chromium 145 only, I did not check Firefox or Safari.

